PR #860, opened 2023-06-22
Fix: Network icons shimmering

egor0798:
## Fixes:

- Added network icon shimmering ro balances, select shards modal and create watch only modal
- Moved Chain component to ui-redesign
- Created ChainIcon component with shimmering


Asmadek (approved):

tuul-wq (changes requested):
Minor suggestions

tuul-wq (approved):

PR #910, opened 2023-06-28
Fix: App stuck after multiple signings

egor0798:


src/renderer/components/common/QrCode/QrReader/QrSignatureReader.tsx
```diff
@@ -131,6 +137,7 @@ const QrSignatureReader = ({
           await startScanning();
         }
       } catch (error) {
+        console.log('error_1');
```
Asmadek, line 140: maybe better to delete it?

Asmadek (approved):

tuul-wq (approved):

PR #912, opened 2023-06-29
Fix: native token fee validation

Asmadek:
fix native token fee validation for approve/reject tx,  and multisig/multishard/simple transfer creation

https://app.clickup.com/t/85ztdckm5

### DIFF
--- a/src/renderer/components/common/SelectableSignatory/SelectableSignatory.tsx
+++ b/src/renderer/components/common/SelectableSignatory/SelectableSignatory.tsx
@@ -1,5 +1,3 @@
-// import { useEffect, useState } from 'react';
-
 import cnTw from '@renderer/shared/utils/twMerge';
 import AccountAddress, {
   getAddress,
@@ -20,8 +18,8 @@ type Props<T extends any> = {
   explorers?: Explorer[];
   value: T;
   onSelected: (value: T) => void;
-  chainId?: ChainId;
-  asset?: Asset;
+  chainId: ChainId;
+  asset: Asset;
 } & AccountAddressProps;
 
 const SelectableSignatory = <T extends any>({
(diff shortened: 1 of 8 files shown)


src/renderer/components/common/SelectableSignatory/SelectableSignatory.tsx
```diff
@@ -36,18 +36,10 @@ const SelectableSignatory = <T extends any>({
...
-  const [balance, setBalance] = useState('');
-
-  useEffect(() => {
-    if (chainId && asset) {
-      getBalance(toAccountId(address), chainId, asset.assetId.toString()).then((b) =>
-        setBalance(transferableAmount(b)),
-      );
-    }
-  }, [chainId, asset, address]);
+  const balance = chainId && asset && getLiveBalance(toAccountId(address), chainId, asset.assetId.toString());
```
tuul-wq, line 42: `getLiveBalance` is react hook, so it can't be conditional
If for some reason `chainId` or `asset` change between renders application will fail with error
We gotta make sure that `asset` and `chainId` are defined before we pass it to `SelectableSignatory `

src/renderer/components/common/SelectableSignatory/SelectableSignatory.tsx
```diff
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+// import { useEffect, useState } from 'react';
```
tuul-wq, line 1: Can be deleted 😉

src/renderer/screens/Transfer/components/TransferForm.tsx
```diff
@@ -215,6 +215,7 @@ export const TransferForm = ({
     if (nativeTokenBalance) {
       return new BN(fee).lte(new BN(nativeTokenBalance));
     }
+
```
tuul-wq, line 218: 👍

tuul-wq (changes requested):

tuul-wq (approved):

egor0798 (approved):

PR #920, opened 2023-06-30
fix: reset validation on disable light client

Asmadek:


src/renderer/screens/Settings/Networks/Networks.tsx
```diff
@@ -133,11 +142,27 @@ export const Networks = () => {
     };
   };
 
+  const resetBalanceValidation = async (relaychainId: ChainId) => {
+    const parachains = getParachains(relaychainId);
+    const allAccounts = await getAccounts();
+    const uniqAccounts = await uniqBy(allAccounts, 'accountId');
```
tuul-wq, line 148: `uniqBy` is not a promise 🙂 
```javascript
    const uniqAccounts = uniqBy(allAccounts, 'accountId');
```

src/renderer/screens/Settings/Networks/Networks.tsx
```diff
@@ -133,11 +142,27 @@ export const Networks = () => {
...
+
+    parachains.forEach(({ chainId, assets }) => {
```
tuul-wq, line 150: I'm worried that this can iterate a lot like 15000 times for 100 accounts (50 para, 3 assets, 100 accs)
Maybe there is some way to make a batch update?

tuul-wq (changes requested):

src/renderer/screens/Settings/Networks/Networks.tsx
Asmadek: Batch update will be implemented in dexie 4.0, (I will add todo comment) now we have only put, which needs to rewrite full balance object, and need to get all balances first, but now current version works really fast for 30 accounts

tuul-wq (approved):

PR #921, opened 2023-06-30
Fix: Validation

tuul-wq:
- Removed fee validation for multisig staking (validators, destination)
- Validate deposit before fee
- Generic for dropdown
- Set new selected network of staking in localstorage (it was meant to work but got lost for whatever reason)

[Task](https://app.clickup.com/t/863g9y8nf)


src/renderer/services/staking/eraService.ts
tuul-wq, line 1: Renamed a couple of variables in order align with `getUnbondingPeriod`

Asmadek (approved):

egor0798 (approved):

PR #922, opened 2023-06-30
Fix: changed signatory name source priority

egor0798:
Fixes:

- Changed signatory name source priority
- Added padding for version block on settings

### DIFF
--- a/src/renderer/components/common/SignatoryCard/SignatoryCard.tsx
+++ b/src/renderer/components/common/SignatoryCard/SignatoryCard.tsx
@@ -18,6 +18,7 @@ const IconProps = {
 type Props = {
   explorers?: Explorer[];
   status?: SigningStatus;
+  matrixId?: string;
 } & AccountAddressProps;
 
 const SignatoryCard = ({
@@ -26,12 +27,20 @@ const SignatoryCard = ({
   addressFont = 'text-body text-inherit',
   size = 20,
   name,
+  matrixId,
   ...addressProps
 }: Props) => {
   const { getLiveAccounts } = useAccount();
   const address = getAddress(addressProps);
   const account = getLiveAccounts().find((a) => toAddress(a.accountId) === address);
   const popoverItems = useAddressInfo(address, explorers, true);
+
+  if (!popoverItems.find((item) => item.title === 'Matrix ID') && matrixId) {
+    popoverItems.push({
+      title: 'Matrix ID',
+      items: [{ id: matrixId, value: matrixId }],
+    });
+  }
 
   return (
     <InfoPopover data={popoverItems} buttonClassName="w-full" position="right-0 left-unset">
--- a/src/renderer/components/common/SignatoryCard/SignatoryCard.tsx
+++ b/src/renderer/components/common/SignatoryCard/SignatoryCard.tsx
@@ -7,8 +7,6 @@ import { Icon } from '@renderer/components/ui';
 import { Explorer } from '@renderer/domain/chain';
 import { SigningStatus } from '@renderer/domain/transaction';
 import useAddressInfo from '@renderer/components/common/AccountAddress/useAddressInfo';
-import { useAccount } from '@renderer/services/account/accountService';
-import { toAddress } from '@renderer/shared/utils/address';
 
 const IconProps = {
   SIGNED: { className: 'group-hover:hidden text-text-positive', name: 'checkLineRedesign' },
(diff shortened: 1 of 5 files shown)


Asmadek (approved):

src/renderer/screens/Operations/common/utils.ts
```diff
@@ -219,3 +222,23 @@ export const getTransactionAmount = (tx: Transaction): string | null => {
 
   return null;
 };
+
+export const getSignatoryName = (
```
tuul-wq, line 226: Better use domain types, there are no `id's` down below
```suggestion
export const getSignatoryName = (
  signatoryId: AccountId,
  tx: MultisigTransaction,
  contacts: Contact[],
  accounts: Account[],
  addressPrefix?: number,
): string => {
```

src/renderer/screens/Operations/common/utils.ts
```diff
@@ -219,3 +222,23 @@ export const getTransactionAmount = (tx: Transaction): string | null => {
...
+  signatoryId: AccountId,
+  tx: MultisigTransaction,
```
tuul-wq, line 228: If we only get `signatories` from `tx` then it's preferably to pass `tx.signatories`
So the argument would become - `signatories: Signatory[]`

src/renderer/screens/Operations/components/Log.tsx
```diff
@@ -37,50 +34,29 @@ const EventMessage: Partial<Record<SigningStatus | 'INITIATED', string>> = {
...
-  };
-
   const getEventMessage = (event: MultisigEvent): string => {
-    const signatory = getSignatory(event.accountId);
     const isCreatedEvent =
-      signatory?.accountId === tx.depositor && (event.status === 'SIGNED' || event.status === 'PENDING_SIGNED');
+      event.accountId === tx.depositor && (event.status === 'SIGNED' || event.status === 'PENDING_SIGNED');
 
-    const signatoryName = signatory?.name || toAddress(event.accountId, { chunk: 5, prefix: chain?.addressPrefix });
+    const signatory = getSignatoryName(event.accountId, tx, contacts, accounts, connection?.addressPrefix);
```
tuul-wq, line 55: `signatoryName` was a good name, even function are called so ☺️

src/renderer/screens/Operations/components/Log.tsx
```diff
@@ -1,30 +1,27 @@
...
-import { AccountId } from '@renderer/domain/shared-kernel';
-import { useContact } from '@renderer/services/contact/contactService';
-import { useAccount } from '@renderer/services/account/accountService';
+import { AccountDS, ContactDS } from '@renderer/services/storage';
 
 type Props = {
   tx: MultisigTransaction;
   account?: MultisigAccount;
   connection?: ExtendedChain;
+  accounts: AccountDS[];
```
tuul-wq, line 23: `Account` `Contact` if possible 👍

src/renderer/screens/Operations/components/OperationFullInfo.tsx
```diff
@@ -144,14 +152,15 @@ const OperationFullInfo = ({ tx, account }: Props) => {
         </div>
 
         <ul className="flex flex-col gap-y-0.5">
-          {signatoriesList.map(({ accountId, name }) => {
+          {signatoriesList.map(({ accountId, matrixId }) => {
             return (
```
tuul-wq, line 156: `return` can be omitted 🙂

src/renderer/components/common/SignatoryCard/SignatoryCard.tsx
```diff
@@ -26,13 +27,21 @@ const SignatoryCard = ({
   addressFont = 'text-body text-inherit',
   size = 20,
   name,
+  matrixId,
   ...addressProps
 }: Props) => {
   const { getLiveAccounts } = useAccount();
   const address = getAddress(addressProps);
   const account = getLiveAccounts().find((a) => toAddress(a.accountId) === address);
```
tuul-wq, line 35: In worst case we're guaranteed to receive `address`
`name={getSignatoryName(accountId, tx, contacts, accounts, connection?.addressPrefix)}`
So condition will never fall into `account?.name` - `<AccountAddress ... name={name || account?.name} />`
Guess it's safe to remove `useAccount` from `SignatoryCard`
Please correct me if I'm wrong 🙏

tuul-wq (commented):
Cool!

PR #925, opened 2023-07-01
feat: add latest release description and zip for mac

pgolovkin:
Add latest release description and zip for mac. This is required for the electron auto-update start.
1. yaml files for each platform with release metadata https://www.electron.build/auto-update#file-generated-and-uploaded-in-addition
2. zip archives for mac build is required https://www.electron.build/auto-update#quick-setup-guide

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -64,6 +64,10 @@ jobs:
             release/dist/*.exe
             release/dist/*.dmg
             release/dist/*.AppImage
+            release/dist/latest.yml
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -107,3 +111,5 @@ jobs:
             *.exe
             *.AppImage
             *.sha256
+            latest*.yml
+            *mac.zip
```
stepanLav, line 115: Added new artefacts to release creation step

PR #928, opened 2023-07-03
Fix: added assets sortig

egor0798:


src/renderer/screens/Balances/NetworkBalances/NetworkBalances.tsx
```diff
@@ -70,20 +71,37 @@ const NetworkBalances = ({
...
+          );
+        })
+        .sort((a, b) => {
+          const aTotal = balancesObject[a.assetId.toString()] ? totalAmount(balancesObject[a.assetId.toString()]) : '0';
+          const bTotal = balancesObject[b.assetId.toString()] ? totalAmount(balancesObject[b.assetId.toString()]) : '0';
+
+          const aFormatted = formatBalance(aTotal, a.precision);
+          const bFormatted = formatBalance(bTotal, b.precision);
+
+          return Number(bFormatted.value) - Number(aFormatted.value) || a.name.localeCompare(b.name);
```
Asmadek, line 101: `value` is a shortened value and 10 will be higher than 1M, I am not sure about requirements but it looks like total amount should be compared

Asmadek (changes requested):

Asmadek (approved):

src/renderer/screens/Balances/components/NetworkBalances/NetworkBalances.tsx
```diff
@@ -70,20 +50,29 @@ const NetworkBalances = ({
...
-      );
-    }
-
-    const balance = balancesObject[asset.assetId];
-
-    return (
-      !hideZeroBalance || !balance || balance.verified !== true || (balance && totalAmount(balance) !== ZERO_BALANCE)
-    );
-  });
+  const filteredAssets = useMemo(
```
tuul-wq, line 53: `const balances = getLiveNetworkBalances(accountIds, chain.chainId);`
It gives us new array instance every time balances are being updated in DB which is fine
But here comes `useMemo` 😄
Memo will store all those different `balances` with all related filter results that actually won't be any helpful because we can't receive same `balances` instance again so the memo cannot provide already calculated value.

tuul-wq (changes requested):
Nice 👍
Left one comment about `useMemo`

tuul-wq (approved):

PR #929, opened 2023-07-03
Fix: fixed connection undefined for signatory card

egor0798:


src/renderer/screens/Operations/components/OperationFullInfo.tsx
```diff
@@ -155,7 +155,7 @@ const OperationFullInfo = ({ tx, account }: Props) => {
           {signatoriesList.map(({ accountId, matrixId }) => (
             <li key={accountId}>
               <SignatoryCard
-                addressPrefix={connection.addressPrefix}
+                addressPrefix={connection?.addressPrefix}
```
pgolovkin, line 158: Address prefix should be independent from the network connection. It's a part of the configuration (chain.json).

PR #932, opened 2023-07-03
Feat: Use accordion

tuul-wq:
- Use accordion on `Assets` / `Operations`
- Added `card-shadow` on hover/focus to accordion blocks (according to design)
- Moved `Balances` to `Assets`, updated exports

[Task](https://app.clickup.com/t/86787vvpq)

### DIFF
--- a/src/renderer/routes/index.tsx
+++ b/src/renderer/routes/index.tsx
@@ -12,9 +12,9 @@ const routesConfig: RouteObject[] = [
     path: Paths.ROOT,
     element: <Layouts.PrimaryLayout />,
     children: [
-      { index: true, element: <Navigate to={Paths.BALANCES} replace /> },
+      { index: true, element: <Navigate to={Paths.ASSETS} replace /> },
       { path: Paths.OPERATIONS, element: <Screens.Operations /> },
-      { path: Paths.BALANCES, element: <Screens.Assets.Assets /> },
+      { path: Paths.ASSETS, element: <Screens.Assets.Assets /> },
       { path: Paths.NOTIFICATIONS, element: <Screens.Notifications /> },
       {
         path: Paths.ADDRESS_BOOK,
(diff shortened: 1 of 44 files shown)


src/renderer/screens/Assets/Assets.tsx
```diff
@@ -135,17 +132,18 @@ const Balances = () => {
...
-                  hideZeroBalance={hideZeroBalance}
-                  searchSymbolOnly={searchSymbolOnly}
-                  query={query.toLowerCase()}
-                  chain={chain}
-                  accountIds={accountIds}
-                  canMakeActions={checkCanMakeActions()}
-                  onReceiveClick={onReceive(chain)}
-                  onTransferClick={onTransfer(chain)}
-                />
+                <li key={chain.chainId} className="w-[546px] mx-auto">
```
tuul-wq, line 135: Retrieved `<li>` from `NetworkAssets` so child component could be independent from environment

src/renderer/theme/default.css
```diff
@@ -61,6 +61,7 @@
   --divider: rgba(69, 69, 137, 0.06);
 
   /* BACKGROUND */
+  --background-default: #f8f8fa;
```
tuul-wq, line 64: Added default background color, so we won't be afraid of opacity that might play in

src/renderer/routes/index.tsx
```diff
@@ -14,7 +14,7 @@ const routesConfig: RouteObject[] = [
     children: [
       { index: true, element: <Navigate to={Paths.BALANCES} replace /> },
       { path: Paths.OPERATIONS, element: <Screens.Operations /> },
-      { path: Paths.BALANCES, element: <Screens.Balances /> },
```
tuul-wq, line 17: The idea of `balances` page that's named `assets` in navigation menu did not let me go for a while.
Balance doesn't mean much without an asset it relates to. 
So it seemed logical to finally make "Assets first" and "Balances second" and rename corresponding folder/files.

egor0798 (approved):

PR #934, opened 2023-07-03
fix: metadata portal link leads to metadata page

pgolovkin:
Link example https://metadata.novasama.io/?tab=1#/kusama

### DIFF
--- a/src/renderer/components/common/QrCode/common/constants.ts
+++ b/src/renderer/components/common/QrCode/common/constants.ts
@@ -42,11 +42,11 @@ export const SIGNED_TRANSACTION_BULK = taggedUnion('SignaturesBulk', {
   4: ['V1', ['payload', array(MULTI_SIGNATURE)]],
 });
 
-const METADATA_PORTAL_URL = 'https://metadata.novasama.io/#/';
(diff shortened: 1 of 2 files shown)


src/renderer/components/common/QrCode/common/constants.ts
```diff
@@ -42,11 +42,11 @@ export const SIGNED_TRANSACTION_BULK = taggedUnion('SignaturesBulk', {
   4: ['V1', ['payload', array(MULTI_SIGNATURE)]],
 });
 
-const METADATA_PORTAL_URL = 'https://metadata.novasama.io/#/';
+const METADATA_PORTAL_METADATA_URL = 'https://metadata.novasama.io/?tab=1#/';
```
tuul-wq, line 45: Are you sure about `https://metadata.novasama.io/?tab=1#/`?
Maybe it's something like this - `https://metadata.novasama.io/#/?tab=1`?

tuul-wq (approved):

src/renderer/components/common/QrCode/common/constants.ts
pgolovkin: Yes, I'm sure. I've checked it. https://metadata.novasama.io/#/?tab=1/kusama does nothing.

PR #937, opened 2023-07-04
Fix: added signatory checkbox disabled if it selected account in either of tabs

egor0798:


src/renderer/components/modals/CreateMultisigAccount/components/SelectSignatories.tsx
```diff
@@ -109,6 +109,9 @@ export const SelectSignatories = ({ isActive, wallets, accounts, contacts, onSel
     });
   };
 
+  const isDisabled = (index: string, accountId: AccountId) =>
```
tuul-wq, line 112: A bit cleaner version with type and braces:
```suggestion
const isDisabled = (index: string, accountId: AccountId): boolean => {
    return (
      isSameItemSelected(index, selectedWallets[accountId]) || isSameItemSelected(index, selectedContacts[accountId])
    );
  };
```

tuul-wq (approved):
Good to see how smooth this fix was added 👍

PR #942, opened 2023-07-04
Feat: added truncate function

egor0798:
Added simple truncate function for cases where we don't need adapvtive truncate. 
[task](https://app.clickup.com/t/85ztek7qw)


src/renderer/shared/utils/strings.ts
```diff
@@ -57,3 +57,16 @@ export const includes = (value?: string, searchString?: string): boolean => {
...
 };
+
+/**
+ * Truncate text leaving fixed number of characters
+ * @param text text to truncate
+ * @param start number of leading symbols
+ * @param end number of ending symbols
+ * @return {String}
+ */
+export const truncate = (text: string, start = 5, end = 5): string => {
```
Asmadek, line 68: we have exactly the same function `toShortAddress`, maybe combine it or use one inside another?

Asmadek (approved):
one comment

tuul-wq (approved):

PR #944, opened 2023-07-05
Feat: Join existing room

tuul-wq:
- Added ability to join existing rooms
- Added alert if we're trying to create existing multisig account

[Task](https://app.clickup.com/t/860qf564r)
image


src/renderer/components/modals/CreateMultisigAccount/CreateMultisigAccount.tsx
```diff
@@ -74,43 +75,69 @@ export const CreateMultisigAccount = ({ isOpen, onClose }: Props) => {
...
-  const onCreateAccount = async (name: string, threshold: number) => {
+  const onCreateAccount = async (name: string, threshold: number, creatorId: AccountId): Promise<void> => {
     setName(name);
     toggleLoading();
     toggleResultModal();
 
-    const inviter = signatories.find((s) => s.matrixId === matrix.userId);
-    if (!inviter || !threshold) return;
+    try {
+      const roomId = matrix.joinedRooms()[0]?.roomId;
```
Asmadek, line 84: shouldn't be accountId provided here?

Asmadek (commented):
one question

Asmadek (approved):

PR #947, opened 2023-07-06
fix: loading balance for multishard

Asmadek:


src/renderer/screens/Assets/components/NetworkAssets/NetworkAssets.tsx
```diff
@@ -41,15 +44,34 @@ export const NetworkAssets = ({
...
+  const [balancesObject, setBalancesObject] = useState<Record<string, Balance>>({});
+
+  const accountIds = useMemo(
+    () =>
+      accounts.reduce<AccountId[]>((acc, account) => {
+        if (!account.chainId || account.chainId === chain.chainId) acc.push(account.accountId);
+
+        return acc;
+      }, []),
+    [chain.chainId, accounts.map((a) => a.accountId).join('')],
```
tuul-wq, line 56: Let's make it a variable
And add `{ }` inside `useMemo` for readability

src/renderer/screens/Assets/components/NetworkAssets/NetworkAssets.tsx
```diff
@@ -41,15 +44,34 @@ export const NetworkAssets = ({
...
 
   const balances = getLiveNetworkBalances(accountIds, chain.chainId);
 
-  const balancesObject =
-    balances?.reduce<Record<string, Balance>>((acc, balance) => {
-      acc[balance.assetId] = sumBalances(balance, acc[balance.assetId]);
+  useEffect(() => {
+    const newBalancesObject = Object.values(groupBy(balances, 'assetId')).reduce<Record<string, Balance>>(
+      (acc, balances: BalanceDS[]) => {
+        if (balances.length !== new Set(accountIds).size) return acc;
```
tuul-wq, line 64: `new Set(accountIds).size` can be moved out of `newBalancesObject` so it's not being calculated every iteration 😉
We can invert the condition so `return acc` will present only once

src/renderer/screens/Assets/components/NetworkAssets/NetworkAssets.tsx
```diff
@@ -63,12 +85,13 @@ export const NetworkAssets = ({
 
       const balance = balancesObject[asset.assetId];
 
-      return !hideZeroBalance || !balance?.verified || totalAmount(balance) !== ZERO_BALANCE;
+      return !hideZeroBalance || balance?.verified === false || (balance && totalAmount(balance) !== ZERO_BALANCE);
```
tuul-wq, line 88: 👍

tuul-wq (changes requested):

tuul-wq (approved):

PR #949, opened 2023-07-06
fix: first balance load for multisig

Asmadek:


src/renderer/screens/Assets/Assets.tsx
```diff
@@ -58,18 +58,22 @@ export const Assets = () => {
...
+  ).length;
+
   const sortedChains = useMemo(
     () =>
       sortChains(
         Object.values(connections).filter((c) => {
           const isDisabled = c.connection.connectionType === ConnectionType.DISABLED;
           const hasMultisigAccount = activeAccounts.some(isMultisig);
-          const hasMultiPallet = !hasMultisigAccount || Boolean(c.api?.tx.multisig);
+          const hasMultiPallet = !hasMultisigAccount || c.connection.hasMultisigPallet !== false;
```
tuul-wq, line 71: `c.connection.hasMultisigPallet !== false;` should be `true` or `undefined`?

src/renderer/domain/connection.ts
```diff
@@ -9,6 +9,7 @@ export type Connection = {
   customNodes?: RpcNode[];
   activeNode?: RpcNode;
   disconnect?: () => void;
+  hasMultisigPallet?: boolean;
```
tuul-wq, line 12: Does It need to be optional?
I see `hasMultisigPallet: Boolean(api?.tx.multisig)` down below, looks like we have the value all the time

tuul-wq (commented):
Some comments

src/renderer/screens/Assets/Assets.tsx
Asmadek: right, it is for the first loading

src/renderer/domain/connection.ts
Asmadek: Not all the time, we have some time between first app start and first chain connection

Same for when we will add new chains

tuul-wq (approved):

PR #956, opened 2023-07-08
Feat: Use callDataDecoder

pgolovkin:
Sorry for Java style. It's just and attempt to use OOP design with patterns.


pgolovkin:
Call data was parsed with new approach. Bonus - added method and section for unknown operation if call data is provided
image

src/renderer/services/transaction/callDataDecoder/callDataDecoder.ts
Asmadek, line 1: Better to rewrite this class in style of our services. 

We can use style of this [function](https://github.com/novasamatech/nova-spektr/blob/23a0b348f1c9808b331b51e22ca1694236915917/src/renderer/services/transaction/transactionService.ts#L73) to avoid a lot of if-else cases. 

Even we can make universal function and use data from `api.tx.palletName.callName.meta` and use `palletName.callName` as `TransactionType`. But it will lead to rewrite code where we use our custom `TransactionType`

src/renderer/services/transaction/callDataDecoder/callDataDecoder.ts
```diff
@@ -0,0 +1,505 @@
...
+  }
+}
+
+class MultisigAsMultiCallDataArgsParser extends AbstractCallDataArgsParser {
+  public parseDecodedCallArgs(
+    method: string,
+    section: string,
+    decoded: SubmittableExtrinsic<'promise'>,
+  ): Record<string, any> {
+    if (decoded.args.length == 5) {
```
Asmadek, line 425: Better to avoid to use `magic numbers` and use constants. Here you can use `OLD_MULTISIG_ARGS_AMOUNT` from `constants` file

src/renderer/services/transaction/callDataDecoder/callDataDecoder.ts
pgolovkin: I've tried this approach and faced problems:
1. For assets and balances transfers we have a list of methods `const transferMethods = ['transfer', 'transferKeepAlive', 'transferAllowDeath'];`
2. The unknown operation type. I tried to add this type of operation but it leads to lot of changes in the code.

tuul-wq:
Moved from OOP to Functional approach

tuul-wq (approved):

src/renderer/screens/Operations/common/utils.ts
```diff
@@ -231,14 +238,18 @@ export const getSignatoryName = (
   accounts: Account[],
   addressPrefix?: number,
 ): string => {
+  const finderFn = <T extends { accountId: AccountId }>(collection: T[]): T | undefined => {
```
egor0798, line 241: nice!

src/renderer/screens/Operations/components/modals/ApproveTx.tsx
```diff
@@ -69,13 +78,22 @@ const ApproveTx = ({ tx, account, connection }: Props) => {
...
+  useEffect(() => {
+    if (!signAccount?.accountId) return;
 
-  const transactionTitle = getTransactionTitle(tx.transaction);
+    setApproveTx(getMultisigTx(signAccount?.accountId));
+  }, [tx, accounts.length, signAccount?.accountId, txWeight]);
+
+  useEffect(() => {
+    setFeeTx(getMultisigTx(TEST_ADDRESS));
+  }, [tx, accounts.length, signAccount?.accountId, txWeight]);
```
egor0798, line 89: Do we need accounts.length as a dependancy here?

src/renderer/services/transaction/callDataDecoder.ts
```diff
@@ -0,0 +1,241 @@
...
+  };
+
+  const parseBatch = (
+    method: string,
+    section: string,
+    address: Address,
+    decoded: SubmittableExtrinsic<'promise'>,
+    api: ApiPromise,
+  ): DecodedTransaction => {
+    let transactionType: TransactionType | undefined = undefined;
```
egor0798, line 43: We can leave it just 
`let transactionType: TransactionType | undefined`

src/renderer/screens/Operations/components/modals/ApproveTx.tsx
egor0798, line 1: Should we update RejectTx in the same manner since these files almost identical?

egor0798 (commented):
couple comments

egor0798 (approved):

PR #962, opened 2023-07-12
Fix: Chains sort by groups

tuul-wq:
- Sort by groups according to [doc](https://docs.google.com/document/d/1_3esMRcmV3wrYwZrrM-0auycWwzMEnqA9DDyech74c8/edit#heading=h.kz34ra1ily4z)

[Task](https://app.clickup.com/t/2yt0dk7)

### DIFF
--- a/src/renderer/screens/Assets/Assets.tsx
+++ b/src/renderer/screens/Assets/Assets.tsx
@@ -1,11 +1,11 @@
-import { useEffect, useMemo, useState } from 'react';
+import { useEffect, useState } from 'react';
 
 import { Icon } from '@renderer/components/ui';
 import { useI18n } from '@renderer/context/I18nContext';
 import { useNetworkContext } from '@renderer/context/NetworkContext';
 import { Asset } from '@renderer/domain/asset';
 import { Chain } from '@renderer/domain/chain';
-import { ConnectionStatus, ConnectionType } from '@renderer/domain/connection';
+import { ConnectionType } from '@renderer/domain/connection';
 import { SigningType } from '@renderer/domain/shared-kernel';
 import { useToggle } from '@renderer/shared/hooks';
 import { useChains } from '@renderer/services/network/chainsService';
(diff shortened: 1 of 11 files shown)


src/renderer/screens/Assets/Assets.tsx
```diff
@@ -51,28 +51,25 @@ export const Assets = () => {
...
-          const isDisabled = c.connection.connectionType === ConnectionType.DISABLED;
-          const hasMultisigAccount = activeAccounts.some(isMultisig);
-          const hasMultiPallet = !hasMultisigAccount || c.connection.hasMultisigPallet !== false;
-
-          return !isDisabled && hasMultiPallet;
-        }),
-      ),
-    [availableConnectionsAmount, activeAccounts],
-  );
+  useEffect(() => {
```
tuul-wq, line 57: Switched to `useEffect` because we'll end up storing lots of sorted arrays in memory

src/renderer/services/network/chainsService.ts
```diff
@@ -45,21 +51,67 @@ export function useChains(): IChainService {
...
+    return concat(
+      [polkadot, kusama].filter(nonNullable),
+      sortBy(parachains, 'name'),
+      sortBy(numberchains, 'name'),
+      sortBy(testnets, 'name'),
+    );
+  };
+
+  const sortChainsByBalance = (chains: Chain[], balances: Balance[]): Chain[] => {
+    const dotsama = { withBalance: [], noBalance: [] };
```
Asmadek, line 73: maybe better to call it `relaychains`?

src/renderer/services/network/chainsService.ts
```diff
@@ -45,21 +51,67 @@ export function useChains(): IChainService {
...
-    return compact([polkadot, kusama, ...sortBy(parachains, 'name'), ...sortBy(testnets, 'name')]) as T[];
+    return concat(
+      [polkadot, kusama].filter(nonNullable),
+      sortBy(parachains, 'name'),
+      sortBy(numberchains, 'name'),
+      sortBy(testnets, 'name'),
+    );
+  };
+
+  const sortChainsByBalance = (chains: Chain[], balances: Balance[]): Chain[] => {
```
pgolovkin, line 72: Isn't it better to change the name? Sorting works not only by balances.

Asmadek (approved):
small suggestion

egor0798 (approved):

src/renderer/services/network/chainsService.ts
tuul-wq: Good suggestion but I'm not sure about that
Name is pretty long already and if I change it to something like `sortChainsByGroupsNamesAndBalance` 
It would bring more questions to reader of the code at least about groups 
If he needs more details, he will jump into implementation anyway
So I would choose a compromise with current name 😊

pgolovkin (approved):

PR #972, opened 2023-07-18
fix: decoded tx condition

pgolovkin:


src/renderer/domain/transaction.ts
```diff
@@ -94,9 +94,9 @@ export type MultisigTransactionKey = Pick<
 >;
 
 export function isDecodedTx(tx: Transaction | DecodedTransaction): tx is DecodedTransaction {
-  const hasType = 'type' in (tx as Transaction);
-  const hasMethod = 'method' in (tx as DecodedTransaction);
-  const hasSection = 'section' in (tx as DecodedTransaction);
+  const hasType = tx.type != undefined;
+  const hasMethod = (tx as DecodedTransaction).method != undefined;
+  const hasSection = (tx as DecodedTransaction).section != undefined;
```
Asmadek, line 99: better to use `!==`

If you want to check `falsy` values (not only undefined) you can use `Boolean(value)`

Asmadek (approved):
With comment

PR #979, opened 2023-07-20
feat: rolling log file appender

pgolovkin:
Enterprise-like rolling file appender that allows to save logs with date in the name
image

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -25,18 +25,8 @@ log.transports.file.archiveLogFn = (oldLogFile: LogFile): void => {
   const info = path.parse(file);
 
   try {
-    const date = new Date();
-    let newFileName = path.join(
-      info.dir,
-      info.name + '.' + date.getFullYear() + '-' + (date.getMonth() + 1) + '-' + date.getDate() + info.ext,
-    );
-    if (fs.existsSync(newFileName)) {
-      let i: number = 1;
-      do {
-        i++;
-      } while (fs.existsSync(newFileName + '-' + i));
-      newFileName = newFileName + '-' + i;
-    }
+    const date = new Date().toISOString();
+    let newFileName = path.join(info.dir, info.name + '.' + date + info.ext);
     fs.renameSync(file, newFileName);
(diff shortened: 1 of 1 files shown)


src/main/main.ts
```diff
@@ -17,7 +19,29 @@ log.transports.console.useStyles = true;
...
+    let newFileName = path.join(
+      info.dir,
+      info.name + '.' + date.getFullYear() + '-' + (date.getMonth() + 1) + '-' + date.getDate() + info.ext,
+    );
+    if (fs.existsSync(newFileName)) {
+      let i: number = 1;
+      do {
+        i++;
+      } while (fs.existsSync(newFileName + '-' + i));
+      newFileName = newFileName + '-' + i;
```
Asmadek, line 38: Is it Okay to add index to the end of the extension, not filename?

pgolovkin: It's ok. I've checked it in the article https://www.baeldung.com/java-logging-rolling-file-appenders
However it may be better to add dot instead of dash.

egor0798 (approved):

src/main/main.ts
```diff
@@ -17,7 +19,29 @@ log.transports.console.useStyles = true;
...
+  const file = oldLogFile.toString();
+  const info = path.parse(file);
+
+  try {
+    const date = new Date();
+    let newFileName = path.join(
+      info.dir,
+      info.name + '.' + date.getFullYear() + '-' + (date.getMonth() + 1) + '-' + date.getDate() + info.ext,
+    );
+    if (fs.existsSync(newFileName)) {
```
stepanLav, line 33: I would suggest to use:
const date = new Date().toISOString();
in order to avoid duplication problem

Screenshot 2023-07-26 at 11 01 12

pgolovkin: Looks better 
image

Asmadek (approved):

stepanLav (approved):

tuul-wq (approved):

PR #984, opened 2023-07-24
CI. Release workflow fix: add skip if file name wasn't changed

stepanLav:
That command behave different in linux github image which leads to problem:
`mv: 'latest-linux.yml' and 'latest-linux.yml' are the same file`

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -98,7 +98,9 @@ jobs:
           for file in *; do
             if [ -f "$file" ]; then
               new_name=$(echo "$file" | tr ' ' '-')
-              mv "$file" "$new_name"
+              if [ "$new_name" != "$file" ]; then # Skip files with no spaces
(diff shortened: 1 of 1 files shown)


pgolovkin (approved):

tuul-wq (approved):

.github/workflows/release.yaml
```diff
@@ -98,7 +98,9 @@ jobs:
           for file in *; do
             if [ -f "$file" ]; then
               new_name=$(echo "$file" | tr ' ' '-')
-              mv "$file" "$new_name"
+              if [ "$new_name" != "$file" ]; then # Skip files with no spaces
```
egor0798, line 101: Same result can be achieved with `-n` flag.
Docs: 
-n, --no-clobber
do not overwrite an existing file

egor0798 (approved):

PR #987, opened 2023-07-26
Feat: signature verify

Asmadek:
* Feat: signature verify
* fix: signing redesign


src/renderer/services/transaction/transactionService.ts
```diff
@@ -440,6 +442,12 @@ export const useTransaction = (): ITransactionService => {
       });
   };
 
+  const verifySignature = (payload: string | Uint8Array, signature: HexString, accountId: AccountId): Boolean => {
+    const result: VerifyResult = signatureVerify(payload, signature, accountId);
+
+    return result.isValid;
```
tuul-wq, line 448: Not sure about `result` variable, can use inline function call
```suggestion
    return signatureVerify(payload, signature, accountId).isValid;
```

tuul-wq (approved):

PR #1002, opened 2023-08-04
Feat: support max weight calc for unknown tx

Asmadek:
feat: support max weight calc for unknown tx
fix: use correct nonce definition 
fix: update some packages


src/renderer/screens/Operations/components/modals/ApproveTx.tsx
```diff
@@ -88,10 +90,11 @@ const ApproveTx = ({ tx, account, connection }: Props) => {
   }, [tx, signAccount?.accountId, txWeight]);
 
   useEffect(() => {
-    if (!tx.transaction || !connection.api) return;
```
valentunn, line 91: What does initial `tx.transaction` mean in this context? And why do we replace it with `getTxFromCallData(connection.api, tx.callData)`? Is that because `tx.transaction` might be invalid tx if method in calldata is unknown to spektr?

valentunn (approved):

src/renderer/services/transaction/common/constants.ts
```diff
@@ -1,5 +1,5 @@
 export const MAX_WEIGHT = {
-  refTime: '640000000',
+  refTime: '0',
```
valentunn, line 2: Are we sure we want to put `0` here?

egor0798 (approved):

src/renderer/screens/Operations/components/modals/ApproveTx.tsx
Asmadek: Yes, we don't fill transaction field for unknown operation. Because it have our inner format

PR #1015, opened 2023-08-15
Fix: UI exports

tuul-wq:
- Converted `import` + `export` to `re-export`


src/renderer/entities/chain/ui/ChainTitle/ChainTitle.tsx
```diff
@@ -32,9 +32,7 @@ export const ChainTitle = ({ as: Tag = 'div', fontClass, className, iconSize = 1
   return (
     <Tag className={cnTw('flex items-center gap-x-2', className)}>
       <ChainIcon src={chainObj?.icon} name={chainObj?.name} size={iconSize} />
-      <TextBase as="span" className={cnTw('text-text-tertiary text-footnote', fontClass)}>
```
tuul-wq, line 35: I assume `<TextBase />` should be used as a template component only inside `Typography`

Asmadek (approved):

PR #1016, opened 2023-08-15
Feat/opration UI redesign

egor0798:
## Description:
Aim of this refactoring is to encapsulate logic related to SingingType and WalletType in common components so we can add new signing type without affecting operation components and without having to rewtite every operation component when new SigninType is added
***
## Changes:
* created operation feature and sub features init and sign
* All Init operation components now composed of 3 common component - `OperationHeader`, `OperationFooter `and `OperationFrom`
    *  `OperationHeader` handles WalletType related UI and logic - selecting signatory, selecting shards
    *  `OperationForm` handles operation related form control and validation
    *  `OperationFooter` displays general operation information like network fee and multisig deposit
* Scaning step removed from all operations and now it's a part of Signing step
* Signing step now consists of top level signing component which renders specific flow based on SigningType. Currently only VaultSigning flow is present


src/renderer/pages/Staking/Operations/Bond/Bond.tsx
```diff
@@ -283,51 +275,16 @@ export const Bond = () => {
...
-              transactions.length > 1 ? txAccounts.map((t) => t.accountId) : [(signer || txAccounts[0])?.accountId]
-            }
-            txPayloads={txPayloads}
-            multiQr={transactions.length > 1}
+            chainId={chainId}
+            api={api}
+            addressPrefix={addressPrefix}
+            signatory={signer}
+            accounts={txAccounts}
+            transactions={multisigTx ? [multisigTx] : transactions}
```
ERussel, line 285: I think it is not scalable if we will need to separate multisig logic from non multisig logic on each flow. Normally the flow must be agnostic of this and as I remember we discussed wrapper pattern to address this issue.

src/renderer/entities/transaction/lib/validateBalance.ts
```diff
@@ -0,0 +1,62 @@
...
+  if (assetId === '0') return Promise.resolve(undefined);
+
+  return getBalance(toAccountId(transaction.address), chainId, '0');
+};
+
+const validateBalanceForAmount = async ({ transaction, ...props }: Props): Promise<boolean> => {
+  const amount = transaction.args.value;
+  const transferableBalance = transferableAmount(await getTokenBalance({ transaction, ...props }));
+
+  return new BN(transferableBalance).gt(new BN(amount));
```
ERussel, line 49: I think it must be ```gte``` not ```gt```

src/renderer/features/operation/init/ui/OperationHeader/OperationHeader.tsx
```diff
@@ -0,0 +1,68 @@
...
+          account={firstAccount as MultisigAccount}
+          invalid={invalid}
+          error={error}
+          getSignatoryOption={getSignatoryOption}
+          onSignatoryChange={onSignatoryChange}
+        />
+      )}
+
+      {isMultishard(firstAccount) &&
+        (isMultiselect ? (
```
ERussel, line 52: Not sure that I got the idea. Can we just name MultiSelectMultishard and SingleSelectMultishard instead of specifying the name of the feature where it is used Staking/Transfer? For example, we will add another feature: governance and it just reuses selection logic from Staking it will be bad idea to have MultishardStaking header for it.

src/renderer/pages/Staking/Operations/ChangeValidators/InitOperation/InitOperation.tsx
```diff
@@ -195,41 +161,25 @@ const InitOperation = ({ api, chainId, accounts, asset, addressPrefix, onResult
             onDepositChange={setDeposit}
           />
         }
+        header={
+          <OperationHeader
+            chainId={chainId}
+            accounts={accounts}
+            isMultiselect
+            invalid={accountIsMultisig ? !isValidDeposit : !isValidFee}
```
ERussel, line 169: Looks like this are general validations. Probably we can create include them into OperationHeader or create a helper function for them. The main point not to duplicate them in each module

src/renderer/features/operation/sign/ui/VaultSigning/VaultSigning.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+  const handleSignature = async (data: string | string[]) => {
+    const isMultishard = Array.isArray(data);
+    const signatures = isMultishard ? (data as HexString[]) : [data as HexString];
+    const accountIds = isMultiframe ? accounts.map((t) => t.accountId) : [(signatory || accounts[0])?.accountId];
+
+    const isVerified = signatures.every((signature, index) => {
+      // TODO: Research complex verification
+      // TODO: research multishard signature verification
+      if (isMultishard) return true;
```
ERussel, line 49: How did we verify in the previously?

src/renderer/pages/Transfer/components/TransferForm.tsx
```diff
@@ -383,36 +370,37 @@ export const TransferForm = ({
...
-                api={api}
-                asset={nativeToken}
-                transaction={transferTx}
-                onFeeLoading={setFeeLoading}
-                onFeeChange={updateFee}
-              />
-            )}
-          </DetailRow>
-        </div>
+        {/*<div className="flex flex-col items-center mt-2 gap-y-4">*/}
```
Asmadek, line 373: Do you need this commented code?

src/renderer/features/operation/sign/ui/VaultSigning/VaultSigning.tsx
Asmadek: Just verify every signature doesn't work in some cases, and we decided to temporarily disable it

Asmadek (changes requested):

tuul-wq (changes requested):
Great step in the right direction!

src/renderer/pages/Staking/Operations/Bond/Bond.tsx
egor0798: As I meantioned before I agree with you but I think it would be better to revisit that problen when we finish transaction service refactoring and implementing state manager for operation flow

src/renderer/features/operation/init/model/errors.ts
```diff
@@ -0,0 +1,9 @@
+import { ObjectValues } from '@renderer/domain/utility';
+
+export const OperationError = {
+  INVALID_FEE: 'staking.notEnoughBalanceForFeeError',
+  INVALID_DEPOSIT: 'staking.notEnoughBalanceForDepositError',
+  EMPTY_ERROR: ' ', // error for invalid state with no hint text
```
Asmadek, line 6: Why empty error is `' '` not empty string?

src/renderer/entities/transaction/lib/validateBalance.ts
```diff
@@ -0,0 +1,91 @@
...
+};
+
+export const getOperationErrors = (
+  isFeeInvalid: boolean,
+  isDepositInvalid: boolean,
+  hasOtherErrors?: boolean,
+): OperationErrorType[] => {
+  console.log('isFeeInvalid', isFeeInvalid);
+  console.log('isDepositInvalid', isDepositInvalid);
+  console.log('hasOtherErrors', hasOtherErrors);
```
Asmadek, line 75: Do you need this console log?

Asmadek (approved):
few comments

ERussel (approved):

tuul-wq (approved):

PR #1020, opened 2023-08-17
Feat: Alert redesign

tuul-wq:
- Update styles for `<Alert />`
- Removed unused tailwind styles
- Added custom padding for cases where we must include border into overall padding value
```javascript
theme: {
    extend: {
        padding: {
          '2b': '7px',
          '3b': '11px',
          '4b': '15px',
          '5b': '19px',
      },
    }
}
```

[Task](https://app.clickup.com/t/85ztnq3at)


tailwind.config.ts
```diff
@@ -83,6 +53,12 @@ export default {
         15: '3.75rem',
       },
       fontSize: fontSizes,
+      padding: {
+        '2b': '7px',
```
Asmadek, line 57: we also have `6px` padding, please add it too

Asmadek (approved):

PR #1021, opened 2023-08-17
Feat: new colors and shadows

Asmadek:


tailwind.config.ts
```diff
@@ -16,57 +16,10 @@ export default {
...
-          shade: {
-            8: 'rgba(0, 0, 0, 0.08)',
-            12: 'rgba(0, 0, 0, 0.12)',
-            32: 'rgba(0, 0, 0, 0.32)',
-            40: 'rgba(0, 0, 0, 0.40)',
-            48: 'rgba(0, 0, 0, 0.48)',
-          },
-        },
-        ...colors,
+        'shadow-primary': '0 4px 4px rgba(69, 69, 137, 0.12), 0 -0.5px 0 rgba(69, 69, 137, 0.04)',
```
tuul-wq, line 19: Why you decided to leave shadows here and not inside `default.css`?

tuul-wq (approved):
I'm so glad that we're going to have less colors 😎

PR #1027, opened 2023-08-21
feat: coingecko price adapter

Asmadek:
* Create base for fiat price feature
* Add coingecko price adapter
* Dev documentation for feature

### DIFF
--- a/docs/fiat-price.md
+++ b/docs/fiat-price.md
@@ -1,0 +1,55 @@
+# Fiat price
+
+Now fiat price feature consists of `PriceAdapter` type. Each adapter should implement two functions 1. to return price for selected crypto assets with selected currencies (fiat or crypto) and 2. return history data for selected asset, currency and time range. 
+
+```typescript
+  getPrice: (ids: AssetId[], currencies: Currency[], includeRateChange: boolean) => Promise<PriceObject>;
+  getHistoryData: (id: AssetId, currency: Currency, from: number, to: number) => Promise<PriceRange[]>;
+```
+
+And we have two util functions, which can convert `PriceObject` format to `PriceDB` format.
+
+`PriceObject` is based on coingecko format.
+`PriceDB` is based on IndexedDB storage object.
+
+```typescript
+type PriceObject = Record<AssetId, AssetPrice>;
+type AssetPrice = Record<Currency, PriceItem>;
+type PriceItem = {
+  price: number;
+  change: number;
+};
+```
+
+example 
+
+```typescript
+const priceObject = {
+  polkadot: {
+    usd: {
+      price: 4.1
+      change: -0.1
+    }
+  }
+}
+```
+
+```typescript
+type PriceDB = {
+  assetId: AssetId;
+  currency: Currency;
+  price: number;
+  change: number;
+};
+```
+
+example 
+
+```typescript
+const priceArray = [{
+  assetId: 'polkadot',
+  currency: 'usd',
+  price: 4.1,
+  change: -0.1
+}]
+```
--- a/src/renderer/shared/api/price/coingecko/CoingeckoAdapter.ts
+++ b/src/renderer/shared/api/price/coingecko/CoingeckoAdapter.ts
@@ -1,0 +1,48 @@
+import { AssetId, Currency, PriceObject, PriceAdapter, PriceItem, PriceRange } from '../common/types';
+import { getCurrencyChangeTitle } from '../common/utils';
+import { COINGECKO_URL } from './consts';
+
+export class CoinGeckoAdapter implements PriceAdapter {
+  async getPrice(ids: AssetId[], currencies: Currency[], includeRateChange: boolean): Promise<PriceObject> {
+    const response = await fetch(`${COINGECKO_URL}/simple/price`, {
+      method: 'GET',
+      body: JSON.stringify({
+        ids: ids.join(','),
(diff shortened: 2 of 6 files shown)


docs/fiat-price.md
```diff
@@ -0,0 +1,55 @@
...
+      price: 4.1
+      change: -0.1
+    }
+  }
+}
+```
+
+```typescript
+type PriceDB = {
+  assetId: AssetId;
```
ERussel, line 39: I would name this priceAssetId to not mix up with the assetId in the chain.json.

src/renderer/shared/api/price/coingecko/CoingeckoAdapter.ts
```diff
@@ -0,0 +1,48 @@
...
+      }),
+    });
+
+    const data = await response.json();
+
+    const result = ids.reduce<PriceObject>((result, assetId) => {
+      result[assetId] = currencies.reduce<Record<string, PriceItem>>((result, currency) => {
+        result[currency] = {
+          price: data[assetId][currency],
+          change: data[assetId][getCurrencyChangeTitle(currency)],
```
ERussel, line 22: ```getCurrencyChangeKey``` suits better I think

egor0798 (approved):

docs/fiat-price.md
pgolovkin: I'm not sure that the attribute name should be changed. It's a part of the type `PriceDB` and won't be mixed up with chains.json.

src/renderer/shared/api/price/coingecko/CoingeckoAdapter.ts
```diff
@@ -0,0 +1,48 @@
+import { AssetId, Currency, PriceObject, PriceAdapter, PriceItem, PriceRange } from '../common/types';
+import { getCurrencyChangeTitle } from '../common/utils';
+import { COINGECKO_URL } from './consts';
+
+export class CoinGeckoAdapter implements PriceAdapter {
```
tuul-wq, line 5: `PriceAdapter ` is a `Type`
It's more correct to make it an `interface`

By the way, why you decided to use a class? 
Right now (and I hope it will be the same in the future) it doesn't store any state
and provides only 2 pure functions
You didn't want to go `hook style` like we did in services? Or there is another reason? 🤔

src/renderer/shared/api/price/common/types.ts
```diff
@@ -0,0 +1,21 @@
...
+export type PriceObject = Record<AssetId, AssetPrice>;
+export type PriceRange = [number, string];
+export type PriceDB = {
+  assetId: AssetId;
+  currency: Currency;
+  price: number;
+  change: number;
+};
+
+export type PriceAdapter = {
```
tuul-wq, line 18: I assume that `Fiat` suffix can narrow `Price` meaning to our specific case
`FiatPriceAdapter`, `FiatPriceObject` etc.

src/renderer/shared/api/price/coingecko/CoingeckoAdapter.ts
```diff
@@ -0,0 +1,48 @@
...
+      body: JSON.stringify({
+        ids: ids.join(','),
+        vs_currencies: currencies.join(','),
+        include_24hr_change: includeRateChange ? 'true' : 'false',
+      }),
+    });
+
+    const data = await response.json();
+
+    const result = ids.reduce<PriceObject>((result, assetId) => {
```
tuul-wq, line 20: Added minor readability

```suggestion
    return ids.reduce<PriceObject>((acc, assetId) => {
      acc[assetId] = currencies.reduce<Record<string, PriceItem>>((result, currency) => {
        result[currency] = {
          price: data[assetId][currency],
          change: data[assetId][getCurrencyChangeTitle(currency)],
        };

        return result;
      }, {});

      return acc;
    }, {});
```

tuul-wq (changes requested):

src/renderer/shared/api/price/coingecko/CoingeckoAdapter.test.ts
tuul-wq, line 1: We should not rely on real API because it might change or return unexpected data
Need to return mocked data from `fetch` for each test suite

src/renderer/shared/api/price/coingecko/CoingeckoAdapter.ts
```diff
@@ -0,0 +1,52 @@
+import { AssetId, Currency, PriceObject, PriceAdapter, PriceItem, PriceRange } from '../common/types';
+import { getCurrencyChangeKey } from '../common/utils';
+import { COINGECKO_URL } from './consts';
+
+export class CoinGeckoAdapter implements PriceAdapter {
+  async getPrice(ids: AssetId[], currencies: Currency[], includeRateChange: boolean): Promise<PriceObject> {
+    const url = new URL(`${COINGECKO_URL}/simple/price`);
```
tuul-wq, line 7: Any benefits in using `URL` instead of `fetch` with params?

tuul-wq (changes requested):

src/renderer/shared/api/price/common/types.ts
Asmadek: It can be not fiat, you can use crypto assets for price

tuul-wq (approved):

PR #1029, opened 2023-08-21
Wallet connect research

Asmadek:


src/main/main.ts
```diff
@@ -33,7 +33,7 @@ log.transports.file.archiveLogFn = (oldLogFile: LogFile): void => {
   }
 };
 
-Object.assign(console, log.functions);
+// Object.assign(console, log.functions);
```
pgolovkin, line 36: it shouldn't be changed.

pgolovkin (changes requested):
Good draft implementation! Left a comment.

pgolovkin:
I suggest to convert his PR to the draft and back to it when will be working on the Wallet Connect. We can base the solution on it or use as a reference.

PR #1034, opened 2023-08-23
Fix: Storybook

tuul-wq:
- Storybook update to version 7 (swc)
- Updated all stories to support v7
- Fixed `Alert`, `StatusMark` styles according to new colors

image

[Task](https://app.clickup.com/t/862k9kvwr)


src/renderer/shared/ui/Buttons/ButtonIcon/common/constants.ts
```diff
@@ -9,8 +11,12 @@ export const SizeStyleBG: Record<'sm' | 'md', string> = {
 };
 
 export const ButtonStyle = 'rounded';
-export const ButtonStyleBG = 'rounded-full transition-colors ' + 'hover:bg-bg-primary-hover active:bg-bg-primary-hover';
+export const ButtonStyleBG = cnTw(
```
Asmadek, line 14: What do you think to rename it to `ButtonStyleWithBG` to better understand that it's not only button background 

`SizeStyleBG` too

Asmadek (approved):

PR #1038, opened 2023-08-25
feat: update metadata form chain

Asmadek:


src/renderer/entities/multisig/lib/multisigTx/common/consts.ts
```diff
@@ -1 +1,2 @@
 export const QUERY_INTERVAL = 5000;
+export const MULTISIG_EXTRINSIC_CALL_INDEX = 3;
```
pgolovkin, line 2: Is it the same for all chains? For example `Edgeware` multisig pallet is old one and the index may be different.

Asmadek: Yes, it's always has such index

pgolovkin (approved):

tuul-wq (approved):

PR #1039, opened 2023-08-28
feat: cache metadata

Asmadek:
Implements cache metadata [research](https://github.com/novasamatech/research-docs/blob/main/nova-spektr/metadata-cache/metadata-cache.md)


tuul-wq (approved):

src/renderer/entities/metadata/lib/metadataService.ts
```diff
@@ -0,0 +1,62 @@
...
+      const lastMetadata = metadata.reduce<Metadata>((acc, md) => {
+        if (md.version >= (acc.version || -1)) {
+          return md;
+        }
+
+        return acc;
+      }, {} as Metadata);
+
+      return lastMetadata;
+    }
```
valentunn, line 31: What are the benenefits of retaining all previously fetched versions in metadata in the cache? As I can see, we only allow to fetch current version of metadata from the cache, meaning all outdated version will effectively become useless. I suggest deleting all previous versions once we synced new one

src/renderer/entities/network/lib/provider/WrappedScProvider.ts
```diff
@@ -0,0 +1,19 @@
...
+        const metadata = await getMetadata();
+        if (metadata) return metadata;
+      }
+
+      return super.send(method, params);
+    }
+  }
+
+  return WrappedScProvider;
+};
```
valentunn, line 19: Why did we decide to step away from solution described in the research docs?

Initital plan was to have a single wrapped provider, lets say `MetadataCachingProvider`, which will accept an instance of any other `Provider` in its constructor and will delegate all calls except GET_METADATA_METHOD to this instance.

This way, we dont have to create two similar implementations (`createWrappedScProvider ` and `createWrappedWsProvider `) but rather have a single wrapper that can be combined with any other `Provider` implementation

src/renderer/entities/network/lib/networkService.ts
```diff
@@ -285,6 +305,12 @@ export const useNetwork = (networkSubscription?: ISubscriptionService<ChainId>):
 
   const validateRpcNode = (chainId: ChainId, rpcUrl: string): Promise<RpcValidation> => {
     return new Promise((resolve) => {
+      const WsProvider = createWrappedWsProvider(async () => {
+        const metadata = await getMetadata(chainId);
+
+        return metadata?.metadata;
+      });
```
valentunn, line 312: This seems to be a dublicate of code piece inside `createWebsocketProvider `

src/renderer/entities/network/lib/networkService.ts
```diff
@@ -150,14 +154,26 @@ export const useNetwork = (networkSubscription?: ISubscriptionService<ChainId>):
     const knownChainId = getKnownChain(chainId);
 
     if (knownChainId) {
+      const ScProvider = createWrappedScProvider(async () => {
+        const metadata = await getMetadata(chainId);
+
+        return metadata?.metadata;
+      });
```
valentunn, line 161: Closure passed to `createWrappedScProvider` seems to be the same in all 3 places provider is created. Maybe worth to make it part of WrappedProvider itself? For example. we could pass `chainId` and `getMetadata/IMetadataService` to wrapper provider constructor

valentunn (changes requested):

src/renderer/entities/metadata/lib/metadataService.ts
Asmadek: This solution will helps us to decode callData for transactions created with old metadata.

valentunn: So you are planning to use `IMetadataService` directly when decoding, right? Because requests to metadata made via ApiPromise wont use cache due to `params.length == 0` check

Asmadek: Right, metadata service will be used directly for it. But we have to research how to detect correct version

valentunn (approved):

egor0798 (approved):

PR #1043, opened 2023-08-31
Feat: label redesign

Asmadek:


src/renderer/shared/ui/Label/Label.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { LabelStyle, LabelTextStyle, LabelPallet } from './constants';
+
+type Props = {
+  className?: string;
+  pallet?: LabelPallet;
+};
+
+export const Label = ({ className, children, pallet = 'default' }: PropsWithChildren<Props>) => (
+  <div className={cnTw('rounded-lg py-0.5 px-2 group w-fit', LabelStyle[pallet], className)} data-testid="label">
+    <CaptionText className={cnTw('uppercase', LabelTextStyle[pallet])}>{children}</CaptionText>
```
tuul-wq, line 14: I assume it's better to declare `children: string`, so nobody provides `ReactNode` in here
Like you did in `<BlockExplorer />` 👍

src/renderer/shared/ui/LabelHelpbox/LabelHelpBox.tsx
```diff
@@ -1,22 +1,29 @@
...
 import { cnTw } from '@renderer/shared/lib/utils';
 
 type Props = {
   className?: string;
+  disabled?: boolean;
 };
 
-export const LabelHelpBox = ({ className, children }: PropsWithChildren<Props>) => (
-  <div
+export const LabelHelpBox = ({ className, children, disabled = false, ...props }: PropsWithChildren<Props>) => (
```
tuul-wq, line 11: What kind of props might come here? Without explicit declaration in `Props` it's not clear
Maybe it's not needed at all?
Also I would add `children: string` as well 😉

src/renderer/shared/ui/LabelHelpbox/LabelHelpBox.tsx
```diff
@@ -1,22 +1,29 @@
...
       className,
     )}
+    disabled={disabled}
     data-testid="labelHelpbox"
+    {...props}
   >
-    <BodyText>{children}</BodyText>
-    <Icon name="learn-more" className="group-hover:text-icon-hover group-active:text-icon-active" size={16} />
-  </div>
+    <FootnoteText className={cnTw('text-text-primary', disabled && 'text-text-tertiary')}>{children}</FootnoteText>
```
tuul-wq, line 22: This is optional, could leave as you've done
```suggestion
    <FootnoteText className={disabled ? 'text-text-tertiary' : 'text-text-primary'}>{children}</FootnoteText>
```

src/renderer/shared/ui/Popovers/Tooltip/Tooltip.tsx
```diff
@@ -5,23 +5,37 @@ import { cnTw } from '@renderer/shared/lib/utils';
 import { HelpText } from '../../Typography';
 import './Tooltip.css';
 
-type PopoverProps = ComponentProps<typeof Popover> & { pointer?: 'up' | 'down' };
+type PointerDirection = 'up' | 'down';
+type PointerPosition = 'start' | 'center' | 'end';
+type PopoverProps = ComponentProps<typeof Popover> & {
```
tuul-wq, line 10: Maybe just `Props`?

tuul-wq (changes requested):
Nice!

PR #1044, opened 2023-09-01
Feat: Cross chain config

tuul-wq:
- Load cross-chain config from github on Transfer modal mount
- Use data from storage before we get newest version from the network
- Tests for service and effector model

[Task](https://app.clickup.com/t/862katqmx)


src/renderer/shared/api/cross-chain/common/types.ts
```diff
@@ -0,0 +1,75 @@
...
+      mode: FeeMode;
+      instructions: keyof Instructions;
+    };
+  };
+};
+
+type XcmTransferType = 'xtokens' | 'xcmpallet';
+type PathType = 'absolute' | 'relative';
+
+// const enum Action {
```
tuul-wq, line 67: @Asmadek If this is not helpful for your task, I will delete it

src/renderer/shared/api/cross-chain/crossChainService.ts
```diff
@@ -0,0 +1,24 @@
+import { XCM_URL, XCM_KEY } from './common/constants';
+import { XcmConfig } from './common/types';
+
+export const fetchXcmConfig = async (): Promise<XcmConfig> => {
```
tuul-wq, line 4: Used js module instead of hook-like service to eliminate redundant function creation during React renders
This also makes mocking in tests much easier

src/renderer/widgets/SendAssetModal/model/send-asset.ts
```diff
@@ -0,0 +1,46 @@
+import { createStore, createEffect, createEvent, sample, forward } from 'effector';
+
+import { XcmConfig, fetchXcmConfig, getXcmConfig, saveXcmConfig } from '@renderer/shared/api/cross-chain';
+
+export const $finalConfig = createStore<any | null>(null);
```
tuul-wq, line 5: Placeholder name for the "final config" that's going to be used in UI

tuul-wq:
Task linked: [CU-862katqmx Cross-chain transfer config reading and parsing](https://app.clickup.com/t/862katqmx)

Asmadek (approved):

egor0798 (approved):

PR #1045, opened 2023-09-04
Fix: added env variable for updater

egor0798:


src/main/index.ts
```diff
@@ -1,12 +1,18 @@
...
+import * as process from 'process';
 
 import { MainWindow } from './main';
 import { makeAppWithSingleInstanceLock } from './factories/instance';
 import { makeAppSetup } from './factories/setup';
 
-makeAppWithSingleInstanceLock(async () => {
-  app.commandLine.appendSwitch('autoplay-policy', 'no-user-gesture-required');
+const setupAutoUpdate = () => {
+  console.log('process.env.BUILD_SOURCE', process.env.BUILD_SOURCE);
```
Asmadek, line 10: Remove console log or make it more meaningful

Asmadek (approved):

src/main/index.ts
```diff
@@ -1,12 +1,16 @@
...
+import * as process from 'process';
 
 import { MainWindow } from './main';
 import { makeAppWithSingleInstanceLock } from './factories/instance';
 import { makeAppSetup } from './factories/setup';
 
-makeAppWithSingleInstanceLock(async () => {
-  app.commandLine.appendSwitch('autoplay-policy', 'no-user-gesture-required');
+const setupAutoUpdate = () => {
+  if (process.env.BUILD_SOURCE !== 'github') {
```
tuul-wq, line 10: We usually make fast return this way 😉

```suggestion
  if (process.env.BUILD_SOURCE !== 'github') return;
```

tuul-wq (approved):
Approve with 1 comment

.github/workflows/release.yaml
```diff
@@ -8,6 +8,7 @@ on:
 env:
   CSC_FOR_PULL_REQUEST: true
   CI: true
+  BUILD_SOURCE: 'github'
```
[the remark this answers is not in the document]
egor0798: yes, that's enough. It's added here cause it shoud be set only with CI

stepanLav (approved):

PR #1048, opened 2023-09-05
Feat: cross chain service

Asmadek:


src/renderer/shared/api/cross-chain/crossChainService.ts
```diff
@@ -22,3 +40,252 @@ export const getXcmConfig = (): XcmConfig | null => {
 export const saveXcmConfig = (config: XcmConfig) => {
   localStorage.setItem(XCM_KEY, JSON.stringify(config));
 };
+
+export const getAvailableDirections = (chains: ChainXCM[], assetId: number, chainId: ChainId): XcmTransfer[] => {
+  const chain = chains.find((c) => c.chainId === chainId);
```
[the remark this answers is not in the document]
Asmadek: I think it's better to implement this feature on higher level, where we will have connection status, and here work only with xcm config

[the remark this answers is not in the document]
pgolovkin: In that case I think that a higher-level function for getting available Direction has to be implemented. Otherwise this method will be used directly that may lead to errors.

src/renderer/shared/api/cross-chain/crossChainService.ts
```diff
@@ -22,3 +40,252 @@ export const getXcmConfig = (): XcmConfig | null => {
...
+    return AbsoluteAssetLocation(api, assets[asset.assetLocation].multiLocation, amount);
+  }
+  if (asset.assetLocationPath.type === 'concrete') {
+    return ConcreteAssetLocation(api, asset.assetLocationPath.path, amount);
+  }
+};
+
+const RelativeAssetLocation = (
+  api: ApiPromise,
+  assetLocation: LocalMultiLocation | undefined,
```
tuul-wq, line 148: `assetLocation?: LocalMultiLocation` suppose it works fine 😎

src/renderer/shared/api/cross-chain/crossChainService.ts
```diff
@@ -22,3 +40,252 @@ export const getXcmConfig = (): XcmConfig | null => {
...
+
+  if (asset.assetLocationPath.type === 'absolute') {
+    return AbsoluteAssetLocation(api, assets[asset.assetLocation].multiLocation, amount);
+  }
+  if (asset.assetLocationPath.type === 'concrete') {
+    return ConcreteAssetLocation(api, asset.assetLocationPath.path, amount);
+  }
+};
+
+const RelativeAssetLocation = (
```
tuul-wq, line 146: Better stick to current styling - `camelCase`

tuul-wq (changes requested):
Nice 🔥
Left a couple of comments to improve the service

src/renderer/shared/api/cross-chain/common/testConfig.ts
tuul-wq, line 1: I think we can move this config to `/common/__tests__/mock` 
in order to make it stay closer to place it's being used in
and eliminate unwanted export

tuul-wq:
Task linked: [CU-862kbacfp Cross-chain service](https://app.clickup.com/t/862kbacfp)

PR #1053, opened 2023-09-07
Feat: Transaction service refactoring

egor0798:
## Resources:
* [Transaction Service sequnce diagramm](https://app.diagrams.net/#G1-0zUvBHA_di3J-nytr2D7H57eRyyGERl#%7B%22pageId%22%3A%22gk_CA3zQrX1POXpU8ZN8%22%7D)
## Changes:
* Created ExtrinsicService and moved extrinsic creation logic from transaction service
* Made transaction service statefull so UI would not be responsible for creating transaction
* Removed multisig transaction creation logic from UI
* Integrated new transaction service to transfer flow


src/renderer/widgets/SendAssetModal/ui/SendAssetModal.tsx
```diff
@@ -30,39 +30,38 @@ type Props = {
...
   const [signatory, setSignatory] = useState<Account>();
   const [description, setDescription] = useState('');
-  const [transferTx, setTransferTx] = useState<Transaction>({} as Transaction);
-  const [multisigTx, setMultisigTx] = useState<Transaction>();
   const [unsignedTx, setUnsignedTx] = useState<UnsignedTransaction>({} as UnsignedTransaction);
   const [signature, setSignature] = useState<HexString>('0x0');
 
   const connection = connections[chain.chainId];
-  const transaction = multisigTx || transferTx;
+  // const transaction = multisigTx || transferTx;
```
tuul-wq, line 45: I assume this can be deleted

src/renderer/widgets/SendAssetModal/ui/SendAssetModal.tsx
```diff
@@ -30,39 +30,38 @@ type Props = {
...
+  const transaction = txs[0];
+
   const checkBalance = () =>
     validateBalance({
       api,
       transaction,
       chainId: chain.chainId,
       assetId: asset.assetId.toString(),
       getBalance,
-      getTransactionFee,
```
tuul-wq, line 65: It really doesn't work `getTransactionFee` does it?

src/renderer/widgets/SendAssetModal/ui/components/ActionSteps/Submit.tsx
```diff
@@ -79,43 +77,17 @@ export const Submit = ({ api, tx, multisigTx, account, unsignedTx, signature, de
...
-
-          await Promise.all([addMultisigTx(newTx), addEventWithQueue(event)]);
-
-          console.log(`New transfer was created with call hash ${newTx.callHash}`);
-
-          if (matrix.userIsLoggedIn) {
-            sendMultisigEvent(account.matrixRoomId, newTx, typedParams);
+          const result = buildMultisigTx(tx, multisigTx, params as ExtrinsicResultParams, account, description);
+
+          if (result) {
```
tuul-wq, line 83: `result` can be `undefined`?
If so we'd rather log that in console

tuul-wq (changes requested):
Code has become much cleaner, nice!

src/renderer/entities/multisig/lib/multisigTx/common/utils.ts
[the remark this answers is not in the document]
egor0798: we have separation between multisigTx lib and multisigEvent lib. Do you think we should merge them?
image

src/renderer/entities/transaction/lib/extrinsicService.ts
```diff
@@ -0,0 +1,322 @@
...
+      return methods.assets.transfer(
+        {
+          id: transaction.args.asset,
+          target: transaction.args.dest,
+          amount: transaction.args.value,
+        },
+        info,
+        options,
+      );
+    },
```
valentunn, line 86: Why do we still have all case-by-case logic here? The main goal of ExtrinsicService was to allow each flow to setup up a call or a setup of calls by its own to allow for more flexibility? I remember we decided to postpone the task for initial refactoring (which is already merged)

src/renderer/pages/Staking/Operations/Bond/Bond.tsx
```diff
@@ -146,73 +144,34 @@ export const Bond = () => {
...
-        args: { targets: validators },
-      };
+      const nominateTx = buildTransaction(TransactionType.NOMINATE, address, chainId, { targets: validators });
 
-      return {
-        ...commonPayload,
-        type: TransactionType.BATCH_ALL,
-        args: { transactions: [bondTx, nominateTx] },
-      };
+      return buildTransaction(TransactionType.BATCH_ALL, address, chainId, { transactions: [bondTx, nominateTx] });
```
valentunn, line 171: This should be done automatically by ExtrinsicService

src/renderer/pages/Staking/Operations/ChangeValidators/ChangeValidators.tsx
```diff
@@ -192,6 +161,7 @@ export const ChangeValidators = () => {
   };
 
   const explorersProps = { explorers, addressPrefix, asset };
+  const multisigTx = isMultisig(txAccounts[0]) ? wrapTx(txs[0], api, addressPrefix) : undefined;
```
valentunn, line 164: Why do we still have `isMultisig` in concrete flows? The primary goal was to abstract it away

valentunn (changes requested):
PR doesnt follow discussed architecture and doesnt solve problems that we stated as a DOD for this task

valentunn:
Please consider the following DOD for this PR:

**No single screen should have `if(multisig)` logic**. Otherwise whole refactoring becomes uselles

src/renderer/entities/transaction/lib/extrinsicService.ts
ERussel: Agree here. As the diagram in the pull request shows the ExtrinsicService must depend on ExtrinsicBuilder which helps to built extrinsics but still agnostic from the concrete call being used. Only modules/logic which creates an extrinsic based on user input should get the builder and set call/module name and params.

src/renderer/entities/transaction/lib/transactionService.ts
```diff
@@ -445,6 +184,30 @@ export const useTransaction = (): ITransactionService => {
     return signatureVerify(payloadToVerify, signature, accountId).isValid;
   };
 
+  const wrapTx = (transaction: Transaction, api: ApiPromise, addressPrefix: number) => {
+    wrappers.forEach((wrapper) => {
+      if ('signatoryId' in wrapper) {
```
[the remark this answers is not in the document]
ERussel: I would even go further. We can assume that each wrapper has wrap function and then just call this function. Then the wrapper decides whether it must do something to transaction or return it as it is.

src/renderer/pages/Staking/Operations/Bond/Bond.tsx
```diff
@@ -146,73 +144,34 @@ export const Bond = () => {
     const transactions = getBondTxs(Object.keys(validators));
 
     if (signer && isMultisig(txAccounts[0])) {
-      const multisigTx = getMultisigTx(txAccounts[0], signer.accountId, transactions[0]);
-      setMultisigTx(multisigTx);
+      const wrapAsMulti = {
```
ERussel, line 147: It is not actually a Wrapper. It is just a type that contain parameters for concrete wrapper. The Wrapper is just a function (can be carried if it needs additional parameters) that accepts Transaction and return Transaction

src/renderer/pages/Staking/Operations/Bond/Bond.tsx
```diff
@@ -146,73 +144,34 @@ export const Bond = () => {
...
-      const bondTx = {
-        ...commonPayload,
-        type: TransactionType.BOND,
-        args: {
-          value: stakeAmount,
-          controller: address,
-          payee: destination?.type === RewardsDestination.TRANSFERABLE ? { Account: destination.address } : 'Staked',
-        },
-      };
+      const bondTx = buildTransaction(TransactionType.BOND, address, chainId, {
```
ERussel, line 163: Continuing discussion of ExtrinsicBuilder started in the ExtrinsicService, we can take this call as an example.

Instead of passing the parameters to the Extrinsic Service and create a huge switch case there we can apply inversion of control. That is being, instead buildTransaction must accept a closure with ExtrinsicBuilder parameter and each module like Bond will use extrinsic builder to set call and parameters.

This will allow calls logic incapsulation inside module they are being used in. If we need such transaction logic shared between modules then we can put it in a sharing function. The second point is that we get rid of switch...case in the ExtrinsicService and make it transaction agnostic which allows him concentrate only on general transaction parameters such as mortality, genesis hash and so on.

src/renderer/pages/Staking/Operations/ChangeValidators/ChangeValidators.tsx
ERussel: Agree here. As I understand it must be in the OperationHeader. The idea was that if an operation has a wrapper (such as multisig) then it must contain OperationHeader in the tree that wraps the final transaction object to introduce additional logic to transaction. Here as we can see we do this in the ui for the call logic.

ERussel (changes requested):

src/renderer/widgets/SendAssetModal/ui/SendAssetModal.tsx
egor0798: Yeah, cause we need to pass transaction object

src/renderer/entities/multisig/lib/multisigTx/common/utils.ts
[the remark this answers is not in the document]
tuul-wq: Now I see, so it can stay as it is right now

tuul-wq (approved):

Asmadek (approved):

ERussel (approved):

valentunn (approved):

PR #1055, opened 2023-09-08
feat: xcm transfer flow

Asmadek:


src/renderer/shared/api/xcm/xcmService.ts
```diff
@@ -143,94 +181,115 @@ const getRelativeAssetLocation = (
...
-        Fungible: amount.toNumber(),
-      },
-    },
-  });
+    ],
+  };
 };
 
 export const getDestinationLocation = (
-  api: ApiPromise,
```
tuul-wq, line 213: 👍

src/renderer/widgets/SendAssetModal/ui/components/TransferForm.tsx
```diff
@@ -60,9 +74,18 @@ export const TransferForm = ({
   onSubmit,
   onChangeAmount,
   onTxChange,
+  onDestinationChainChange,
+  onDestinationChange,
   feeIsLoading,
   fee,
   deposit,
+  destinations,
+  xcmFee,
```
tuul-wq, line 83: Maybe it's a bit better to use a single `xcmParams` instead of many xcm params one by one?

src/renderer/widgets/SendAssetModal/ui/components/TransferForm.tsx
```diff
@@ -77,20 +100,33 @@ export const TransferForm = ({
...
+    setValue,
     formState: { isValid, isDirty },
   } = useForm<TransferFormData>({
     mode: 'onChange',
     defaultValues: { amount: '', destination: '', signatory: '', description: '' },
   });
 
+  useEffect(() => {
+    const destinationOptions = destinations.map(getChainOption);
+    destinationOptions.length && setDestinationOptions(destinations.map(getChainOption));
```
tuul-wq, line 119: Let's use verbose condition `if (...) { }`

src/renderer/widgets/SendAssetModal/ui/components/TransferForm.tsx
```diff
@@ -77,20 +100,33 @@ export const TransferForm = ({
...
+    setValue('destinationChain', destinationOptions[0]);
+  }, [destinations.length]);
+
   const amount = watch('amount');
   const destination = watch('destination');
+  const destinationChain = watch('destinationChain');
+
+  useEffect(() => {
+    destinationChain && onDestinationChainChange(destinationChain.value);
```
tuul-wq, line 128: Let's use verbose condition `if (...) { }` here as well

src/renderer/widgets/SendAssetModal/ui/components/TransferForm.tsx
```diff
@@ -142,36 +182,72 @@ export const TransferForm = ({
...
+
+      setMultisigTx(multisigTx);
     }
 
     setTransferTx(transferPayload);
     onTxChange(transferPayload);
-  }, [account, signer, destination, amount]);
+  }, [account, signer, destination, amount, destinationChain, xcmFee, xcmAsset, xcmDest, isXcmTransfer]);
+
+  const getXcmTransferType = (type: XcmTransferType) => {
```
tuul-wq, line 203:
```suggestion
  const getXcmTransferType = (type: XcmTransferType): TransactionType => {
```

src/renderer/widgets/SendAssetModal/model/send-asset.ts
```diff
@@ -1,10 +1,47 @@
...
+  AssetXCM,
+  getDestinationLocation,
+  getAssetLocation,
+  getAccountLocation,
+  estimateRequiredDestWeight,
+} from '@renderer/shared/api/xcm';
 import { xcmModel } from '@renderer/entities/xcm';
+import { Chain } from '@renderer/entities/chain';
+import { Asset } from '@renderer/entities/asset';
+import * as assetGuardModel from '@renderer/features/assets/AssetRouteGuard/model/asset-guard';
```
tuul-wq, line 18: As I see now, `asset-guard` model was not meant to be used outside of it's feature, so we need to remove it from reexports

tuul-wq: And declare a local event to indicate that flow has started

src/renderer/widgets/SendAssetModal/model/send-asset.ts
```diff
@@ -15,17 +52,175 @@ const getConfigFx = attach({ effect: xcmModel.effects.getConfigFx });
 const saveConfigFx = attach({ effect: xcmModel.effects.saveConfigFx });
 const fetchConfigFx = attach({ effect: xcmModel.effects.fetchConfigFx });
 
+const getParaIdFx = createEffect(async (api: ApiPromise): Promise<number | null> => {
+  try {
+    return await getParachainId(api);
```
tuul-wq, line 57: When returning Promise `await` is not necessary
```suggestion
    return getParachainId(api);
```

tuul-wq (changes requested):

src/renderer/shared/api/xcm/xcmService.ts
```diff
@@ -143,94 +181,115 @@ const getRelativeAssetLocation = (
...
 };
 
-const getChildLocation = (api: ApiPromise, parachainId: number, accountId?: AccountId): VersionedMultiLocation => {
+export const getAccountLocation = (accountId?: AccountId): Object | undefined => {
+  return {
+    V2: {
+      parents: 0,
+      interior: {
+        X1: {
+          accountId32: {
```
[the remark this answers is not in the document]
Asmadek: right, it will be converted to right inside of polkadot js api

src/renderer/widgets/SendAssetModal/ui/components/ActionSteps/InitOperation.tsx
```diff
@@ -107,8 +153,15 @@ export const InitOperation = ({
         nativeToken={nativeToken}
         addressPrefix={addressPrefix}
         fee={fee}
+        xcmFee={xcmFee}
```
egor0798, line 156: Maybe create object prop with keys like xcmFee, xcmWeight so it would be more compact and you can pass it like xcmPros={xcmFee, xcmWeight}

egor0798 (approved):

src/renderer/features/operation/init/ui/OperationFooter.tsx
```diff
@@ -23,12 +27,17 @@ export const OperationFooter = ({
...
+  xcmConfig,
+  onXcmFeeChange,
   onDepositChange,
   onFeeChange,
   onFeeLoading,
 }: Props) => {
   const { t } = useI18n();
 
+  // TODO: Check why transaction can be empty
+  const isXcmTransfer = XcmTypes.includes(transaction?.type);
```
valentunn, line 39: Is this TODO supposed to be left for the feature? In that case its worth to create a separate task to not to forget about it

src/renderer/shared/api/xcm/xcmService.ts
```diff
@@ -67,35 +65,59 @@ export const getEstimatedWeight = (
...
+export const estimateRequiredDestWeight = (
+  config: XcmConfig,
+  assetLocation: AssetLocation,
+  originChain: string,
+  xcmTransfer: XcmTransfer,
+): BN => {
+  const weight = getEstimatedWeight(
+    config.instructions,
+    xcmTransfer.destination.fee.instructions,
+    new BN(xcmTransfer.destination.fee.mode.value),
```
valentunn, line 105: `fee.mode.value` is only present if `fee,mode.type == "proportional"`. Otherwise we should request node for the fee
By the way, why do we need fee mode when estmiating weight at all? Weight itself only depends on `baseInstructionWeight` and `instructions`

src/renderer/shared/api/xcm/xcmService.ts
```diff
@@ -104,8 +126,22 @@ const JunctionType: Record<string, string> = {
   accountId: 'AccountId32',
   generalIndex: 'GeneralIndex',
 };
+
 type JunctionTypeKey = keyof typeof JunctionType;
 
+const JunctionWeight: Record<JunctionTypeKey, number> = {
```
valentunn, line 132: I suggest to rename it to something like `JunctionHierarhyLevel` to not consude it with weight calculations

src/renderer/shared/api/xcm/xcmService.ts
```diff
@@ -143,94 +181,115 @@ const getRelativeAssetLocation = (
...
   const { parachainId: _, ...location } = assetLocation;
 
-  return api.createType('VersionedMultiAsset', {
-    V2: {
-      id: {
-        Concrete: {
-          parents: 0,
-          interior: Object.values(location).length ? createJunctionFromObject(location) : 'Here',
+  return {
+    V2: [
```
valentunn, line 190: It is not always V2, might be V1 as well, depending on a chain (V2 might not be present in VersionedMultiAsset on older versions of the pallet)

src/renderer/shared/api/xcm/xcmService.ts
```diff
@@ -143,94 +181,115 @@ const getRelativeAssetLocation = (
...
+      {
+        id: {
+          Concrete: {
+            parents: 0,
+            interior: Object.values(location).length ? createJunctionFromObject(location) : 'Here',
```
valentunn, line 195: This seems to be dublicating across multiple functions

src/renderer/shared/api/xcm/xcmService.ts
```diff
@@ -143,94 +181,115 @@ const getRelativeAssetLocation = (
...
+          },
+        },
```
valentunn, line 197: Seems like we can split construction of `id` from the rest of the asset (`fungible` part is always the same)

src/renderer/shared/api/xcm/xcmService.ts
valentunn, line 282: Soon we will be integrating EVM chains will use AccountId20, dunno if something should be changed right now, just for you to have a context

src/renderer/shared/api/xcm/xcmService.ts
```diff
@@ -240,15 +299,15 @@ const getChildLocation = (api: ApiPromise, parachainId: number, accountId?: Acco
...
     V2: {
       parents: 0,
       interior: createJunctionFromObject(location),
     },
-  });
+  };
 };
 
-const getParentLocation = (api: ApiPromise, accountId?: AccountId): VersionedMultiLocation => {
+const getParentLocation = (accountId?: AccountId): Object => {
```
valentunn, line 310: What are thee resons for loosing strict typing of return value here?

valentunn (commented):
I havent checked React-specific code, only services part. Leaving the rest for @tuul-wq

src/renderer/entities/transaction/lib/validateBalance.ts
```diff
@@ -59,10 +59,11 @@ const validateBalanceForFee = async ({ transaction, getTransactionFee, api, ...p
   const transferableBalance = transferableAmount(tokenBalance);
   const transferableNativeTokenBalance = transferableAmount(nativeTokenBalance);
   const fee = await getTransactionFee(transaction, api);
+  const feeBN = new BN(fee).add(new BN(transaction.args.xcmFee || 0));
```
valentunn, line 62: `fee` is paid in origin chain's comission assets and the xcmFee is paid in sending asset. I dont think its right to add them to use in validations

src/renderer/entities/transaction/lib/transactionService.ts
```diff
@@ -299,20 +361,23 @@ export const useTransaction = (): ITransactionService => {
     [TransactionType.MULTISIG_CANCEL_AS_MULTI]: ({ threshold, otherSignatories, maybeTimepoint, callHash }, api) =>
       api.tx.multisig.cancelAsMulti(threshold, otherSignatories, maybeTimepoint, callHash),
     // TODO: finish during XCM transfer
```
ERussel, line 363: Do we have a task for this?

src/renderer/entities/transaction/lib/transactionService.ts
```diff
@@ -299,20 +361,23 @@ export const useTransaction = (): ITransactionService => {
...
+      return api.tx.xcmPallet.limitedReserveTransferAssets(xcmDest, xcmBeneficiary, xcmAsset, 0, { Unlimited: true });
     },
-    [TransactionType.XCM_TELEPORT]: () => {
-      return {} as SubmittableExtrinsic<'promise'>;
+    [TransactionType.XCM_TELEPORT]: ({ xcmDest, xcmBeneficiary, xcmAsset }, api) => {
+      return api.tx.xcmPallet.limitedTeleportAssets(xcmDest, xcmBeneficiary, xcmAsset, 0, { Unlimited: true });
     },
-    [TransactionType.POLKADOT_XCM_LIMITED_TRANSFER]: () => {
-      return {} as SubmittableExtrinsic<'promise'>;
+    [TransactionType.POLKADOT_XCM_LIMITED_TRANSFER]: ({ xcmDest, xcmBeneficiary, xcmAsset }, api) => {
```
ERussel, line 370: Why to we need to mark it as Polkadot?

src/renderer/shared/api/xcm/xcmService.ts
```diff
@@ -143,94 +181,115 @@ const getRelativeAssetLocation = (
...
+    V2: [
+      {
+        id: {
+          Concrete: {
+            parents: 0,
+            interior: createJunctionFromObject(location),
+          },
+        },
+        fun: {
+          Fungible: amount.toNumber(),
```
stepanLav, line 199: Number can only safely store up to 53 bits.
https://app.clickup.com/t/862kd1bj9
need BN

src/renderer/shared/api/xcm/xcmService.ts
```diff
@@ -143,94 +181,115 @@ const getRelativeAssetLocation = (
...
+    V2: [
+      {
+        id: {
+          Concrete: {
+            parents: 1,
+            interior: createJunctionFromObject(assetLocation),
+          },
+        },
+        fun: {
+          Fungible: amount.toNumber(),
```
stepanLav, line 223: need to use BN insted

src/renderer/shared/api/xcm/xcmService.ts
```diff
@@ -143,94 +181,115 @@ const getRelativeAssetLocation = (
...
+    V2: [
+      {
+        id: {
+          Concrete: {
+            parents,
+            interior: createJunctionFromObject(location),
+          },
+        },
+        fun: {
+          Fungible: amount.toNumber(),
```
stepanLav, line 249: need to use BN instead

src/renderer/widgets/SendAssetModal/model/send-asset.ts
```diff
@@ -15,17 +52,175 @@ const getConfigFx = attach({ effect: xcmModel.effects.getConfigFx });
...
+
 forward({
   from: xcmConfigRequested,
   to: [getConfigFx, fetchConfigFx],
 });
 
+sample({
+  source: { asset: assetGuardModel.$asset, chain: assetGuardModel.$chain, config: $finalConfig },
+  filter: (props: {
+    asset: Asset | null;
```
tuul-wq, line 71: Using props via `Gate` will make sure that `chain` and `asset` are always defined
Otherwise we won't event be allowed to open `SendAssetModal`

src/renderer/widgets/SendAssetModal/model/send-asset.ts
```diff
@@ -15,17 +52,175 @@ const getConfigFx = attach({ effect: xcmModel.effects.getConfigFx });
...
+    config: $finalConfig,
+    api: $api,
+    amount: $amount,
+    xcmAsset: $xcmAsset,
+    xcmFee: $xcmFee,
+  },
+  fn: ({ api, xcmAsset, config, amount, xcmFee }) => {
+    if (!api || !xcmAsset || !amount || !config || !xcmFee) return null;
+
+    return getAssetLocation(api, xcmAsset, config.assetsLocation, new BN(amount).add(new BN(xcmFee))) || null;
```
tuul-wq, line 208: Ideally `fn` in `sample` must be a pure function but right now it's ok

tuul-wq (approved):

PR #1057, opened 2023-09-12
feat: myself button for xcm transfer

Asmadek:
imageimage

### DIFF
--- a/src/renderer/widgets/SendAssetModal/ui/components/TransferForm.tsx
+++ b/src/renderer/widgets/SendAssetModal/ui/components/TransferForm.tsx
@@ -23,6 +23,7 @@ import { Chain } from '@renderer/entities/chain';
 import { getChainOption } from '../common/utils';
 import { DropdownOption, DropdownResult } from '@renderer/shared/ui/types';
 import { XcmTransfer, XcmTransferType } from '@renderer/shared/api/xcm';
+import AccountSelectModal from '@renderer/pages/Operations/components/modals/AccountSelectModal/AccountSelectModal';
 
 const DESCRIPTION_MAX_LENGTH = 120;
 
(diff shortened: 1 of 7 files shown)


src/renderer/widgets/SendAssetModal/ui/components/TransferForm.tsx
```diff
@@ -345,6 +352,27 @@ export const TransferForm = ({
     }
   };
 
+  const handleMyselfClick = () => {
+    destinationChainAccounts?.length && destinationChainAccounts?.length > 1
+      ? setSelectAccountModalOpen(true)
+      : account && handleAccountSelect(account);
+  };
+
+  const handleAccountSelect = (account: Account) => {
```
tuul-wq, line 361:
```typescript
const handleAccountSelect = (account?: Account) => {
  setSelectAccountModalOpen(false);

  if (!account) return;

  const prefix = destinations.find((c) => c.chainId === destinationChain?.value)?.addressPrefix || SS58_DEFAULT_PREFIX;
  const address = toAddress(account.accountId, { prefix });

  setValue('destination', address, { shouldValidate: true });
};
```

src/renderer/widgets/SendAssetModal/ui/components/TransferForm.tsx
```diff
@@ -345,6 +352,27 @@ export const TransferForm = ({
     }
   };
 
+  const handleMyselfClick = () => {
```
tuul-wq, line 355: As I see in worst case scenario `destinationChainAccounts` is `[]` and cannot be `undefined`.
```typescript
const handleMyselfClick = () => {
  if (destinationChainAccounts.length > 1) {
    setSelectAccountModalOpen(true);
  } else {
    handleAccountSelect(account);
  }
};
```

tuul-wq (changes requested):
Just a couple of comments

PR #1060, opened 2023-09-15
feat: placeholder for select dropdown

Asmadek:
* feat: placeholder for select dropdown
* feat: add calculate fee for xcm from api


tuul-wq (approved):
A couple of improvements

PR #1062, opened 2023-09-15
Fix: Approve xcm

tuul-wq:
- Correct filter of WOW accounts on `Approve modal`
- Info popover for Approve signatories can drop out until you pass more that 7 accounts (same logic from on Validators modal)

image

### DIFF
--- a/src/renderer/pages/Operations/components/modals/SignatorySelectModal.tsx
+++ b/src/renderer/pages/Operations/components/modals/SignatorySelectModal.tsx
@@ -4,46 +4,44 @@ import { AccountDS } from '@renderer/shared/api/storage';
 import { Asset } from '@renderer/entities/asset';
 import { Chain } from '@renderer/entities/chain';
 import { SelectableSignatory } from '@renderer/entities/signatory';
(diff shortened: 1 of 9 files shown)


tuul-wq:
Task linked: [CU-862kd5y3x Multisig XCM approve\reject lead to empty screen.](https://app.clickup.com/t/862kd5y3x)

src/renderer/pages/Operations/components/modals/SignatorySelectModal.tsx
```diff
@@ -4,46 +4,44 @@ import { AccountDS } from '@renderer/shared/api/storage';
...
-                name={a.name}
-                asset={asset}
-                chainId={chain.chainId}
-                explorers={chain.explorers}
-                value={a}
-                onSelected={onSelect}
-              />
-            </li>
-          ))}
+      <ul className={cnTw('mt-1', accounts.length > 7 && 'max-h-[332px] overflow-y-auto')}>
```
Asmadek, line 29: can you add same for `AccountSelectModal`?

Asmadek (approved):

PR #1076, opened 2023-09-18
Feat: Currency settings

egor0798:
changes:

- currency settings modal
- grouped select


src/renderer/widgets/SelectCurrencyModal/ui/SelectCurrencyModal.tsx
```diff
@@ -0,0 +1,94 @@
...
+    element: [currency.code, currency.symbol, currency.name].filter(Boolean).join(' • '),
+  });
+
+  const currenciesGroups: DropdownOptionGroup<number>[] = [
+    { label: t('settings.currency.cryptocurrenciesLabel'), options: cryptoCurrencies.map(getCurrencyOption) },
+    { label: t('settings.currency.popularFiatLabel'), options: popularFiatCurrencies.map(getCurrencyOption) },
+    { label: t('settings.currency.unpopularFiatLabel'), options: unpopularFiatCurrencies.map(getCurrencyOption) },
+  ];
+
+  const handleClose = () => {
```
tuul-wq, line 48: To make css modal tranistion play take a look at
```javascript
  const closeContactModal = () => {
    toggleIsModalOpen();
    setTimeout(onClose, DEFAULT_TRANSITION);
  };
```

src/renderer/widgets/SelectCurrencyModal/ui/SelectCurrencyModal.tsx
```diff
@@ -0,0 +1,94 @@
...
+    toggleIsOpen();
+    onClose();
+  };
+
+  const saveChanges = () => {
```
tuul-wq, line 53: Ideally, form should be done as a `feature` with it's model that has `submitted` event that handles ongoing actions and condition

tuul-wq (changes requested):
PR looks nice!
Added comments about using effector and some code style guides 😉

src/renderer/shared/ui/Dropdowns/GroupedSelect/GroupedSelect.tsx
Asmadek, line 1: We already have such feature in `Select` component with disabled attribute for option. If you want to stay this component, please, rework places where we use `Select` (cross chain transfer)  to this component to make our code cleaner and used in one way.

Asmadek (changes requested):

src/renderer/shared/ui/Modals/BaseModal/BaseModal.tsx
```diff
@@ -33,7 +33,7 @@ const BaseModal = ({
 
   return (
     <Transition appear show={isOpen} as={Fragment}>
-      <Dialog as="div" className="relative z-50" onClose={onClose}>
+      <Dialog as="div" className="relative z-50" onClose={() => onClose()}>
```
tuul-wq, line 36: Something was wrong with previous implementation?

src/renderer/entities/price/model/currency-model.ts
```diff
@@ -8,6 +8,14 @@ const $currencyConfig = createStore<CurrencyItem[]>([]);
 const $activeCurrency = createStore<CurrencyItem | null>(null);
 const $activeCurrencyCode = createStore<CurrencyItem['code'] | null>(null);
 
+const $cryptoCurrencies = $currencyConfig.map((config) => config.filter((c) => c.category === 'crypto'));
```
tuul-wq, line 11: These derived stores should be in `form-model`, entity must contain most generic data and logic for it's transformation like: 
- config / changeConfig
- activeCurrency / changeActiveCurrency
and so on

src/renderer/features/currency/ChangeCurrency/model/currency-form.ts
tuul-wq, line 1: Why not `CurrencyForm` instead of `ChangeCurrency`?
And I don't see any ui for this model, it can easily be moved out of widget to feature
If you need this feature elsewhere in the app, you just import it without duplicating the UI from `SelectCurrencyModal`

src/renderer/features/currency/ChangeCurrency/model/currency-form.ts
```diff
@@ -0,0 +1,63 @@
...
+  to: $activeCurrency,
+});
+
+forward({
+  from: fiatFlagChanged,
+  to: $fiatFlag,
+});
+
+// save changes
+sample({
```
tuul-wq, line 42: Because we don't have a unified form we have to call `changed` events one by one, that's fine
But there is a much cleaner way with [effector/patronum](https://patronum.effector.dev/methods/spread/#call-events-on-store-changes)
It would look something like this:
```javascript
sample({
  clock: formSubmitted,
  source: { fiatFlag: $fiatFlag, activeCurrency: $activeCurrency },
  target: spread({
    targets: {
      fiatFlag: priceProviderModel.events.fiatFlagChanged,
      activeCurrency: currencyModel.events.currencyChanged,
    }
  })
})
```
I've never used it before, so If it doesn't work, please let me know 😉

tuul-wq: Overall I assume it would be right to use `effector-forms` even though we have 2 fields without any inputs.
With form it would be easier to add new fields (store + event + validation) instead of doing it by hand
I was wrong guessing that 2 non-input fields are ok to not introduce the form

src/renderer/widgets/SelectCurrencyModal/ui/SelectCurrencyModal.tsx
```diff
@@ -0,0 +1,89 @@
...
+  return (
+    <BaseModal
+      isOpen={isOpen}
+      closeButton
+      contentClass="py-4 px-5"
+      panelClass="w-[440px]"
+      title={t('settings.currency.modalTitle')}
+      onClose={handleClose}
+    >
+      <form className="flex flex-col gap-y-4" onSubmit={saveChanges}>
```
tuul-wq, line 67: `<CurrencyForm onSubmit={closeCurrencyModal} />` is all we need here
```javascript
const closeCurrencyModal = () => {
    toggleIsModalOpen();
    setTimeout(onClose, DEFAULT_TRANSITION);
  };
```

tuul-wq (changes requested):
Added additional comments

Asmadek (approved):

PR #1084, opened 2023-09-20
Feat: Price provider model

tuul-wq:
- Price provider model
- Fiat service
- Currency model


src/renderer/app/App.tsx
```diff
@@ -22,6 +24,10 @@ const App = () => {
   const appRoutes = useRoutes(routesConfig);
   const { getAccounts } = useAccount();
 
+  const [assetsPrices, activeCurrency] = useUnit([priceProviderModel.$assetsPrices, currencyModel.$activeCurrency]);
```
tuul-wq, line 27: This is added temporary, to make effector model work you need to import it somewhere.
Can be deleted after Egor's PR

tuul-wq:
Task linked: [CU-866avwp8j Create fiat values data structure and service](https://app.clickup.com/t/866avwp8j)

src/renderer/shared/api/price-provider/common/constants.ts
```diff
@@ -0,0 +1,10 @@
+export const COINGECKO_URL = 'https://api.coingecko.com/api/v3';
+
+export const CURRENCY_URL =
+  'https://raw.githubusercontent.com/novasamatech/nova-wallet-android/develop/feature-currency-impl/src/main/res/raw/currencies.json';
```
valentunn, line 4: I suggest bundling the currency file with the application and not fetching it from the network. Otherwise, we might forget that Spektr depends on this source file in Nova Android and perform a breaking change that will break Spektr

src/renderer/entities/price/model/currency-model.ts
```diff
@@ -0,0 +1,84 @@
+import { createEvent, createStore, createEffect, forward, sample } from 'effector';
+
+import { kernelModel } from '@renderer/shared/core';
+import { CurrencyConfig, fiatService } from '@renderer/shared/api/price-provider';
+import { DEFAULT_CURRENCY_CODE, DEFAULT_CURRENCY_CONFIG } from '../lib/constants';
+
+export const $currencyConfig = createStore<CurrencyConfig[]>([]);
```
Asmadek, line 7: are you sure about naming? `$currencyConfig` variable and `CurrencyConfig` type are confused, because first is array but second is object.

src/renderer/shared/api/price-provider/service/fiatService.ts
```diff
@@ -0,0 +1,71 @@
...
+function saveFiatProvider(provider: string) {
+  localStorageService.saveToStorage(FIAT_PROVIDER, provider);
+}
+
+function getAssetsPrices<T extends any>(defaultPrices: T): T {
+  return localStorageService.getFromStorage(ASSETS_PRICES, defaultPrices);
+}
+
+function saveAssetsPrices<T extends any>(prices: T) {
+  localStorageService.saveToStorage(ASSETS_PRICES, prices);
```
Asmadek, line 70: Why did you decide to store prices in local storage?

src/renderer/entities/price/model/price-provider-model.ts
```diff
@@ -0,0 +1,98 @@
...
+forward({ from: fiatFlagChanged, to: [$fiatFlag, saveFiatFlagFx] });
+
+forward({ from: priceProviderChanged, to: [$priceProvider, savePriceProviderFx] });
+
+forward({ from: fetchAssetsPricesFx.doneData, to: [$assetsPrices, saveAssetsPricesFx] });
+
+sample({
+  clock: assetsPricesRequested,
+  source: $priceProvider,
+  target: [$priceProvider, saveAssetsPricesFx],
```
Asmadek, line 93: I am not clearly understand why do you need to provide `$priceProvider` to `$priceProvider`?

src/renderer/entities/price/model/price-provider-model.ts
```diff
@@ -0,0 +1,98 @@
...
+});
+
+forward({ from: getFiatFlagFx.doneData, to: $fiatFlag });
+
+forward({ from: getPriceProviderFx.doneData, to: $priceProvider });
+
+forward({ from: getAssetsPricesFx.doneData, to: $assetsPrices });
+
+sample({
+  clock: [$priceProvider, currencyModel.$activeCurrency],
```
Asmadek, line 75: is clock really needed here? sample will be run on source changed, and here source is the same as clock

Asmadek (approved):

egor0798 (approved):

src/renderer/entities/price/model/price-provider-model.ts
tuul-wq: Added missing `assetsPricesRequested` to clock array and it will need data from the source
Otherwise yes, you're right about source update without clock 👍

```typescript
sample({
  clock: [assetsPricesRequested, $priceProvider, currencyModel.$activeCurrency],
  source: { provider: $priceProvider, currency: currencyModel.$activeCurrency },
  filter: ({ provider, currency }) => provider !== null && currency !== null,
  fn: ({ provider, currency }) => {
    return { provider: provider!, currencies: [currency!.coingeckoId], includeRates: true };
  },
  target: fetchAssetsPricesFx,
});
```

tuul-wq: By the way I've asked about type guards for complex arguments, and core contributor told me not to hesitate using bang operator, because we are sure in our filter
Or we should provide a type guard with full typing which is really verbose and seems bizzare

> Подскажите, как корректно следить за 2 сторами в sample, сторы могут быть null, 
> и filter должен пропускать только, если оба стора не пусты?
> 
> Получается что-то вроде этого, но TS не пускает из-за error: 'fn result should extend target type';
> 
> sample({
>   clock: [$first, $second],
>   source: { first: $first, second: $second },
>   filter: ({ first, second }) => first && second,
>   fn: ({ first, second }) => ({ one: first, two: second }),
>   target: myEffect,
> });
----

> всё правильно, только в filter сделай корректную проверку на null, а то сейчас 0 и пустую строку не пропустит тоже.
> чтобы успокоить ts можно или типизировать filter для сужения типа source, или проще поставить восклицательные знаки в fn.

> если нет предубеждений против !, то я бы просто их повтыкал :) так ты просто говоришь ts, что тут точно не будет null|undefined, что ты лучше знаешь

PR #1086, opened 2023-09-21
Adding system tests

stepanLav:
This PR adds base implementation of system tests which uses [Playwright](https://playwright.dev/).
How to use it may be found here - https://github.com/novasamatech/nova-spektr/tree/feat/system_tests/tests/system#system-tests


playwright.config.ts
```diff
@@ -0,0 +1,77 @@
...
+  },
+
+  /* Configure projects for major browsers */
+  projects: [
+    {
+      name: 'chromium',
+      use: { ...devices['Desktop Chrome'] },
+    },
+
+    // {
```
Asmadek, line 40: Do we need all this commented code?

tests/system/BaseTestConfig.ts
```diff
@@ -0,0 +1,7 @@
+class BaseTestConfig {
```
Asmadek, line 1: I am not sure that class is needed here, simple object is good enough.
And what is the reason to use snake case here? and not camel case which we use everywhere?

Asmadek (commented):
Maybe add readme where describe how to write tests and page/elements classes?

tests/system/BaseTestConfig.ts
stepanLav: ty! It's my fault (python code wants to get out 😅) I will fix it

playwright.config.ts
stepanLav: yeah, you are right, I experimented with different browsers and use default generated config, I will clean it up

playwright.config.ts
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * See https://playwright.dev/docs/test-configuration.
+ */
+export default defineConfig({
+  testDir: './tests/system',
+  /* Run tests in files in parallel */
+  fullyParallel: true,
+  /* Fail the build on CI if you accidentally left test.only in the source code. */
+  forbidOnly: !!process.env.CI,
```
tuul-wq, line 17:
```suggestion
  forbidOnly: Boolean(process.env.CI),
```

tests/system/BaseTestConfig.ts
tuul-wq: Agree with Alexander

tests/system/matrix/authorisation/matrix.auth.system.test.tsx
tuul-wq, line 1: Do we need such long names, because we already in `/matrix/authorisation` directory?

tests/system/pages/_elements/LoginPageElements.ts
```diff
@@ -0,0 +1,13 @@
+import { BasePageElements } from './BasePageElements';
+
+export class LoginPageElements implements BasePageElements {
```
tuul-wq, line 3: Plain js object instead of class?

tuul-wq (approved):

tests/system/pages/_elements/LoginPageElements.ts
stepanLav: In that case I will force to implement additional interface for LoginPageElements in order to get easy access and checks for parameters, currently it looks like overenginering, but when project will grow up, it will be tricky to manage different pages with all elements without any support from IDE. Approach with the interface looks more bulky:

```ts

interface BasePageElements {
  url: string;
}

interface LoginPageElements extends BasePageElements {
  enterAccountAddress: string;
  accountNameField: string;
  watchOnlyButton: string;
  polkadotVaultButton: string;
  continueButton: string;
  accessDeniedText: string;
  firstInfoButton: string;
  subscanLabel: string;
}

const LoginPageElements: LoginPageElements = {
  url: 'https://localhost:3000/#/onboarding',
  enterAccountAddress: 'Enter or paste your account address',
  accountNameField: 'Enter a name for your wallet',
  watchOnlyButton: 'text=Watch-only',
  polkadotVaultButton: 'text=Polkadot Vault',
  continueButton: 'Continue',
  accessDeniedText: 'text=Access denied!',
  firstInfoButton: '[id="headlessui-menu-button-\\:r4\\:"]',
  subscanLabel: 'View on Subscan',
};
```

Or you have better solution which I missed?

tuul-wq: Yeah, there's no right way in TS to declare an object with properties besides `BasePageElements` and have a nice type completion....
But using `class` just to store a bunch on properties isn't right as well
@Asmadek @egor0798 maybe you guys have some ideas how to add type completion for `LoginPageElements` not declaring additional interface?

stepanLav: decided to live that approach as it is

PR #1089, opened 2023-09-25
feat: custom menu

pgolovkin:
@stepanLav this PR has to be tested on Windows, Mac, Linux.
`Nova Spektr` should be in the menu but not the `nova-spektr`.

### DIFF
--- a/src/main/vectormenu.ts
+++ b/src/main/vectormenu.ts
@@ -1,6 +1,6 @@
 import { app, shell, Menu, MenuItem, MenuItemConstructorOptions } from 'electron';
 
-import { PLATFORM } from '@shared/constants';
+import { ENVIRONMENT, PLATFORM } from '@shared/constants';
 
 export function buildMenuTemplate(): Menu {
   const template: Array<MenuItemConstructorOptions | MenuItem> = [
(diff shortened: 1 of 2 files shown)


src/main/vectormenu.ts
```diff
@@ -0,0 +1,182 @@
...
+          label: 'Zoom Out',
+        },
+        {
+          role: 'togglefullscreen',
+          label: 'Toggle Full Screen',
+        },
+        {
+          role: 'toggleDevTools',
+          label: 'Toggle Developer Tools',
+        },
```
Asmadek, line 85: Maybe better to hide dev tools for prod version?

src/main/vectormenu.ts
```diff
@@ -0,0 +1,188 @@
+import { app, shell, Menu, MenuItem, MenuItemConstructorOptions } from 'electron';
+
+import { ENVIRONMENT, PLATFORM } from '@shared/constants';
+
+export function buildMenuTemplate(): Menu {
+  const template: Array<MenuItemConstructorOptions | MenuItem> = [
```
tuul-wq, line 6:
```suggestion
  const Template: MenuItemConstructorOptions[] | MenuItem[] = [
```

tuul-wq (approved):
Nice 👍

src/main/vectormenu.ts
pgolovkin: Agree with the type, but why the constant name starts from the capital `T`?

Asmadek (approved):

PR #1096, opened 2023-09-27
Feat: Fiat update rules

egor0798:
## Changes:
- Added prices update request to Assets, Staking and Operations Pagw
- Added prices update request to Staking, Transfer and Multisig operations

### DIFF
--- a/src/renderer/entities/price/model/__tests__/price-provider-model.test.ts
+++ b/src/renderer/entities/price/model/__tests__/price-provider-model.test.ts
@@ -78,9 +78,11 @@ describe('entities/price/model/price-provider-model', () => {
   test('should update $assetPrices when currencyChanged', async () => {
     jest.spyOn(fiatService, 'getAssetsPrices').mockReturnValue(prices);
 
-    const scope = fork();
-    await allSettled(kernelModel.events.appStarted, { scope });
+    const scope = fork({
+      values: new Map().set(priceProviderModel.$assetsPrices, {}),
+    });
+    const initialPrices = priceProviderModel.$assetsPrices;
     await allSettled(currencyModel.events.currencyChanged, { scope, params: 1 });
-    expect(scope.getState(priceProviderModel.$assetsPrices)).toEqual(prices);
+    expect(scope.getState(priceProviderModel.$assetsPrices)).not.toEqual(initialPrices);
   });
 });
--- a/src/renderer/entities/price/model/__tests__/price-provider-model.test.ts
+++ b/src/renderer/entities/price/model/__tests__/price-provider-model.test.ts
@@ -75,7 +75,7 @@ describe('entities/price/model/price-provider-model', () => {
     expect(scope.getState(priceProviderModel.$assetsPrices)).toEqual(prices);
   });
 
-  test('should update asset prices on currency change', async () => {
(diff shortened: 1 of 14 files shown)


src/renderer/entities/price/model/__tests__/price-provider-model.test.ts
```diff
@@ -74,4 +74,13 @@ describe('entities/price/model/price-provider-model', () => {
     await allSettled(priceProviderModel.events.assetsPricesRequested, { scope, params: { includeRates: false } });
     expect(scope.getState(priceProviderModel.$assetsPrices)).toEqual(prices);
   });
+
+  test('should update asset prices on currency change', async () => {
```
tuul-wq, line 78: Let's stick to actual `store` / `events` names
```suggestion
  test('should update $assetPrices when currencyChanged', async () => {
```

src/renderer/entities/price/model/__tests__/price-provider-model.test.ts
```diff
@@ -74,4 +74,13 @@ describe('entities/price/model/price-provider-model', () => {
...
+    jest.spyOn(fiatService, 'getAssetsPrices').mockReturnValue(prices);
+
+    const scope = fork();
+    await allSettled(kernelModel.events.appStarted, { scope });
+    await allSettled(currencyModel.events.currencyChanged, { scope, params: 1 });
+    expect(scope.getState(priceProviderModel.$assetsPrices)).toEqual(prices);
```
tuul-wq, line 84: I suppose we need to compare value before and value after `currencyModel.events.currencyChanged`
instead of checking `prices` after initial event (`kernelModel.events.appStarted`)
By the way, you can provide `$assetPrices` inside `fork` without emitting `appStarted` take a look at test located above `test('should fetch $assetsPrices when assetsPricesRequested')`

tuul-wq (changes requested):

src/renderer/entities/transaction/ui/Deposit/Deposit.tsx
```diff
@@ -25,5 +29,14 @@ export const Deposit = memo(({ api, asset, threshold, className, onDepositChange
...
+  if (!fiatFlag) {
+    return <AssetBalance value={deposit} asset={asset} className={className} />;
+  }
+
+  return (
+    <div className="flex flex-col gap-y-0.5 items-end">
+      <AssetBalance value={deposit} asset={asset} className={className} />
+      <AssetFiatBalance asset={asset} amount={deposit} />
+    </div>
+  );
```
Asmadek, line 41: What is the pros of such approach and not 

```typescript
  return (
    <div className="flex flex-col gap-y-0.5 items-end">
      <AssetBalance value={deposit} asset={asset} className={className} />
      { fiatFlag && <AssetFiatBalance asset={asset} amount={deposit} />}
    </div>
  )
```

Asmadek: Same for other places where you use such approach

Asmadek (approved):

tuul-wq (changes requested):

tuul-wq (approved):

PR #1102, opened 2023-09-29
Feat: removed multishard wallets from signatory options

egor0798:


src/renderer/entities/account/model/account.ts
```diff
@@ -1,12 +1,12 @@
 import { createKeyMulti } from '@polkadot/util-crypto';
 import { u8aToHex } from '@polkadot/util';
 
-import { AccountDS, ID } from '@renderer/shared/api/storage';
+import { AccountDS, ID, WalletDS } from '@renderer/shared/api/storage';
```
tuul-wq, line 4: Wallet type is enough

src/renderer/entities/account/model/account.ts
```diff
@@ -112,10 +112,11 @@ export const isMultishard = (account?: Account | MultisigAccount): boolean => {
   return Boolean(account.walletId);
 };
 
-export function isWalletContact(account?: Account | MultisigAccount): boolean {
+export function isWalletContact(account?: Account | MultisigAccount, wallet?: WalletDS): boolean {
   if (!account) return false;
+  const notMultishard = !wallet || wallet.type !== WalletType.MULTISHARD_PARITY_SIGNER;
```
tuul-wq, line 117: Better use positive variables accompanied with negative operator
```suggestion
  const isMultishard = wallet && wallet.type === WalletType.MULTISHARD_PARITY_SIGNER;
```

src/renderer/entities/account/model/account.ts
```diff
@@ -112,10 +112,11 @@ export const isMultishard = (account?: Account | MultisigAccount): boolean => {
...
 
-  return account.signingType !== SigningType.WATCH_ONLY && !isMultisig(account);
+  return account.signingType !== SigningType.WATCH_ONLY && notMultishard && !isMultisig(account);
```
tuul-wq, line 119: Would be great to add variable for this part of condition
```javascript
  const isWatchOnly = account.signingType === SigningType.WATCH_ONLY
```

tuul-wq (changes requested):

Asmadek (approved):

tuul-wq (approved):

PR #1117, opened 2023-10-04
feat: remove old log files

pgolovkin:
1. Keep 10 log files - may be discussed
2. Remove old files when new files is added
3. If the user already has more then 10 log files then extra-files will be removed

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -11,6 +11,7 @@ import { createWindow } from './factories/create';
 import { buildMenuTemplate } from '@main/vectormenu';
 
 const { MAIN, TITLE } = APP_CONFIG;
+const MAX_LOG_FILES_TO_KEEP = 10;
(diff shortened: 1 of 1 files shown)


src/main/main.ts
```diff
@@ -24,7 +25,14 @@ log.transports.file.maxSize = 1048576 * 5; // 5 MB;
 log.transports.file.archiveLogFn = (oldLogFile: LogFile): void => {
```
tuul-wq, line 25: Would be great to make it like this:
```javascript
log.transports.file.archiveLogFn = archiveLogFn;

// somewhere at the bottom
function archiveLogFn(oldLogFile: LogFile) {
  // logic
}
```

tuul-wq (approved):

Asmadek (approved):

PR #1120, opened 2023-10-04
Fix: network fiat value sum

egor0798:


src/renderer/shared/lib/utils/balance.ts
```diff
@@ -220,3 +220,24 @@ export const formatFiatBalance = (balance = '0', precision = 0): FormattedBalanc
     decimalPlaces,
   };
 };
+
+export const getRoundedFiatValue = (assetBalance: string, price: number, precision: number) => {
```
stepanLav, line 224: I would suggest to add some tests:
```typescript
import { getRoundedFiatValue } from '@renderer/shared/lib/utils';

describe('getRoundedFiatValue', () => {
  it('should return correct rounded fiat value when fiat balance is greater than or equal to 1', () => {
    const assetBalance = '2';
    const price = 1.5;
    const precision = 2;
    const result = getRoundedFiatValue(assetBalance, price, precision);
    expect(result.toString()).toEqual('3.00'); // 2 * 1.5 = 3.00
  });

  it('should return correct rounded fiat value when fiat balance is less than 1', () => {
    const assetBalance = '0.000123456';
    const price = 1;
    const precision = 6;
    const result = getRoundedFiatValue(assetBalance, price, precision);
    expect(result.toString()).toEqual('0.000123'); // Rounded to 5 decimal places
  });

  it('should return correct rounded fiat value when fiat balance is zero', () => {
    const assetBalance = '0';
    const price = 1;
    const precision = 2;
    const result = getRoundedFiatValue(assetBalance, price, precision);
    expect(result.toString()).toEqual('0.00');
  });
});
```

tuul-wq (approved):

src/renderer/shared/lib/utils/balance.ts
```diff
@@ -220,3 +220,24 @@ export const formatFiatBalance = (balance = '0', precision = 0): FormattedBalanc
...
+  const bnPrecision = new BNWithConfig(precision);
+  const TEN = new BNWithConfig(10);
+  const bnFiatBalance = new BNWithConfig(fiatBalance.toString()).div(TEN.pow(bnPrecision));
+
+  if (bnFiatBalance.gte(1)) {
+    return bnFiatBalance.decimalPlaces(2);
+  }
+
+  const decimalPart = bnFiatBalance.toString().split('.')[1];
+  const decimalPlaces = Math.max((decimalPart || '').search(/[1-9]/) + 1, 5);
```
Asmadek, line 240: I think this part can be moved to function because formatFiatBalance use it too

Asmadek (approved):

PR #1128, opened 2023-10-05
Fix: added number formatting to prevent exponential notation

egor0798:


src/renderer/shared/lib/utils/strings.ts
```diff
@@ -86,3 +86,22 @@ export const formatSectionAndMethod = (section: string, method: string): string
...
   return `${splitFn(sectionSplit)}: ${splitFn(methodSplit)}`;
 };
+
+/**
+ * Formats number without exponential notation and removes trailing zeroe
+ * @param value number to format
+ * @param maxPrecision Maximum number of characters in decimal part
+ * @return {String}
+ */
+export const toFixedNotation = (value: number, maxPrecision = 20) => {
```
tuul-wq, line 96: Return type would be nice

src/renderer/shared/lib/utils/strings.ts
```diff
@@ -86,3 +86,22 @@ export const formatSectionAndMethod = (section: string, method: string): string
...
+  const fixedValue = value.toFixed(maxPrecision);
+  const decimalPart = fixedValue.split('.')[1];
+  if (!decimalPart) return value.toString();
+
+  const trailingZeros = decimalPart.search(/0*(?![1-9])+$/g);
+  if (trailingZeros === -1 || trailingZeros === decimalPart.length) {
+    return fixedValue;
+  } else {
```
tuul-wq, line 104: `else` is not required after `if - return`

tuul-wq (approved):
Overall is fine 👍

Asmadek (approved):

PR #1141, opened 2023-10-11
Feat: select signatory update

egor0798:
## Changes:
- Updated select signatory modal style
- Added support for shard account as signatory 

image


src/renderer/entities/signatory/ui/SelectableSignatory/SelectableSignatory.tsx
```diff
@@ -1,51 +1,63 @@
...
+  addressPrefix,
+  chainId,
+  walletId,
+  onSelected,
 }: Props<T>) => {
-  const address = getAddress(addressProps);
+  const wallets = useUnit(walletModel.$wallets);
+  const signatoryWallet = wallets.find((w) => w.id === walletId);
+
+  if (!signatoryWallet) return <></>;
```
Asmadek, line 33: `useLiveQuery` use `useEffect` inside, return before hook declaration may cause errors

src/renderer/entities/signatory/ui/SelectableSignatory/SelectableSignatory.tsx
```diff
@@ -1,51 +1,63 @@
...
 
   const { getLiveBalance } = useBalance();
-  const popoverItems = useAddressInfo(address, explorers, true);
   const balance = getLiveBalance(toAccountId(address), chainId, asset.assetId.toString());
 
+  // support for legacy signatory as shard from multishard account
+  const walletType =
+    signatoryWallet.type === WalletType.MULTISHARD_PARITY_SIGNER
+      ? WalletType.SINGLE_PARITY_SIGNER
+      : signatoryWallet.type;
```
Asmadek, line 44: looks like this code can be used in other places, can you make it as a function?

src/renderer/pages/Operations/common/utils.ts
```diff
@@ -218,3 +236,39 @@ export const getSignatoryName = (
...
 };
+
+export const getSignatoryAccounts = (
+  accounts: Account[],
+  wallets: Wallet[],
+  events: MultisigEvent[],
+  signatories: Signatory[],
+  chainId: ChainId,
+) => {
+  const walletsDict = keyBy(wallets, 'id');
```
Asmadek, line 247: In most of places we use map instead of dict

src/renderer/pages/Operations/common/utils.ts
```diff
@@ -218,3 +236,39 @@ export const getSignatoryName = (
...
+  return signatories.reduce((accum: Account[], signatory) => {
+    const filteredAccounts = accounts.filter(
+      (a) => a.accountId === signatory.accountId && !events.some((e) => e.accountId === a.accountId),
+    );
+
+    const signatoryAccount = filteredAccounts.find((a) => {
+      const isCurrentChain = accountUtils.isChainIdMatch(a, chainId);
+      const walletType = walletsDict[a.walletId].type;
+      // TODO add wallet connect
+      const isCorrectWalletType = walletType === WalletType.SINGLE_PARITY_SIGNER;
```
Asmadek, line 258: maybe add function here too like `isChainIdMatch`?

Asmadek (changes requested):
To avoid double review, better to use dest branch which was an origin for this branch, dest branch will change automatically after merge of origin branch

src/renderer/entities/signatory/ui/SelectableSignatory/SelectableSignatory.tsx
egor0798: I moved this code into WalletIcon component in other PR because from now on we show legacy multishard wallet as polakadot vault.

src/renderer/pages/Operations/common/utils.ts
egor0798: added to wallet utils

src/renderer/entities/wallet/lib/wallet-utils.ts
```diff
@@ -23,3 +26,8 @@ function isMultisig(wallet?: Wallet | null): boolean {
 function isWatchOnly(wallet?: Wallet | null): boolean {
   return wallet?.type === WalletType.WATCH_ONLY;
 }
+
+function isWalidSignatory(wallet?: Wallet | null) {
```
Asmadek, line 30: `isValidSignatory`

Asmadek (approved):

src/renderer/pages/Operations/common/utils.ts
```diff
@@ -218,3 +233,37 @@ export const getSignatoryName = (
...
+export const getSignatoryAccounts = (
+  accounts: Account[],
+  wallets: Wallet[],
+  events: MultisigEvent[],
+  signatories: Signatory[],
+  chainId: ChainId,
+) => {
+  const walletsMap = new Map(wallets.map((wallet) => [wallet.id, wallet]));
+
+  return signatories.reduce((accum: Account[], signatory) => {
```
tuul-wq, line 246: We use `acc` instead of `accum` across the app

src/renderer/entities/wallet/lib/wallet-utils.ts
```diff
@@ -6,8 +6,11 @@ export const walletUtils = {
   isMultiShard,
   isMultisig,
   isWatchOnly,
+  isWalidSignatory,
 };
 
+const VALID_SIGNATORY_WALLETS = [WalletType.SINGLE_PARITY_SIGNER];
```
tuul-wq, line 12: I'd rather incapsulate this constant in the function it's being used

tuul-wq (changes requested):

src/renderer/entities/wallet/lib/wallet-utils.ts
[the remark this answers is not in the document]
egor0798: So I'll leave it in wallet-utils for now.

src/renderer/entities/wallet/lib/wallet-utils.ts
```diff
@@ -6,6 +6,7 @@ export const walletUtils = {
   isMultiShard,
   isMultisig,
   isWatchOnly,
+  isWalidSignatory: isValidSignatory,
```
tuul-wq, line 9:
```suggestion
  isValidSignatory,
```

src/renderer/entities/wallet/lib/wallet-utils.ts
```diff
@@ -23,3 +24,10 @@ function isMultisig(wallet?: Wallet | null): boolean {
 function isWatchOnly(wallet?: Wallet | null): boolean {
   return wallet?.type === WalletType.WATCH_ONLY;
 }
+
+function isValidSignatory(wallet?: Wallet | null) {
```
tuul-wq, line 28:
```suggestion
function isValidSignatory(wallet?: Wallet | null): boolean {
```

src/renderer/pages/Operations/common/utils.ts
```diff
@@ -218,3 +233,37 @@ export const getSignatoryName = (
...
+  return signatories.reduce((acc: Account[], signatory) => {
+    const filteredAccounts = accounts.filter(
+      (a) => a.accountId === signatory.accountId && !events.some((e) => e.accountId === a.accountId),
+    );
+
+    const signatoryAccount = filteredAccounts.find((a) => {
+      const isChainMatch = accountUtils.isChainIdMatch(a, chainId);
+      const wallet = walletsMap.get(a.walletId);
+
+      return isChainMatch && walletUtils.isWalidSignatory(wallet);
```
tuul-wq, line 255: Don't forget to fix `isWalidSignatory` 😉

src/renderer/pages/Operations/common/utils.ts
```diff
@@ -218,3 +233,37 @@ export const getSignatoryName = (
...
   return toAddress(signatoryId, { chunk: 5, prefix: addressPrefix });
 };
+
+export const getSignatoryAccounts = (
+  accounts: Account[],
+  wallets: Wallet[],
+  events: MultisigEvent[],
+  signatories: Signatory[],
+  chainId: ChainId,
+) => {
```
tuul-wq, line 243: Please add return type

src/renderer/pages/Operations/common/utils.ts
tuul-wq: Let's say I have 1 Legacy Multisig account with 3 signatories:
- **my** Signle_shard_parity_signer (root)
- **my** Multishard_parity_signer (root)
- friend's account
According to `isValidSignatory` Single_shard is the only one I will see
Seems to me that I should be able to choose from 2 signatories (single_shard & multishard), am I right?
@pgolovkin Please, correct me if I'm wrong from the business perspective

tuul-wq (changes requested):

src/renderer/pages/Operations/common/utils.ts
egor0798: This case is handled in `getSignatoryAccount` where `isValidSignatory ` is used. I think `isValidSignatory` should represetn what type of wallet can we use as signatory in current state and legacy case should be handled outside of this method

tuul-wq (approved):

PR #1146, opened 2023-10-15
Feat: operation details update

egor0798:
## Changes:
- updated operation details to display wallet
- updated signatories list and moved to separate component

image


src/renderer/pages/Operations/components/OperationFullInfo.tsx
```diff
@@ -127,53 +84,9 @@ const OperationFullInfo = ({ tx, account }: Props) => {
...
-                matrixId={matrixId}
-                explorers={connection?.explorers}
-                name={getSignatoryName(accountId, tx.signatories, contacts, accounts, connection.addressPrefix)}
-                status={getSignatoryStatus(accountId)}
-              />
-            </li>
-          ))}
-        </ul>
-      </div>
+      {account && <OperationSignatories tx={tx} connection={connection} account={account} />}
```
tuul-wq, line 87: 👍

src/renderer/entities/wallet/ui/WalletRow/WalletRow.tsx
tuul-wq, line 1: Feels like `WalletRow` is a name tightly coupled with some table or list
You tried to implement [this component](https://www.figma.com/file/IzDQl7uzHQVhdnf4tpzXVx/Spektr-Design-System?type=design&node-id=4286-10444&mode=dev), at least partially?

tuul-wq (approved):
Approve with questions

src/renderer/pages/Operations/components/OperationSignatories.tsx
```diff
@@ -0,0 +1,171 @@
...
+export const OperationSignatories = ({ tx, connection, account }: Props) => {
+  const { t } = useI18n();
+  const { getLiveTxEvents } = useMultisigEvent({});
+
+  const { signatories, accountId, chainId, callHash, blockCreated, indexCreated } = tx;
+  const events = getLiveTxEvents(accountId, chainId, callHash, blockCreated, indexCreated);
+
+  const contacts = useUnit(contactModel.$contacts);
+  const accounts = useUnit(walletModel.$accounts);
+  const walletsMap = new Map(useUnit(walletModel.$wallets).map((w) => [w.id, w]));
```
Asmadek, line 35: you can use [useStoreMap](https://effector.dev/docs/api/effector-react/usestoremap/), but I am not sure what way is better

Asmadek (approved):

src/renderer/pages/Operations/components/OperationSignatories.tsx
egor0798: `useStoreMap`  description in docs says it's common use is to subscribe to changes in selected part of store only and that doesn't look like our case so I think just `new Map` constructor is fine

PR #1153, opened 2023-10-17
feat: ExportAddrV2 support

pgolovkin:
Support new data structure `ExportAddrV2` for dynamic derivations.
As a result 3 types of onboarding with Polkadot Vault are supported:
1. QR code with string
2. QR code with ExportAddrV1 -> lead to the multishard wallet flow
3. QR code with ExportAddrV2 (with only one root key) -> lead to dynamic derivations wallet flow

### DIFF
--- a/src/renderer/components/common/QrCode/QrReader/QrReader.tsx
+++ b/src/renderer/components/common/QrCode/QrReader/QrReader.tsx
@@ -130,7 +130,7 @@ const QrReader = ({
 
     if (fountainResult) {
       // decode the 1st frame --> it's a single frame QR
-      let result = EXPORT_ADDRESS.decode(fountainResult.slice(3));
+      const result = EXPORT_ADDRESS.decode(fountainResult.slice(3));
       isComplete.current = true;
       onResult?.(makeResultPayload(result.payload));
     } else {
@@ -184,7 +184,7 @@ const QrReader = ({
         continue;
       }
 
-      let result = EXPORT_ADDRESS.decode(fountainResult.slice(3));
+      const result = EXPORT_ADDRESS.decode(fountainResult.slice(3));
       isComplete.current = true;
       onResult?.(makeResultPayload(result.payload));
       break;
--- a/src/renderer/components/common/QrCode/QrReader/QrReader.tsx
+++ b/src/renderer/components/common/QrCode/QrReader/QrReader.tsx
@@ -6,7 +6,7 @@ import { useEffect, useRef } from 'react';
 import { cnTw, validateSignerFormat } from '@renderer/shared/lib/utils';
 import { CryptoTypeString } from '@renderer/shared/core';
 import { useI18n } from '@renderer/app/providers';
-import { ErrorFields, EXPORT_ADDRESS, EXPORT_ADDRESS_2, FRAME_KEY } from '../common/constants';
+import { ErrorFields, EXPORT_ADDRESS, FRAME_KEY } from '../common/constants';
 import { QR_READER_ERRORS } from '../common/errors';
 import { DecodeCallback, ErrorObject, Progress, QrError, SeedInfo, VideoInput } from '../common/types';
 import RaptorFrame from './RaptorFrame';
@@ -65,13 +65,9 @@ const QrReader = ({
   };
 
   const makeResultPayload = <T extends string | SeedInfo[] | { addr: SeedInfo }>(data: T): SeedInfo[] => {
-    if (typeof data !== 'string') {
-      if (data.constructor === Array) {
-        return data as SeedInfo[];
-      } else {
-        return [(data as { addr: SeedInfo }).addr as SeedInfo];
-      }
-    }
+    if (Array.isArray(data)) return data;
+
+    if (typeof data !== 'string') return [data.addr];
 
     return [
       {
@@ -111,7 +107,7 @@ const QrReader = ({
 
     isComplete.current = true;
     onProgress?.({ decoded: 1, total: 1 });
-    onResult?.(makeResultPayload<string>(signerAddress));
+    onResult?.(makeResultPayload(signerAddress));
 
     return true;
   };
@@ -134,14 +130,9 @@ const QrReader = ({
 
     if (fountainResult) {
       // decode the 1st frame --> it's a single frame QR
-      let result: { payload: string | SeedInfo[] | { addr: SeedInfo } };
-      try {
-        result = EXPORT_ADDRESS.decode(fountainResult.slice(3));
-      } catch (e) {
-        result = EXPORT_ADDRESS_2.decode(fountainResult.slice(3));
-      }
+      let result = EXPORT_ADDRESS.decode(fountainResult.slice(3));
       isComplete.current = true;
-      onResult?.(makeResultPayload<string | SeedInfo[] | { addr: SeedInfo }>(result.payload));
+      onResult?.(makeResultPayload(result.payload));
     } else {
       // if there is more than 1 frame --> proceed scanning and keep the progress
       onProgress?.({ decoded: 1, total: frameData.total });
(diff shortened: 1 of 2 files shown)


src/renderer/components/common/QrCode/common/constants.ts
```diff
@@ -36,8 +36,14 @@ const SEED_INFO: Codec<SeedInfo> = object(
...
 );
 
+const FEATURE = taggedUnion('VaultFeatures', [['BulkOperations'], ['DynamicDerivations']]);
+
 // Export address format for decoding; Rust enum is a tagged union
 export const EXPORT_ADDRESS = taggedUnion('ExportAddrs', [['V1', ['payload', array(SEED_INFO)]]]);
+export const EXPORT_ADDRESS_2 = taggedUnion('ExportAddrs', [
+  ['V1', ['payload', array(SEED_INFO)]],
+  ['V2', ['payload', object(['addr', SEED_INFO], ['features', array(FEATURE)])]],
+]);
```
valentunn, line 46: Why do we need a separate type for that? `EXPORT_ADDRESS_2` should decode `EXPORT_ADDRESS` payloads as well because its `V1` variant is exactly the same

src/renderer/components/common/QrCode/QrReader/QrReader.tsx
```diff
@@ -128,9 +134,14 @@ const QrReader = ({
 
     if (fountainResult) {
       // decode the 1st frame --> it's a single frame QR
-      const result = EXPORT_ADDRESS.decode(fountainResult.slice(3));
+      let result: { payload: string | SeedInfo[] | { addr: SeedInfo } };
```
valentunn, line 137: `string | SeedInfo[] | { addr: SeedInfo } ` this bit repeats multiple times so its worth extracting it into a separate type

src/renderer/components/common/QrCode/QrReader/QrReader.tsx
```diff
@@ -182,9 +193,14 @@ const QrReader = ({
         continue;
       }
 
-      const result = EXPORT_ADDRESS.decode(fountainResult.slice(3));
+      let result;
+      try {
+        result = EXPORT_ADDRESS.decode(fountainResult.slice(3));
+      } catch (e) {
+        result = EXPORT_ADDRESS_2.decode(fountainResult.slice(3));
+      }
```
valentunn, line 201: This bit repeats two times

src/renderer/components/common/QrCode/QrReader/QrReader.tsx
valentunn: Also, how `payload` here can possibly be a `string` given that we only assign it to `ExposreAddress.decode` or to `ExportAddress_2.decode`?

tuul-wq (changes requested):
Suggestions

src/renderer/components/common/QrCode/QrReader/QrReader.tsx
pgolovkin: It may be a `string` because the user may scan qr code with address. Like Nova Wallet does.

src/renderer/components/common/QrCode/QrReader/QrReader.tsx
```diff
@@ -128,9 +134,14 @@ const QrReader = ({
 
     if (fountainResult) {
       // decode the 1st frame --> it's a single frame QR
-      const result = EXPORT_ADDRESS.decode(fountainResult.slice(3));
+      let result: { payload: string | SeedInfo[] | { addr: SeedInfo } };
+      try {
+        result = EXPORT_ADDRESS.decode(fountainResult.slice(3));
```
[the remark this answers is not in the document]
pgolovkin: Moving all to `EXPORT_ADDRESS` enum.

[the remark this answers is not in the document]
valentunn: Ideally, we should have a single union (currently one named EXPORT_ADDRESS_2) that can decode every possible variant. An exception would mean that it is either an unknown variant for the current version of the app or the input is not correct at all

src/renderer/components/common/QrCode/QrReader/QrReader.tsx
```diff
@@ -128,9 +134,14 @@ const QrReader = ({
...
+      } catch (e) {
+        result = EXPORT_ADDRESS_2.decode(fountainResult.slice(3));
+      }
       isComplete.current = true;
-      onResult?.(makeResultPayload<SeedInfo[]>(result.payload));
+      onResult?.(makeResultPayload<string | SeedInfo[] | { addr: SeedInfo }>(result.payload));
```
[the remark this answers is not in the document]
pgolovkin: It works!

src/renderer/components/common/QrCode/QrReader/QrReader.tsx
pgolovkin: this is made for 1st and next frames handling. In the 1st frame the app is able to understand which kind of QR is present: with the string or Raptor encoded data structure.
In the 2nd and next frames handler the application parses the QR code and accumulates Raptor frames.

src/renderer/components/common/QrCode/QrReader/QrReader.tsx
```diff
@@ -128,9 +130,9 @@ const QrReader = ({
 
     if (fountainResult) {
       // decode the 1st frame --> it's a single frame QR
-      const result = EXPORT_ADDRESS.decode(fountainResult.slice(3));
+      let result = EXPORT_ADDRESS.decode(fountainResult.slice(3));
```
Asmadek, line 133: Why do we need let here?

src/renderer/components/common/QrCode/QrReader/QrReader.tsx
```diff
@@ -182,9 +184,9 @@ const QrReader = ({
         continue;
       }
 
-      const result = EXPORT_ADDRESS.decode(fountainResult.slice(3));
+      let result = EXPORT_ADDRESS.decode(fountainResult.slice(3));
```
Asmadek, line 187: And here?

Asmadek (approved):

src/renderer/components/common/QrCode/QrReader/QrReader.tsx
pgolovkin: this is made for 1st and next frames handling. In the 1st frame the app is able to understand which kind of QR is present: with the string or Raptor encoded data structure.
In the 2nd and next frames handler the application parses the QR code and accumulates Raptor frames.

src/renderer/components/common/QrCode/QrReader/QrReader.tsx
valentunn: Yea I meant we can extract this bit into a separate function

valentunn: P.S. I see that this code was not replaced with just EXPORT_ADDRESS.decode

valentunn (approved):

tuul-wq (approved):

egor0798 (approved):

src/renderer/components/common/QrCode/QrReader/QrReader.tsx
pgolovkin: `String` is possible because we're getting just a raw bytes from the camera device. We don't know what was scanned. It may be:
1. Raw string with address (like for Nova Wallet)
2. ExportAddrV1
3. ExportAddrV2
Depending on the decoding result the application will accordingly open the screen for:
1. Single Polkadot Vault pairing (legacy)
2. Multishard Polkadot Vault pairing (legacy)
3. Polkadot Vault pairing (with dynamic derivations)

PR #1164, opened 2023-10-20
Feat: signing screen update

egor0798:
## Changes:
- signatory, title and padding updated

***

image


Asmadek (approved):

src/renderer/pages/Operations/common/utils.ts
```diff
@@ -140,6 +141,25 @@ export const getModalTransactionTitle = (
   return TransactionTitlesModal[transaction.type](crossChain);
 };
 
+export const getMultisigSignOperationTitle = (
+  crossChain: boolean,
+  t: (key: string) => string,
```
tuul-wq, line 146: Can use `TFunction`

tuul-wq (approved):

PR #1166, opened 2023-10-20
Feat: forget wallet

Asmadek:
image


src/renderer/entities/wallet/model/__tests__/wallet-model.test.ts
```diff
@@ -100,4 +100,31 @@ describe('entities/wallet/model/wallet-model', () => {
...
+    });
+
+    expect(scope.getState(walletModel.$wallets)).toHaveLength(wallets.length);
+
+    await allSettled(walletModel.events.walletRemoved, {
+      scope,
+      params: removedWallet,
+    });
+
+    expect(deleteAccountsSpy).toHaveBeenCalledWith(removedAccounts.map((a) => a.id));
```
tuul-wq, line 124: It's ok to check service function call, but do we actually care about it?
It's more important to get the right `$wallets` after deletion completes

src/renderer/entities/wallet/model/__tests__/wallet-model.test.ts
```diff
@@ -100,4 +100,31 @@ describe('entities/wallet/model/wallet-model', () => {
...
+    expect(scope.getState(walletModel.$wallets)).toEqual(newWallets);
+    expect(scope.getState(walletModel.$accounts)).toEqual(
+      walletMock.accounts.filter((a) => a.walletId !== removedWallet.id),
```
tuul-wq, line 127: Good to place it near `108: const removedAccounts`

src/renderer/entities/wallet/model/wallet-model.ts
```diff
@@ -126,6 +127,20 @@ const multisigWalletUpdatedFx = createEffect(
   },
 );
 
+type RemoveProps = {
+  wallet: Wallet;
+  accounts: Account[];
```
tuul-wq, line 132: I see that we only require array of `id` from account, let's pass them instead of full accounts 😉

src/renderer/entities/wallet/model/wallet-model.ts
```diff
@@ -39,6 +39,7 @@ const walletConnectCreated = createEvent<CreateParams<WalletConnectAccount>>();
 
 const walletSelected = createEvent<WalletId>();
 const multisigAccountUpdated = createEvent<MultisigUpdateParams>();
+const walletRemoved = createEvent<Wallet>();
```
tuul-wq, line 42: Why not `WalletId` like we've done in `walletSelected`?
Good to provide as minimum data as needed 😊

src/renderer/entities/wallet/model/wallet-model.ts
```diff
@@ -203,6 +218,40 @@ sample({
   target: $accounts,
 });
 
+sample({
+  clock: walletRemoved,
+  source: $accounts,
+  fn: (accounts, wallet) => ({ accounts: accounts.filter((a) => a.walletId === wallet.id), wallet }),
```
tuul-wq, line 224: Time for utility function has come 😄
like:
```javascript
function getWalletAccounts(wallet: Wallet, accounts: Account[]): Account[]
```
maybe even generic one to get related accounts
like:
```javascript
function getWalletAccounts<T extends Account>(wallet: Wallet, accounts: T[]): T[]
```

src/renderer/entities/wallet/model/wallet-model.ts
```diff
@@ -203,6 +218,40 @@ sample({
...
+  target: removeWalletFx,
+});
+
+sample({
+  clock: removeWalletFx.doneData,
+  source: $wallets,
+  fn: (wallets, wallet) => wallets.filter((w) => w.id !== wallet.id),
```
tuul-wq, line 231: We don't need the whole `wallet`, `wallet.id` is enough

src/renderer/entities/wallet/model/wallet-model.ts
```diff
@@ -203,6 +218,40 @@ sample({
...
+  target: $wallets,
+});
+
+sample({
+  clock: removeWalletFx.doneData,
+  source: $accounts,
+  fn: (accounts, wallet) => accounts.filter((a) => a.walletId !== wallet.id),
```
tuul-wq, line 238: Would be interesting to use `spread` from `patronum`
So we receive `walletId`, `fn` returns `{ wallets: wallets.filter..., accounts: accounts.filter }`
`target` gets `spread`
line 197 in this file

src/renderer/entities/wallet/model/wallet-model.ts
```diff
@@ -203,6 +218,40 @@ sample({
...
+  target: $accounts,
+});
+
+sample({
+  clock: removeWalletFx.doneData,
+  source: {
+    activeWallet: $activeWallet,
+    wallets: $wallets,
```
tuul-wq, line 246: I think we could do even better
```javascript
sample({
  clock: walletSelected,
  source: $activeWallet,
  filter: (wallet, nextId) => wallet?.id !== nextId,
  fn: (wallet, nextId) => ({ prevId: wallet?.id, nextId }),
  target: walletSelectedFx,
});
```
Skip wallet selection for already active wallet inside `walletSelected` event
So you should do 
```javascript
forward({ from: removeWalletFx.doneData, to: walletSelected });
```

src/renderer/entities/walletConnect/model/wallet-connect-model.ts
```diff
@@ -271,16 +272,25 @@ forward({
...
 
 sample({
-  clock: disconnect,
-  source: {
-    client: $client,
-    session: $session,
+  clock: disconnectCurrentStarted,
+  source: $session,
+  filter: (session): session is SessionTypes.Struct => session !== null,
+  fn: (session) => session!.topic,
```
tuul-wq, line 278: Bang is not needed, isn't it?
```javascript
fn: (session) => session.topic
```

src/renderer/entities/walletConnect/model/wallet-connect-model.ts
```diff
@@ -271,16 +272,25 @@ forward({
...
+  target: disconnectStarted,
+});
+
+sample({
+  clock: disconnectStarted,
+  source: $client,
+  filter: (client, sessionTopic) => client !== null && !!client?.session.get(sessionTopic),
```
tuul-wq, line 285: This could work
```suggestion
  filter: (client, sessionTopic) => Boolean(client?.session.get(sessionTopic)),
```

src/renderer/entities/walletConnect/model/wallet-connect-model.ts
```diff
@@ -271,16 +272,25 @@ forward({
...
+  fn: (client, sessionTopic) => {
+    const session = client!.session.get(sessionTopic);
+
+    return {
```
tuul-wq, line 289:
```suggestion
    => ({
       client: client!.session.get(sessionTopic),
       session: session!,
    })
```

src/renderer/entities/walletConnect/model/wallet-connect-model.ts
```diff
@@ -330,7 +340,8 @@ export const walletConnectModel = {
   $pairings,
   events: {
     connect,
-    disconnect,
+    disconnectCurrentStarted,
```
tuul-wq, line 343: `Disconnect current ...` feels that noun is missing, current topic/session or something else?

src/renderer/pages/Onboarding/WalletConnect/NovaWallet.tsx
```diff
@@ -34,7 +34,7 @@ export const NovaWallet = ({ isOpen, onClose, onComplete }: Props) => {
   const pairings = useUnit(walletConnectModel.$pairings);
   const uri = useUnit(walletConnectModel.$uri);
   const connect = useUnit(walletConnectModel.events.connect);
-  const disconnect = useUnit(walletConnectModel.events.disconnect);
+  const disconnect = useUnit(walletConnectModel.events.disconnectCurrentStarted);
```
tuul-wq, line 37: Events don't require `useUnit`

src/renderer/shared/api/storage/service/storageService.ts
```diff
@@ -86,6 +86,10 @@ class StorageService<T extends { id: K }, K extends IndexableType> {
   delete(id: K): Promise<void> {
     return this.dexieTable.delete(id);
   }
+
+  deleteAll(ids: K[]): Promise<void> {
```
tuul-wq, line 90: 👍

tuul-wq (changes requested):

src/renderer/entities/walletConnect/model/wallet-connect-model.ts
Asmadek: TS show error without !

egor0798 (approved):

tuul-wq (approved):

PR #1167, opened 2023-10-20
Fix: wallet cleanup bugs

egor0798:
## Changes:
- Fixed validators table loading state
- Replaced ValidatorsModal on Staking overview with new component
- Notification dispalay when wallet deleted fix
- Nova wallet receive screen ffix


tuul-wq (approved):
Minor suggestions

src/renderer/pages/Staking/Overview/Overview.tsx
```diff
@@ -169,6 +180,22 @@ export const Overview = () => {
     return acc;
   }, []);
 
+  const [selectedValidators, notSelectedValidators] = nominators.reduce<[Validator[], Validator[]]>(
+    (acc, nominator) => {
+      if (validators[nominator.address]) {
+        acc[0].push({
```
Asmadek, line 186: what do you think to replace acc indexes with constants or use object instead of array?

Asmadek (approved):

src/renderer/pages/Staking/Overview/Overview.tsx
egor0798: Object is more convinient than array but I don't see the need to replace it here, easier to work with arrays as input for ValidatorsModal component

PR #1170, opened 2023-10-23
Fix: No redirect

tuul-wq:
Fixed redundant redirect on wallet change


tuul-wq:
Task linked: [CU-8692xy3a4 Switch wallets. Return user to the Assets page, when switch wallet](https://app.clickup.com/t/8692xy3a4)

Asmadek (approved):

PR #1172, opened 2023-10-24
Fix: wallet display on log screen

egor0798:
## Changes:
- Fixed wallets filtering for lock screen
- Fixed nova wallet icon

### DIFF
--- a/src/renderer/pages/Operations/components/LogModal.tsx
+++ b/src/renderer/pages/Operations/components/LogModal.tsx
@@ -9,13 +9,14 @@ import { TransactionTitle } from './TransactionTitle/TransactionTitle';
 import OperationStatus from './OperationStatus';
 import { getSignatoryName, getTransactionAmount, sortByDateAsc } from '../common/utils';
 import { BaseModal, BodyText, FootnoteText, Identicon } from '@renderer/shared/ui';
-import { toAddress, SS58_DEFAULT_PREFIX, getAssetById } from '@renderer/shared/lib/utils';
+import { getAssetById, SS58_DEFAULT_PREFIX, toAddress } from '@renderer/shared/lib/utils';
 import { ExtrinsicExplorers } from '@renderer/components/common';
 import { useMultisigEvent } from '@renderer/entities/multisig';
 import { MultisigTransactionDS } from '@renderer/shared/api/storage';
 import { AssetBalance } from '@renderer/entities/asset';
-import type { Account, MultisigAccount, Contact } from '@renderer/shared/core';
-import { WalletIcon, walletModel } from '@renderer/entities/wallet';
+import type { Account, Contact, MultisigAccount, Wallet } from '@renderer/shared/core';
+import { WalletType } from '@renderer/shared/core';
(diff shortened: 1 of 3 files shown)


src/renderer/pages/Operations/components/LogModal.tsx
```diff
@@ -35,11 +36,20 @@ const EventMessage: Partial<Record<SigningStatus | 'INITIATED', string>> = {
   ERROR_CANCELLED: 'log.errorCancelledMessage',
 } as const;
 
+const getFilteredWalletsMap = (wallets: Wallet[]): Map<Wallet['id'], Wallet> =>
+  new Map(
```
tuul-wq, line 40: Filling object with a single `reduce` is more straightforward
Also use utility function for this `w.type === WalletType.MULTISHARD_PARITY_SIGNER`

src/renderer/pages/Operations/components/LogModal.tsx
```diff
@@ -35,11 +36,20 @@ const EventMessage: Partial<Record<SigningStatus | 'INITIATED', string>> = {
...
+      .map((w) => [w.id, w]),
+  );
+
 const LogModal = ({ isOpen, onClose, tx, account, connection, contacts, accounts }: Props) => {
   const { t, dateLocale } = useI18n();
   const { getLiveTxEvents } = useMultisigEvent({});
-  const walletsMap = new Map(useUnit(walletModel.$wallets).map((w) => [w.id, w]));
-  const accountsMap = new Map(useUnit(walletModel.$accounts).map((a) => [a.accountId, a]));
+  const filteredWalletsMap = getFilteredWalletsMap(useUnit(walletModel.$wallets));
+  const filteredAccountMap = new Map(
```
tuul-wq, line 50: To reduce logic clutter in UI it's nice to make similar function as `getFilteredWalletsMap`

tuul-wq (changes requested):

src/renderer/pages/Operations/components/LogModal.tsx
```diff
@@ -35,11 +35,32 @@ const EventMessage: Partial<Record<SigningStatus | 'INITIATED', string>> = {
   ERROR_CANCELLED: 'log.errorCancelledMessage',
 } as const;
 
+type WalletsMap = Map<Wallet['id'], Wallet>;
+
+const getFilteredWalletsMap = (wallets: Wallet[]): WalletsMap =>
```
tuul-wq, line 40: Why not this one?
Arrow function without explicit `return` is ok for one-liners but here is another story

```javascript
type WalletsMap = Record<Wallet['id'], Wallet>;

const getFilteredWalletsMap = (wallets: Wallet[]): WalletsMap => {
  return wallets.reduce<WalletsMap>((acc, wallet) => {
    // 2nd condition for legacy multisig
    if (walletUtils.isValidSignatory(wallet) || walletUtils.isMultiShard(wallet)) {
      acc[wallet.id] = wallet;
    }

    return acc;
  }, {});
};
```

tuul-wq (changes requested):

Asmadek (approved):

tuul-wq (approved):

PR #1188, opened 2023-10-27
Fix: validators list, qr timer and confirmation fields order

egor0798:
## Changes:
- Updated countdown z-index
- Fixed validators modal on operation details
- Changed field order for operation confirmation page

### DIFF
--- a/src/renderer/shared/ui/Countdown/Countdown.tsx
+++ b/src/renderer/shared/ui/Countdown/Countdown.tsx
@@ -11,7 +11,7 @@ export const Countdown = ({ countdown, className }: Props) => {
   const { t } = useI18n();
 
   return (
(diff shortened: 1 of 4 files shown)


pgolovkin (approved):

src/renderer/shared/ui/Countdown/Countdown.tsx
```diff
@@ -11,7 +11,7 @@ export const Countdown = ({ countdown, className }: Props) => {
   const { t } = useI18n();
 
   return (
-    <div className={cnTw('flex items-center gap-x-2 mt-2 mb-6', className)}>
+    <div className={cnTw('flex items-center gap-x-2 mt-2 mb-6 z-1', className)}>
```
Asmadek, line 14: are you sure that we have `z-1`? only 0 and 10 there are in documentation. for custom value we should use `z-[1]`

Asmadek (approved):

PR #1220, opened 2023-11-03
Feat: pairing with new Polkadot Vault

Asmadek:
* detect exportAddrV2 code with dynamic derivations feature
* `ManageDynamicDerivations.tsx` file still in progress (copy of managemultishard now)


Asmadek:
Task linked: [CU-8692wrwwj Pairing with new Polkadot Vault QR code (ExportAddr V2)](https://app.clickup.com/t/8692wrwwj)

src/renderer/pages/Onboarding/Vault/ManageDynamicDerivations/ManageDynamicDerivations.tsx
```diff
@@ -0,0 +1,328 @@
...
+                          className="flex gap-2 justify-between items-center"
+                        >
+                          <div className="flex items-center">
+                            <div
+                              className={cnTw(
+                                'bg-divider w-[2px] h-[50px] ml-4',
+                                derivedKeyIndex === derivedKeys.length - 1 && 'h-[26px] mb-[24px]',
+                              )}
+                            ></div>
+                            <div className="bg-divider w-[8px] h-[2px]"></div>
```
sokolova-an, line 297: Do we need divider as a `div`? 
If we do need, let make it self closed `<div className="bg-divider w-[8px] h-[2px]"/>`

sokolova-an (commented):
If `ManageDynamicDerivations` is in progress and we don't want to improve it now maybe we can mark it somehow so we don't forget to cleanup it later?
Would be nice to move functions in a common file since we use them not only here or in a separate helper to have fewer function declarations

src/renderer/components/common/QrCode/common/constants.ts
```diff
@@ -36,7 +36,17 @@ const SEED_INFO: Codec<SeedInfo> = object(
   ['derivedKeys', array(ADDRESS_INFO)],
 );
 
-const FEATURE = taggedUnion('VaultFeatures', [['BulkOperations'], ['DynamicDerivations']]);
+export const enum VaultFeatureTitle {
```
tuul-wq, line 39: Why not simple `VautlFeatures`?

src/renderer/components/common/QrCode/common/types.ts
```diff
@@ -37,6 +38,10 @@ export type SeedInfo = {
   derivedKeys: AddressInfo[];
 };
 
+export type SeedInfoExtended = SeedInfo & {
```
tuul-wq, line 41: Seems strange that it's an extended type but the property is optional
Can we make it required? If not, can we put it into existing `SeedInfo` and make optional?

src/renderer/pages/Onboarding/Vault/Vault.tsx
```diff
@@ -39,16 +68,33 @@ const Vault = ({ isOpen, onClose, onComplete }: Props) => {
...
+  useEffect(() => {
+    if (!qrPayload) return;
+
+    if (isDynamicDerivationSupport(qrPayload[0])) {
+      setQrType(QrCodeType.DYNAMIC_DERIVATIONS);
+
+      return;
+    }
+
+    if (
```
tuul-wq, line 80: Can we check `qrPayload.length > 1` and go to `setQrType(QrCodeType.MULTISHARD);`
Otherwise it's `setQrType(QrCodeType.SINGLESHARD);`?

tuul-wq (changes requested):

egor0798 (approved):

src/renderer/pages/Onboarding/Vault/Vault.tsx
Asmadek: No, we should use such condition to support old multishard accounts

tuul-wq (approved):

sokolova-an (approved):

PR #1233, opened 2023-11-09
Fix: Correct balance assetId

tuul-wq:
Get correct `accountId` & `assetId` when balance received


Asmadek (approved):

sokolova-an (approved):

src/renderer/entities/asset/lib/balanceService.ts
```diff
@@ -260,10 +263,13 @@ export const useBalance = (): IBalanceService => {
 
     return method.multi(assetsTuples, (data: any[]) => {
       const newBalances = data.reduce((acc, accountInfo, index) => {
+        const accountIndex = index % accountIds.length;
+        const assetIndex = Math.floor(index / accountIds.length);
```
valentunn, line 267: We can extract this into a separate function

valentunn (approved):

PR #1237, opened 2023-11-10
Improve chain sorting tests

stepanLav:


src/renderer/entities/network/lib/__tests__/TestHelpers.ts
tuul-wq, line 1: Why file name in `PascalCase`?

tuul-wq (approved):

pgolovkin (approved):

Jess25051980 (approved):

PR #1245, opened 2023-11-14
Feat: Update dependencies

tuul-wq:
- Updated by hand a few dependencies that was mentioned by `dependabot`

### DIFF
--- a/src/main/factories/create.ts
+++ b/src/main/factories/create.ts
@@ -1,6 +1,6 @@
 import { BrowserWindow, BrowserWindowConstructorOptions, session } from 'electron';
 
-import { APP_CONFIG } from '@/app.config';
+import { APP_CONFIG } from '../../../app.config';
 import { ENVIRONMENT } from '../shared/constants';
 
 export function createWindow(settings: BrowserWindowConstructorOptions) {
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -1,11 +1,10 @@
-import { join } from 'path';
-import { BrowserWindow, shell, Menu } from 'electron';
+import { join, parse } from 'path';
+import { readdirSync, rmSync, renameSync } from 'fs';
+import { BrowserWindow, Menu, shell } from 'electron';
 import log, { LogFile } from 'electron-log';
 import windowStateKeeper from 'electron-window-state';
-import * as path from 'path';
-import * as fs from 'fs';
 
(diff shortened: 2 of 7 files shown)


src/main/main.ts
```diff
@@ -56,9 +55,10 @@ export async function MainWindow() {
   ENVIRONMENT.IS_DEV && window.webContents.openDevTools({ mode: 'bottom' });
 
   // Open urls in the user's browser
-  window.webContents.on('new-window', (event, url) => {
-    event.preventDefault();
-    shell.openExternal(url);
+  window.webContents.setWindowOpenHandler((details) => {
```
tuul-wq, line 58: Change to align with `Electron v27`

src/main/main.ts
```diff
@@ -78,22 +78,20 @@ export async function MainWindow() {
   return window;
 }
 
-function rotateLogs(oldLogFile: LogFile): void {
+function rotateLogs(oldLogFile: LogFile) {
```
tuul-wq, line 81: Code style changes

src/main/factories/create.ts
```diff
@@ -1,6 +1,6 @@
 import { BrowserWindow, BrowserWindowConstructorOptions, session } from 'electron';
 
-import { APP_CONFIG } from '@/app.config';
+import { APP_CONFIG } from '../../../app.config';
```
sokolova-an, line 3: why change it from `@/app.config` ? Let's have it as it was

src/main/main.ts
```diff
@@ -1,11 +1,10 @@
...
+import { join, parse } from 'path';
+import { readdirSync, rmSync, renameSync } from 'fs';
+import { BrowserWindow, Menu, shell } from 'electron';
 import log, { LogFile } from 'electron-log';
 import windowStateKeeper from 'electron-window-state';
-import * as path from 'path';
-import * as fs from 'fs';
 
-import { APP_CONFIG } from '@/app.config';
+import { APP_CONFIG } from '../../app.config';
```
sokolova-an, line 7: and here

sokolova-an (approved):

Asmadek (approved):

src/main/factories/create.ts
tuul-wq: This alias was not working for electron
Requires research

PR #1246, opened 2023-11-17
feat: network service refactoring and network model

Asmadek:
# First step of network refactoring 

* Move old `network` entitle to `oldNetwork`
* Refactor `chainSpecService`
* simplify `networkService`
* create `network-model` with effector model for networks
* create universal provider to avoid recreating provider on reconnect

Need to review 
* src/renderer/entities/network/model/network-model.ts
* src/renderer/entities/network/lib/provider/UniversalProvider.ts
* src/renderer/entities/network/lib/networkService.ts
* src/renderer/entities/network/lib/chainsService.ts
* src/renderer/entities/network/lib/chainSpecService.ts


src/renderer/entities/network/lib/common/types.ts
```diff
@@ -5,27 +5,6 @@ import { UnsubscribePromise } from '@polkadot/api/types';
 import { ConnectionType } from '@shared/core';
 import type { Connection, Chain, ChainId, RpcNode, HexString } from '@shared/core';
 
-// =====================================================
```
tuul-wq, line 8: 👍

src/renderer/entities/network/model/network-model.ts
```diff
@@ -0,0 +1,250 @@
...
+import { ApiPromise } from '@polkadot/api';
+
+import { Metadata, ProviderType, chainsService, networkService } from '../lib';
+import { ChainId, kernelModel } from '@renderer/shared/core';
+import { UniversalProvider } from '../lib/provider/UniversalProvider';
+
+const enum NetworkStatus {
+  CONNECTED = 'CONNECTED',
+  DISCONNECTED = 'DISCONNECTED',
+  ERROR = 'ERROR',
```
Asmadek, line 11: Not used now

src/renderer/entities/network/lib/networkService.ts
```diff
@@ -1,388 +1,115 @@
...
-    await updateConnectionState(chainId, {
-      customNodes: (connection.connection.customNodes || []).filter((node) => node.url !== rpcNode.url),
-    });
-  };
+const createWebsocketProvider = (
+  rpcUrls: string[],
+  chainId: ChainId,
+  getMetadata: (chainId: ChainId) => Promise<Metadata | undefined>,
+): ProviderInterface => {
+  // TODO: handle limited retries provider = new WsProvider(node.address, 5000, {1}, 11000);
```
valentunn, line 83: Is this TODO left for another PR?

src/renderer/entities/network/lib/networkService.ts
```diff
@@ -1,388 +1,115 @@
...
+  const CachedWsProvider = createCachedProvider(WsProvider, chainId, getMetadata);
```
valentunn, line 84: I saw that we're creating wrapping both sc and ws providers in cached provider separately. Might be worth to wrap in a single place instead 

``` typescript
const provider = createWebsocketProvider/createScProvider()
const cachedMetadataProvider = createCachedProvider(provider,...)
```

src/renderer/entities/network/lib/networkService.ts
```diff
@@ -1,388 +1,115 @@
...
+  nodes: string[],
+  getMetadata: (chainId: ChainId) => Promise<Metadata | undefined>,
+  onConnected?: (value?: any) => void,
+  onDisconnected?: (value?: any) => void,
+  onError?: (value?: any) => void,
+): UniversalProvider {
+  let provider;
+
+  try {
+    const scProvider = createSubstrateProvider(chainId, getMetadata);
```
valentunn, line 45: Does SC provider do anything hevy until its connected? If it does, we probably want to delay its initialization until we really need to connect to SC, to avoid extra load on the app start

src/renderer/entities/network/lib/networkService.ts
```diff
@@ -1,388 +1,115 @@
...
-      setConnections(connectionsMap);
-    } catch (error) {
-      console.error(error);
-    }
-  };
+  try {
+    await provider?.disconnect();
+  } catch (e) {
+    console.warn(e);
+  }
```
valentunn, line 100: Do we need to call disconnect on bath `ApiPromise` and `ProviderInterface`? Doesnt `ApiPromise.disconnect()` also disconnect its ProviderInterface?

src/renderer/entities/network/model/network-model.ts
```diff
@@ -0,0 +1,250 @@
...
+  return provider;
+});
+
+type CreateApiParams = {
+  chainId: ChainId;
+  provider: UniversalProvider;
+};
+const createApiFx = createEffect(async ({ chainId, provider }: CreateApiParams): Promise<ApiPromise | undefined> => {
+  try {
+    const api = await networkService.createApi(provider);
```
tuul-wq, line 128:
```suggestion
  return networkService.createApi(provider);
```

src/renderer/entities/network/model/network-model.ts
```diff
@@ -0,0 +1,250 @@
...
+
+type ConnectParams = {
+  provider: UniversalProvider;
+  api: ApiPromise;
+  chainId: ChainId;
+};
+const connectFx = createEffect(async ({ provider, api, chainId }: ConnectParams): Promise<ChainId> => {
+  await networkService.connect(provider, api);
+
+  return chainId;
```
tuul-wq, line 144: Why do we return `chainid` after connection?

src/renderer/entities/network/model/network-model.ts
```diff
@@ -0,0 +1,250 @@
...
+sample({
+  clock: initConnection,
+  fn: (chainId: ChainId) => ({
+    chainId,
+    nodes: chains[chainId].nodes.map((node) => node.url),
+  }),
+  target: createProviderFx,
+});
+
+sample({
```
tuul-wq, line 172: No `target` is that ok?

src/renderer/entities/network/model/network-model.ts
```diff
@@ -0,0 +1,250 @@
...
+    provider: providers[chainId],
+    api: apis[chainId],
+    chainId: chainId,
+  }),
+  target: disconnectFx,
+});
+
+const switchProviderTypeFx = createEffect(
+  async ({ provider, type }: { provider: UniversalProvider; type: ProviderType }) => {
+    await provider.setProviderType(type);
```
tuul-wq, line 226: `public setProviderType(providerType: ProviderType) {` is a synchronous  function
So why do we call `await` here?

tuul-wq (changes requested):
Nice 😊

src/renderer/entities/network/model/network-model.ts
```diff
@@ -0,0 +1,250 @@
...
+  // const networkErrorBinded = scopeBind(networkError);
+
+  const provider = networkService.createProvider(
+    chainId,
+    nodes,
+    getMetadata,
+    () => {
+      console.info('🟢 provider connected ==> ', chainId);
+
+      networkConnected(chainId);
```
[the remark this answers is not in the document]
Asmadek: `scopeBind` doesn't work with effects, only in `.watch`

Otherwise, it works well. I don't understand why I can't use it when it works

egor0798 (approved):

PR #1247, opened 2023-11-17
Fix: Removed ShardedAccount

egor0798:
Remove `ShardedAccount` and generate account structure

### DIFF
--- a/src/renderer/entities/wallet/lib/account-utils.ts
+++ b/src/renderer/entities/wallet/lib/account-utils.ts
@@ -79,22 +79,24 @@ function getWalletAccounts<T extends Account>(walletId: Wallet['id'], accounts: 
   return accounts.filter((account) => account.walletId === walletId);
 }
 
-function getAccountsWithGroupedShards<T extends Account>(accounts: T[]): Array<T | Array<T>> {
-  const shardsIndexes: Record<ShardAccount['group'], number> = {};
+function getAccountsWithGroupedShards(
+  accounts: Array<ChainAccount | ShardAccount>,
+): Array<ChainAccount | ShardAccount[]> {
+  const shardsIndexes: Record<ShardAccount['groupId'], number> = {};
 
-  return accounts.reduce<Array<T | Array<T>>>((acc, account, index) => {
+  return accounts.reduce<Array<ChainAccount | ShardAccount[]>>((acc, account) => {
     if (!accountUtils.isShardAccount(account)) {
       acc.push(account);
 
       return acc;
     }
 
-    const existingGroupIndex = shardsIndexes[account.group];
+    const existingGroupIndex = shardsIndexes[account.groupId];
     if (existingGroupIndex !== undefined) {
-      (acc[existingGroupIndex] as Array<T>).push(account);
+      (acc[existingGroupIndex] as ShardAccount[]).push(account);
     } else {
       acc.push([account]);
-      shardsIndexes[account.group] = index;
+      shardsIndexes[account.groupId] = acc.length - 1;
     }
 
     return acc;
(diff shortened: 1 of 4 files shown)


src/renderer/shared/core/types/account.ts
```diff
@@ -29,7 +24,7 @@ export type ChainAccount = BaseAccount & {
 };
 
 export type ShardAccount = BaseAccount & {
-  shardedId: ID;
+  group: ID;
```
sokolova-an, line 27: we also decided to add keyType to ShardAcсount, right?

agree with your message in tg - `groupId` sounds better here

src/renderer/entities/wallet/lib/account-utils.ts
```diff
@@ -83,3 +78,25 @@ function getAllAccountIds(accounts: Account[], chainId: ChainId): AccountId[] {
 function getWalletAccounts<T extends Account>(walletId: Wallet['id'], accounts: T[]): T[] {
   return accounts.filter((account) => account.walletId === walletId);
 }
+
+function getAccountsWithGroupedShards<T extends Account>(accounts: T[]): Array<T | Array<T>> {
+  const shardsIndexes: Record<ShardAccount['group'], number> = {};
```
sokolova-an, line 83: maybe use `Set` here?

in this case, we can check the following conditions with: `shardsSet.has(account.group)`

sokolova-an (commented):
good job!

src/renderer/shared/core/types/account.ts
tuul-wq: yeah, `groupId` is a nice one

src/renderer/entities/wallet/lib/account-utils.ts
```diff
@@ -83,3 +78,25 @@ function getAllAccountIds(accounts: Account[], chainId: ChainId): AccountId[] {
 function getWalletAccounts<T extends Account>(walletId: Wallet['id'], accounts: T[]): T[] {
   return accounts.filter((account) => account.walletId === walletId);
 }
+
+function getAccountsWithGroupedShards<T extends Account>(accounts: T[]): Array<T | Array<T>> {
```
tuul-wq, line 82: Please add test

tuul-wq, line 82:
```suggestion
function getAccountsAndShardGroups<T extends Account>(accounts: T[]): Array<T | Array<T>> {
```

tuul-wq (approved):
Approved with a couple of comments 👍

src/renderer/shared/core/types/account.ts
tuul-wq: Also not sure about simple `ID`, we discussed `uuid` so it won't intersect with other `groupId's`

src/renderer/entities/wallet/lib/account-utils.ts
Asmadek: Good solution `set.has` has `O(1)` time complexity, but I think here we store indexes for shards array, and can't use such solution

Asmadek (approved):

sokolova-an (approved):

src/renderer/shared/core/types/account.ts
egor0798: changed groupId to string and added keyType

src/renderer/entities/wallet/lib/account-utils.ts
egor0798: yeah, we need pairs of groupId - index so we cant use set here

src/renderer/entities/wallet/lib/account-utils.ts
egor0798: done

tuul-wq (approved):

PR #1258, opened 2023-11-28
Transaction builder draft

valentunn:


src/renderer/entities/transactionBuilder/lib/compound-wallet.ts
```diff
@@ -0,0 +1,60 @@
...
+    api: ApiPromise,
+    wallet: Wallet,
+    shards: Account[],
+    innerFactory: (shard: AccountInWallet) => TransactionBuilder
+  ) {
+    this.wallet = wallet
+    this.shards = shards
+    this.api = api
+
+    if (shards.length == 0) throw new Exception("Empty shard list")
```
tuul-wq, line 34: `Exception` is not a native web object, use `new Error('_text_')` instead

src/renderer/entities/transactionBuilder/lib/leaf.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  addCall(call: CallBuilding): void {
+    this.#calls.push(call)
+  }
+
+  resetCalls(): void {
+    this.#calls = []
+  }
+
+  setCall(call: CallBuilding): void {
```
tuul-wq, line 41: What's the difference between `addCall` & `setCall`?

src/renderer/entities/transactionBuilder/lib/compound-wallet.ts
```diff
@@ -0,0 +1,60 @@
...
+import {ApiPromise} from "@polkadot/api";
+import {BaseTxInfo, OptionsWithMeta, UnsignedTransaction} from "@substrate/txwrapper-polkadot";
+import {SubmittableExtrinsic} from "@polkadot/api/types";
+
+export class CompoundWalletTransactionBuilder implements TransactionBuilder {
+
+  effectiveCallBuilder: CallBuilder;
+
+  readonly wallet: Wallet
+  readonly shards: Account[]
```
tuul-wq, line 18: We have a `ShardAccount` for such case

src/renderer/entities/transactionBuilder/lib/compound-wallet.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  readonly api: ApiPromise
+
+  #inner: TransactionBuilder
```
tuul-wq, line 22: What about `internalBuilder`?

src/renderer/entities/transactionBuilder/lib/compound-wallet.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  constructor(
+    api: ApiPromise,
+    wallet: Wallet,
+    shards: Account[],
+    innerFactory: (shard: AccountInWallet) => TransactionBuilder
```
tuul-wq, line 28:
```suggestion
    internalFactory: (compoundWallet: AccountInWallet) => TransactionBuilder
```

tuul-wq (commented):
First part of questions

src/renderer/entities/transactionBuilder/lib/leaf.ts
valentunn: `setCall` is the shortcut for `reseset + addCall`. In general those methods are just my idea what might potentially be needed; it is quite easy to alter them

src/renderer/entities/transactionBuilder/lib/compound-wallet.ts
valentunn: Yeah saw it too. However I was not quite sure whether I can safely cast Account to ShardAccount in all cases  CompoundWalletTransactionBuilder is used

src/renderer/entities/transactionBuilder/lib/compound-wallet.ts
valentunn: For me `internal` sounds a bit confusing, since it creates an association with `somethihg hidden from the outside`. Whereas this meaning is indeed true, imho it is not the main one - what we are really referring to here is that `inner` is the child tx builder that is wrapped by ` CompoundWalletTransactionBuilder`. What do you think about `nested` or `child` naming?

src/renderer/entities/transactionBuilder/lib/compound-wallet.ts
```diff
@@ -0,0 +1,60 @@
...
+import {AccountInWallet, Wallet} from "@shared/core/types/wallet";
+import {Account} from "@shared/core";
+import {Exception} from "@zxing/library";
+import {ApiPromise} from "@polkadot/api";
+import {BaseTxInfo, OptionsWithMeta, UnsignedTransaction} from "@substrate/txwrapper-polkadot";
+import {SubmittableExtrinsic} from "@polkadot/api/types";
+
+export class CompoundWalletTransactionBuilder implements TransactionBuilder {
+
+  effectiveCallBuilder: CallBuilder;
```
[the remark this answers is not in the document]
valentunn: Maybe somtihg like `realCallBuilder` or `userCallBuilder`?

src/renderer/entities/transactionBuilder/lib/compound-wallet.ts
```diff
@@ -0,0 +1,78 @@
...
+import {BaseTxInfo, OptionsWithMeta, UnsignedTransaction} from "@substrate/txwrapper-polkadot";
+import {SubmittableExtrinsic} from "@polkadot/api/types";
+import {LeafTransactionBuilder} from "@entities/transactionBuilder/lib/leaf";
+
+export class CompoundWalletTransactionBuilder implements TransactionBuilder {
+
+  readonly wallet: Wallet
+  readonly allChildrenAccounts: Account[]
+
+  #selectedSChildrenAccounts: Account[]
```
valentunn, line 18: Typo

src/renderer/entities/transactionBuilder/lib/compound-wallet.ts
valentunn: Agreed on `innerBuilder`

src/renderer/entities/transactionBuilder/lib/compound-wallet.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  readonly api: ApiPromise
+
+  #inner: TransactionBuilder
```
valentunn, line 22: Adapt javascript visibility (`#`) to typescript one

src/renderer/entities/transactionBuilder/lib/leaf.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  async unsignedTransaction(options: OptionsWithMeta, info: BaseTxInfo): Promise<UnsignedTransaction> {
+    const nestedUnsignedTxs = this.currentCalls.map(call => call.signing(info, options))
+
+    if (nestedUnsignedTxs.length == 0) throw new Error("Cannot sign empty transaction")
+
+    let maybeWrappedInBatch: UnsignedTransaction
+    if (nestedUnsignedTxs.length > 1) {
+      const innerMethods = nestedUnsignedTxs.map((nestedUnsignedTx) => nestedUnsignedTx.method)
+      maybeWrappedInBatch = methods.utility.batch({calls: innerMethods,}, info, options)
```
valentunn, line 64: batchAll

src/renderer/entities/transactionBuilder/lib/leaf.ts
```diff
@@ -0,0 +1,87 @@
...
+    } else {
+      maybeWrappedInBatch = nestedUnsignedTxs[0]
+    }
+
+    return maybeWrappedInBatch
+  }
+
+  async submittableExtrinsic(): Promise<SubmittableExtrinsic<"promise"> | null> {
+    const feeCalls = this.currentCalls.map(call => call.fee)
```
valentunn, line 73: oldNaming

src/renderer/entities/transactionBuilder/lib/multisig.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+    this.#selectedSignatory = knownSignatoryAccounts[0]
+    this.#inner = innerFactory(this.#selectedSignatory)
+  }
+
+  effectiveCallBuilder(): CallBuilder {
+    return this.#inner.effectiveCallBuilder()
+  }
+
+  visitAll(visitor: Partial<TransactionVisitor>): void {
```
valentunn, line 54: Remove partial

src/renderer/entities/transactionBuilder/lib/factory.ts
```diff
@@ -0,0 +1,75 @@
...
+): TransactionBuilder {
+
+  const walletsById = keyBy(allWallets, 'id')
+  const accountsByAccountId = groupBy(allAccounts, "accountId")
+
+  const createInner: NestedTransactionBuilderFactory = (accountInWallet: AccountInWallet) => {
+    const wallet = accountInWallet.wallet
+
+    switch (walletUtils.getWalletFamily(wallet)) {
+      case WalletType.MULTISIG:
```
valentunn, line 29: WalletFamily

valentunn: P.S. only WalletType is possible

src/renderer/entities/transactionBuilder/lib/helpers/helpers.ts
```diff
@@ -0,0 +1,88 @@
...
+}
+
+export async function getDeposits(transactionBuilder: TransactionBuilder): Promise<AmountReduction> {
+  const reductionBuilder = buildAmountReduction()
+
+  const multisigDeposit = await getMultisigDeposit(transactionBuilder.api)
+
+  transactionBuilder.visitAll({
+    visitMultisig(visit: MultisigVisit) {
+      // TODO check who pays deposit in case multisig is wrapped in some other structure, like proxy
```
Tbaut, line 60: If that helps, the submitter always pays the deposit for the multisig transaction state. Now depending on the action performed, there may be other deposit required. for instance:
- A multisig is creating a proxy. The submitter needs ~20 DOTs for the multisig tx, but the Multisig that will then create the proxy will also need 20 DOTs for the proxy creation deposit
- A Multisig with pure proxy sets an identity. The submitter needs ~20 DOTs for the multisig tx, but the Multisig that will then set an will also need 20 DOTs for the identity deposit

TL;DR, the depositor/submitter needs the multisig tx deposit, but the transaction may fail when being submitted by the last signatory, because of some other deposit.

Jess25051980 (approved):

pgolovkin:
@valentunn don't you mind if I close this PR?

PR #1265, opened 2023-11-28
Fix: staking dd

sokolova-an:


src/renderer/pages/Staking/Overview/Overview.tsx
```diff
@@ -112,7 +114,7 @@ export const Overview = () => {
     const isWalletConnect = walletUtils.isWalletConnect(activeWallet);
     const isPolkadotVault = walletUtils.isPolkadotVault(activeWallet);
 
-    if (isMultisig || isNovaWallet || isWalletConnect || isPolkadotVault) {
+    if (isMultisig || isNovaWallet || isWalletConnect || (isPolkadotVault && addresses.length < 2)) {
```
tuul-wq, line 117: `=== 1` is a more straightforward condition, `< 2` makes me wonder what can go wrong with 2 address
```suggestion
    if (isMultisig || isNovaWallet || isWalletConnect || (isPolkadotVault && addresses.length === 1)) {
```

src/renderer/pages/Staking/Overview/Overview.tsx
```diff
@@ -57,6 +58,7 @@ export const Overview = () => {
 
   const accounts = activeAccounts.reduce<Account[]>((acc, account) => {
     if (accountUtils.isChainIdMatch(account, chainId)) {
+      if (accountUtils.isBaseAccount(account) && walletUtils.isPolkadotVault(activeWallet)) return acc;
```
tuul-wq, line 61: Does it need to be nested?
If chain match or not for BaseAccount and PV wallet we will `return acc` anyway

tuul-wq (changes requested):

src/renderer/pages/Staking/Overview/Overview.tsx
sokolova-an: For `BaseAccount`  in inChainIdMatch we return true always which is okay if it's not the case for Polkadaot Vault. With this nested condition, it's more readable that we want to skip adding BaseAccount  only if it's PolkadotVault

egor0798 (approved):

tuul-wq (approved):

PR #1274, opened 2023-12-03
Feat: Vault details

tuul-wq:
- Polkadot Vault details
- 4 new components: RootAccountLg/Md, DerivedAccount, ContextMenu
- Style fixes for Vault onboarding

Screenshot 2023-12-03 at 2 21 41 PM


tuul-wq:
Task linked: [CU-8692x3zz5 Show Polkadot Vault wallet details](https://app.clickup.com/t/8692x3zz5)

src/renderer/entities/wallet/ui/Cards/DerivedAccount.tsx
```diff
@@ -0,0 +1,132 @@
...
+            )}
+            name="info"
+            onClick={onInfoClick}
+          />
+        )}
+
+        <div
+          className={cnTw(
+            'absolute right-0 pl-2 transition-all opacity-0 bg-white',
+            'group-hover:bg-[#f4f4f8] group-focus:bg-[#f4f4f8]',
```
Asmadek, line 121: Maybe add this color to pallet?

src/renderer/shared/ui/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,73 @@
...
+          {children}
+        </Popover.Panel>
+      </Transition>
+    </Popover>
+  );
+};
+
+type GroupProps = {
+  title?: string;
+  active?: boolean;
```
Asmadek, line 60: Just a question. What is the case for `active` prop?

src/renderer/widgets/WalletDetails/model/vault-details-model.ts
```diff
@@ -0,0 +1,38 @@
...
+const shardsCleared = createEvent();
+
+const chainSetFx = createEffect((chainId: ChainId): Chain | undefined => {
+  return chainsService.getChainById(chainId);
+});
+
+const $shards = createStore<ShardAccount[]>([]).reset(shardsCleared);
+const $chain = createStore<Chain>({} as Chain).reset(shardsCleared);
+
+forward({ from: shardsSelected, to: $shards });
```
Asmadek, line 16: I think better to avoid forward, as recommendation from effector team

src/renderer/widgets/WalletDetails/model/vault-details-model.ts
```diff
@@ -0,0 +1,38 @@
+import { createStore, createEvent, forward, createEffect, sample } from 'effector';
+
+import type { ShardAccount, Chain, ChainId } from '@shared/core';
+import { chainsService } from '@entities/network';
+
+const shardsSelected = createEvent<ShardAccount[]>();
+const shardsCleared = createEvent();
+
+const chainSetFx = createEffect((chainId: ChainId): Chain | undefined => {
```
Asmadek, line 9: I think we should discuss order events/store/effects, because I thought that we have events -> stores -> effects

src/renderer/widgets/WalletDetails/model/wallet-provider-model.ts
```diff
@@ -72,6 +72,38 @@ const $multisigAccount = combine(
...
+
+      const chainId = accountToInsert.chainId;
+      if (acc[chainId]) {
+        acc[chainId].push(account);
+      } else {
+        acc[chainId] = [account];
+      }
+
+      return acc;
+    }, {});
```
Asmadek, line 101: Looks like utility function

src/renderer/widgets/WalletDetails/model/vault-details-model.ts
```diff
@@ -0,0 +1,38 @@
...
+  return chainsService.getChainById(chainId);
```
Asmadek, line 10: Get chain by id it sync function, is it really needed to use it in `effect`, not in `fn`?

src/renderer/widgets/WalletDetails/ui/VaultWalletDetails.tsx
```diff
@@ -0,0 +1,70 @@
...
+              </div>
+            </ContextMenu.Group>
+          </ContextMenu>
+        </div>
+
+        <VaultAccountsList
+          className="h-[381px]"
+          chains={chains}
+          accountsMap={accountsMap}
+          onShardClick={vaultDetailsModel.events.shardsSelected}
```
Asmadek, line 63: As I understand I can open only one sharded account at a time. Is it correct?

Asmadek (approved):

src/renderer/widgets/WalletDetails/ui/VaultWalletDetails.tsx
tuul-wq: Yes, sure

src/renderer/shared/ui/ContextMenu/ContextMenu.tsx
tuul-wq: If we need to hide a group, instead of `{isActive && <ContextMenu.Group>}` we can do this -
```js
<ContextMenu.Group active={explorers.length > 0}>
```
Same `active` flag is used in `<Alert>`

src/renderer/widgets/WalletDetails/model/vault-details-model.ts
tuul-wq: Good point, some stores might depend on events (`reset` events for example)

src/renderer/widgets/WalletDetails/model/vault-details-model.ts
tuul-wq: One day we will create `chains model` and use chains in source
Right now I think it's more correct to call services inside effects, because they do some underground work like network API call
ChainsService only access `chains.json` file but still, good to have all side effects from services in effects

PR #1275, opened 2023-12-04
Fix: Dynamic derivation fixes

tuul-wq:
Fixed:
- Trim long names in constructor
- Reset errors
- Don't reset network after form submission
- Popover positioned in center
- "Edit keys" button text
- More tests


tuul-wq:
Task linked: [CU-869386ub2 DynamicDerivations. Add more keys. There is not trim for long names](https://app.clickup.com/t/869386ub2)

tuul-wq:
Task linked: [CU-86937q80f DynamicDerivations. Add more keys. Should not reset network to default after add key](https://app.clickup.com/t/86937q80f)

Asmadek (approved):

PR #1276, opened 2023-12-04
Fix: staking shards

egor0798:
## Changes:
- Allow staking with root shard for legacy multishards
- Fix staking for singleshard


Asmadek (approved):

src/renderer/pages/Staking/Overview/Overview.tsx
```diff
@@ -57,7 +67,8 @@ export const Overview = () => {
   const explorers = activeChain?.explorers;
 
   const accounts = activeAccounts.reduce<Account[]>((acc, account) => {
```
tuul-wq, line 69:
```javascript
  const accounts = activeAccounts.filter((account) => {
    const isBaseAccount = accountUtils.isBaseAccount(account);
    const isPolkadotVault = walletUtils.isPolkadotVault(activeWallet);
    const hasManyAccounts = activeAccounts.length > 1;

    if (isPolkadotVault && isBaseAccount && hasManyAccounts) return false;

    return accountUtils.isChainIdMatch(account, chainId);
  }, []);

```

tuul-wq (approved):
Approve with code suggestion

PR #1279, opened 2023-12-05
Fix: Receive selector

tuul-wq:
Hide root account for Polkadot Vault in receive modal selector
Show shards addresses

image


tuul-wq:
Task linked: [CU-86937vy4b Update shard selectr on the Receive screen for the Polkadot Vault wallet](https://app.clickup.com/t/86937vy4b)

PR #1280, opened 2023-12-05
Fix: change dd-template to link to github

egor0798:
Instead of file download dd-template link now leads to file on github


src/renderer/features/wallets/ImportKeys/ui/ImportKeysModal.tsx
```diff
@@ -92,7 +90,12 @@ export const ImportKeysModal = ({ isOpen, rootAccountId, existingKeys, onConfirm
           ))}
         </Alert>
 
-        <InfoLink url={templateFile} className="gap-x-1 mt-2 px-3" iconName="import" iconPosition="right" download>
+        <InfoLink
+          url="https://github.com/novasamatech/nova-spektr-utils/blob/main/templates/dd-template.yaml"
```
Asmadek, line 94: I think better to create constant for url

Asmadek (approved):

src/renderer/features/wallets/ImportKeys/ui/ImportKeysModal.tsx
tuul-wq: Agree, `features/wallets/lib/constants` or just here above the component
And use raw link > `https://raw.githubusercontent.com/novasamatech/nova-spektr-utils/main/templates/dd-template.yaml` otherwise CORS would block it 😉

tuul-wq (changes requested):

PR #1282, opened 2023-12-06
Fix: WOW can be used in transfer immediately after switch wallet

egor0798:
Close send asset modal if active wallet is WOW

### DIFF
--- a/src/renderer/entities/asset/ui/AssetCard/AssetCard.tsx
+++ b/src/renderer/entities/asset/ui/AssetCard/AssetCard.tsx
@@ -13,17 +13,18 @@ import { ChainId, Asset, Balance } from '@shared/core';
 import { TokenPrice } from '@entities/price/ui/TokenPrice';
 import { AssetFiatBalance } from '@entities/price/ui/AssetFiatBalance';
 import { priceProviderModel } from '@entities/price';
+import { walletModel, walletUtils } from '@entities/wallet';
 
 type Props = {
   chainId: ChainId;
   asset: Asset;
   balance?: Balance;
-  canMakeActions?: boolean;
 };
 
-export const AssetCard = ({ chainId, asset, balance, canMakeActions }: Props) => {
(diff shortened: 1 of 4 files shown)


egor0798:
Task linked: [CU-86937yycc Transfers. WOW can be used in transfer immediately after switch wallet](https://app.clickup.com/t/86937yycc)

src/renderer/entities/asset/ui/AssetCard/AssetCard.tsx
```diff
@@ -71,7 +72,7 @@ export const AssetCard = ({ chainId, asset, balance, canMakeActions }: Props) =>
             </div>
           )}
         </div>
-        {canMakeActions && (
+        {!walletUtils.isWatchOnly(activeWallet) && (
```
Asmadek, line 75: You can create variable in model, like `$activeWallet` or `$activeAccounts` but they have different styles and better to ask @tuul-wq which one is correct

Asmadek (approved):

src/renderer/entities/asset/ui/AssetCard/AssetCard.tsx
tuul-wq: We can use local model variable when effector model will be implemented for `Assets` page
For now it's absolutely ok to use domain function, especially without props drill 👍

src/renderer/widgets/SendAssetModal/ui/SendAssetModal.tsx
```diff
@@ -56,6 +57,12 @@ export const SendAssetModal = ({ chain, asset }: Props) => {
     priceProviderModel.events.assetsPricesRequested({ includeRates: true });
   }, []);
 
+  useEffect(() => {
```
tuul-wq, line 60: Is it event possible for wallet to change when modal is opened? 🤔

tuul-wq (approved):

PR #1292, opened 2023-12-07
Fix: dd multishard staking

egor0798:


src/renderer/components/common/Scanning/ScanMultiframeQr.tsx
```diff
@@ -62,9 +68,20 @@ const ScanMultiframeQr = ({
       return (async () => {
         const { payload, unsigned } = await createPayload(transactions[index], api);
 
+        const signPayload =
```
tuul-wq, line 71: I guess `createSubstrateSignPayload` works better and already contains needed condition to determine the right payload creator function

tuul-wq (changes requested):

PR #1297, opened 2023-12-08
fix: balances and networks refactoring

Asmadek:
`Network` & `Balances` optimization

### DIFF
--- a/src/renderer/entities/balance/lib/common/constants.ts
+++ b/src/renderer/entities/balance/lib/common/constants.ts
@@ -1,2 +1,2 @@
 export const VERIFY_TIMEOUT = 5000;
-export const SAVE_TIMEOUT = 5000;
+export const SAVE_TIMEOUT = 10000;
--- a/src/renderer/entities/network/model/network-model.ts
+++ b/src/renderer/entities/network/model/network-model.ts
@@ -327,25 +327,16 @@ sample({
   fn: (chains, connections) => {
     const connectionsMap = keyBy(connections, 'chainId');
 
-    const newConnections = Object.entries(chains).map(([chainId, chain]) => {
-      const connection = connectionsMap[chainId];
+    return Object.entries(chains).reduce<Record<ChainId, Connection>>((acc, [chainId, chain]) => {
+      acc[chain.chainId] = connectionsMap[chainId] || {
+        chainId: chain.chainId,
+        canUseLightClient: false,
+        connectionType: ConnectionType.AUTO_BALANCE,
+        customNodes: [],
+      };
 
-      if (connection) {
-        return [chainId, connection];
-      }
-
-      return [
-        chainId,
-        {
-          chainId: chain.chainId,
-          canUseLightClient: false,
-          connectionType: ConnectionType.AUTO_BALANCE,
-          customNodes: [],
-        },
-      ];
-    });
-
-    return Object.fromEntries(newConnections);
+      return acc;
+    }, {});
   },
   target: $connections,
 });
--- a/src/renderer/entities/network/model/network-model.ts
+++ b/src/renderer/entities/network/model/network-model.ts
@@ -323,7 +323,30 @@ sample({
 
 sample({
   clock: populateConnectionsFx.doneData,
-  fn: (connections) => keyBy(connections, 'chainId'),
+  source: $chains,
+  fn: (chains, connections) => {
+    const connectionsMap = keyBy(connections, 'chainId');
+
+    const newConnections = Object.entries(chains).map(([chainId, chain]) => {
+      const connection = connectionsMap[chainId];
+
+      if (connection) {
+        return [chainId, connection];
+      }
+
+      return [
+        chainId,
+        {
+          chainId: chain.chainId,
+          canUseLightClient: false,
+          connectionType: ConnectionType.AUTO_BALANCE,
+          customNodes: [],
+        },
+      ];
+    });
+
+    return Object.fromEntries(newConnections);
+  },
   target: $connections,
 });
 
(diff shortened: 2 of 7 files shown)


egor0798 (approved):

src/renderer/entities/balance/lib/common/constants.ts
```diff
@@ -1,2 +1,2 @@
 export const VERIFY_TIMEOUT = 5000;
-export const SAVE_TIMEOUT = 5000;
+export const SAVE_TIMEOUT = 10000;
```
tuul-wq, line 2: `10000` actually makes a difference?

src/renderer/entities/network/model/network-model.ts
```diff
@@ -323,7 +323,30 @@ sample({
 
 sample({
   clock: populateConnectionsFx.doneData,
-  fn: (connections) => keyBy(connections, 'chainId'),
+  source: $chains,
+  fn: (chains, connections) => {
+    const connectionsMap = keyBy(connections, 'chainId');
+
+    const newConnections = Object.entries(chains).map(([chainId, chain]) => {
```
tuul-wq, line 330: We can iterate only once and not spend resources on tuples
```javascript
    return Object.entries(chains).reduce<Record<ChainId, Connection>>((acc, [chainId, chain]) => {
      acc[chain.chainId] = connectionsMap[chainId] || {
          chainId: chain.chainId,
          canUseLightClient: false,
          connectionType: ConnectionType.AUTO_BALANCE,
          customNodes: [],
        };

      return acc;
    }, {});
```

src/renderer/features/balances/model/balance-subscription-model.ts
```diff
@@ -21,52 +24,73 @@ type SubscribeParams = {
...
-
-        return;
-      }
-
-      if (oldSubscription) {
-        const sameAccounts = oldSubscription?.accounts.join(',') === accountIds.join(',');
-
-        if (sameAccounts) {
-          return;
+    await Promise.all(
```
tuul-wq, line 39: Reduce nesting please

src/renderer/features/balances/model/balance-subscription-model.ts
```diff
@@ -21,52 +24,73 @@ type SubscribeParams = {
...
+            await unsubscribeBalancesFx({
+              chainId: chainId as ChainId,
+              subscription: oldSubscription,
+            });
+          }
+
+          const chain = chains[chainId as ChainId];
+
+          const balanceSubs = balanceSubscriptionService.subscribeBalances(chain, api, accountIds, (balances) => {
+            balances.forEach((balance) => balanceModel.events.balanceUpdated(balance));
```
tuul-wq, line 78: Bind event?

src/renderer/features/balances/model/balance-subscription-model.ts
tuul-wq: Overall this effect looks really fragile
It creates subscription which is fine but all those unsubs 
Please add `TODO` with a [task](https://app.clickup.com/t/86939u5ut) to revisit this effect in future

src/renderer/features/balances/model/balance-subscription-model.ts
```diff
@@ -81,28 +105,39 @@ const unsubscribeBalancesFx = createEffect(async ({ subscription }: UnsubscribeP
     return;
   }
 
-  const [balanceUnsubs, lockUnsubs] = await subscription.subscription;
+  const [balanceUnsubs, lockUnsubs] = await Promise.all(subscription.subscription);
```
tuul-wq, line 108: `subscription.subscription` cannot contain regular functions? Otherwise `Promise.all` won't invoke them

src/renderer/features/balances/model/balance-subscription-model.ts
```diff
@@ -81,28 +105,39 @@ const unsubscribeBalancesFx = createEffect(async ({ subscription }: UnsubscribeP
...
   balanceUnsubs.forEach((fn) => fn());
   lockUnsubs.forEach((fn) => fn());
 });
 
+const populateBalancesFx = createEffect((accounts: AccountId[]): Promise<Balance[]> => {
+  return balanceService.getBalances(accounts);
+});
+
+sample({
+  clock: walletModel.$activeAccounts,
```
tuul-wq, line 119: With this change we store balances only for active accounts, right?
What if we select multisig wallet? We also need to retrieve signatories' balances

src/renderer/features/balances/model/balance-subscription-model.ts
```diff
@@ -81,28 +105,39 @@ const unsubscribeBalancesFx = createEffect(async ({ subscription }: UnsubscribeP
...
+sample({
+  clock: populateBalancesFx.doneData,
+  target: balanceModel.$balances,
+});
+
 sample({
-  clock: [networkModel.$connectionStatuses, walletModel.$activeAccounts],
+  clock: throttle({
+    source: combine([networkModel.$connectionStatuses, walletModel.$activeAccounts]),
+    timeout: 500,
```
tuul-wq, line 132: Good to introduce a constant for `500`
What about other values 250, 750, 1000, 1500?

tuul-wq (changes requested):
Nice update!

src/renderer/entities/balance/lib/common/constants.ts
Asmadek: No big difference

src/renderer/features/balances/model/balance-subscription-model.ts
Asmadek: tried 1000, no big difference with 500

PR #1305, opened 2023-12-11
Feat: rename wallet

egor0798:
## Changes:
- Added `GeneralWalletActions` for forget and rename options
- Added `rename-wallet-model` and tests for that model
- Updated `wallet-model` and `wallet-select-model`

***

image
image
image


src/renderer/widgets/WalletDetails/model/__tests__/rename-wallet-model.test.ts
```diff
@@ -0,0 +1,44 @@
...
+    });
+
+    await allSettled(renameWalletModel.events.formInitiated, { scope, params: walletMock.wallet1 });
+    await allSettled(renameWalletModel.$walletForm.setForm, { scope, params: { name: walletMock.wallet2.name } });
+    await allSettled(renameWalletModel.$walletForm.validate, { scope });
+
+    expect(scope.getState(renameWalletModel.$walletForm.$isValid)).toEqual(false);
+  });
+
+  test('should updated $walletForDetails and $wallets after name change', async () => {
```
tuul-wq, line 30: Tests must cover only `wallet-rename-model` data flow
I'll ask in chat how to test data flow after form submission

src/renderer/widgets/WalletDetails/model/rename-wallet-model.ts
```diff
@@ -0,0 +1,101 @@
...
+});
+
+type SourceParams = {
+  walletToEdit: Wallet;
+  wallets: Wallet[];
+};
+function validateNameExist(value: string, _: unknown, params: SourceParams): boolean {
+  if (!value) return true;
+
+  const isSameName = value.toLowerCase() === params.walletToEdit.name.toLowerCase();
```
tuul-wq, line 63: Are you sure we cannot save the same name even if it's not equal some other wallet name?

src/renderer/widgets/WalletDetails/model/rename-wallet-model.ts
```diff
@@ -0,0 +1,101 @@
...
+          source: combine({
+            walletToEdit: $walletToEdit,
+            wallets: walletModel.$wallets,
+          }),
+          validator: validateNameExist,
+        },
+      ],
+    },
+  },
+  validateOn: ['change', 'submit'],
```
tuul-wq, line 40: `submit` is enough 😉
```suggestion
  validateOn: ['submit'],
```

src/renderer/widgets/WalletDetails/ui/WalletActions/actions/RenameWalletModal.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  useEffect(() => {
+    renameWalletModel.events.formInitiated(wallet);
+  }, [wallet]);
+
+  useEffect(() => {
+    renameWalletModel.events.callbacksChanged({ onSubmit: onClose });
+  }, [onClose]);
+
+  const submitForm = (event: FormEvent) => {
```
tuul-wq, line 35: 👍

src/renderer/widgets/WalletDetails/ui/WalletActions/actions/RenameWalletModal.tsx
```diff
@@ -0,0 +1,64 @@
+import { useForm } from 'effector-forms';
+import { useUnit } from 'effector-react';
+import { FormEvent, useEffect } from 'react';
+
+import { Wallet } from '@shared/core';
+import { BaseModal, Button, Input, InputHint } from '@shared/ui';
+import { useI18n } from '@app/providers';
+import { renameWalletModel } from '@widgets/WalletDetails/model/rename-wallet-model';
```
tuul-wq, line 8: Can be shortened

src/renderer/widgets/WalletDetails/ui/WalletActions/GeneralWalletActions.tsx
```diff
@@ -0,0 +1,48 @@
...
+        <DropdownIconButton.Items>
+          {options.map((option) => (
+            <DropdownIconButton.Item key={option.icon}>
+              <DropdownIconButton.Option option={option} />
+            </DropdownIconButton.Item>
+          ))}
+        </DropdownIconButton.Items>
+      </DropdownIconButton>
+      {/*<DropdownIconButton className="m-1.5" name="more" options={options} optionsClassName="right-0" />*/}
+      <RenameWalletModal wallet={wallet} isOpen={isRenameModalOpen} onClose={toggleIsRenameModalOpen} />
```
tuul-wq, line 45: I don't like this approach of unification
`Rename` & `Forget` modals will stay here but `extraActions` will always stay outside of this component, so I should look here and there to understand what actions I'm able to do
If you put all available actions in dedicated `wallet details modal` then you will see everything at once and have the whole picture 
The ideal solution to remove duplication is to use modal manager 😊 which requires replacement for headlessui modal

tuul-wq (changes requested):
Good, left comments

src/renderer/widgets/WalletDetails/model/rename-wallet-model.ts
egor0798: we can, that's why this condition is here. `return isSameName || isUnique` - name is valid if it's same or unique

src/renderer/widgets/WalletDetails/model/__tests__/rename-wallet-model.test.ts
```diff
@@ -0,0 +1,44 @@
+import { fork, allSettled } from 'effector';
+
+import { walletMock } from './mocks/wallet-mock';
+import { renameWalletModel } from '../rename-wallet-model';
+import { walletModel } from '@entities/wallet';
+import { walletSelectModel } from '@features/wallets';
+
+jest.mock('@app/providers', () => ({
```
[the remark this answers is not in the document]
egor0798: no, deleted it

src/renderer/widgets/WalletDetails/ui/WalletConnectDetails.tsx
```diff
@@ -14,9 +14,18 @@ import { wcDetailsModel } from '../model/wc-details-model';
...
+  Button,
+  ConfirmModal,
+  DropdownIconButton,
+  FootnoteText,
+  Icon,
+  SmallTitleText,
+  StatusModal,
+} from '@shared/ui';
 import { IconNames } from '@shared/ui/Icon/data';
+import { RenameWalletModal } from './WalletActions/RenameWalletModal';
```
tuul-wq, line 28: Rename / Forget and other actions are features
Please move `RenameWallet` to `features/wallet`

src/renderer/entities/wallet/model/wallet-model.ts
```diff
@@ -43,6 +43,7 @@ const walletConnectCreated = createEvent<CreateParams<WalletConnectAccount>>();
 const walletSelected = createEvent<ID>();
 const multisigAccountUpdated = createEvent<MultisigUpdateParams>();
 const walletRemoved = createEvent<ID>();
+const walletRenamed = createEvent<Wallet>();
```
tuul-wq, line 46: I was thinking about this for a while and came to conclusion
_Wallet rename_ is feature that must start and end in it's own place (feature)
Otherwise we start rename inside a feature but the real action ends in ... `entity/wallet-model` that's wrong
We have kind of same approach for wallet creation, but yesterday I've moved `vaultCreated` into `ManageVault` model, it's the right place where process starts and ends
Overall each wallet creation must start and end in appropriate onboarding flow (work to be done)

So the change to be made is to move `walletRenamed` inside `RenameWallet` feature and update `walletModel.$wallet` in the end

tuul-wq (changes requested):

tuul-wq (approved):

PR #1330, opened 2023-12-22
Feat: token component

egor0798:
## Changes:
- Added `Token` component
- Updated `Accordion` component

***

image


src/renderer/shared/ui/Accordion/Accordion.tsx
```diff
@@ -31,7 +31,10 @@ const Button = forwardRef<HTMLButtonElement, PropsWithChildren<ButtonProps>>(
     return (
       <Disclosure.Button
         ref={ref}
-        className={cnTw('group flex items-center justify-between w-full gap-x-2', buttonClass)}
+        className={cnTw(
+          'group flex items-center justify-between w-full gap-x-2 hover:bg-action-background-hover',
```
Asmadek, line 35: Looks like it needs border radius here

src/renderer/shared/ui/Token/Token.tsx
```diff
@@ -0,0 +1,50 @@
...
+  description?: ReactNode;
+  buttonClass?: string;
+  iconWrapper?: string;
+  iconOpened?: IconNames;
+  iconClosed?: IconNames;
+  onClick?: () => void;
+};
+
+const Button = forwardRef<HTMLButtonElement, PropsWithChildren<ButtonProps>>(
+  ({ buttonClass, children, description, ...props }, ref) => {
```
Asmadek, line 26: is `...props` needed here? You don't use it

src/renderer/shared/ui/Token/Token.tsx
```diff
@@ -0,0 +1,50 @@
...
+    );
+  },
+);
+
+type ContentProps = {
+  as?: ElementType;
+  className?: string;
+};
+
+const Content = ({ ...props }: PropsWithChildren<ContentProps>) => {
```
Asmadek, line 43: Looks like same as `(props: PropsWithChildren<ContentProps>)`

Asmadek (approved):
Few comments

PR #1334, opened 2023-12-26
fix: missed data from matrix event

Asmadek:
* use accounts ref for events handlers


Asmadek:
Task linked: [CU-86936824v Multisig operations. Missed data from matrix event.](https://app.clickup.com/t/86936824v)

tuul-wq (approved):

PR #1335, opened 2023-12-26
Feat: updated accordion styles

egor0798:
## Changes:
- Updated `Accordion` icon and hover styles
- Updated `Accrodion` usages

***

image


src/renderer/shared/assets/images/arrows/down.svg
tuul-wq, line 1: As I seen, we already have `chevron up & down`
If there's no need in `arrow up & down` better to delete them 😉

tuul-wq (approved):
Approve with 1 comment

Asmadek (approved):

PR #1337, opened 2023-12-26
Fix: active wallet select after forgetting last wallet

egor0798:
Check for updating previously selected wallet in db prevented new wallet from being selected.


src/renderer/entities/wallet/model/wallet-model.ts
```diff
@@ -8,6 +8,7 @@ import { modelUtils } from '../lib/model-utils';
 import { accountUtils } from '../lib/account-utils';
 
 const $wallets = createStore<Wallet[]>([]);
+console.log($wallets.getState());
```
tuul-wq, line 11: Remove
If you'd like to watch store value you should use `$store.watch((val) => ...);`

Asmadek (approved):

PR #1338, opened 2023-12-26
Fix: Block explorer

tuul-wq:
- Fixed Block explorer context menu appearing beneath root account


src/renderer/entities/wallet/ui/MultishardAccountsList/MultishardAccountsList.tsx
```diff
@@ -29,7 +29,7 @@ export const MultishardAccountsList = ({ accounts, chains, className }: Props) =
             explorers={RootExplorers}
             button={
               <ContactItem
-                className="sticky top-0 bg-white z-10 py-4"
```
tuul-wq, line 32: Sticky was not working anyway
We need to add `className` inside `ContextMenu` if we want to do that stuff
Right now it's not important

Asmadek (approved):

egor0798 (approved):

PR #1340, opened 2023-12-26
fix: reject reason for rejector

Asmadek:
* update transaction after reject/approve, not only when disconnected from matrix


Asmadek:
Task linked: [CU-8693d7v6a Matrix events. Reject transaction. Won't set for initiator application.](https://app.clickup.com/t/8693d7v6a)

tuul-wq (approved):

egor0798 (approved):

PR #1346, opened 2023-12-28
fix: network update

Asmadek:
* use create instead of update when we try to update default connection state

### DIFF
--- a/src/renderer/entities/network/lib/common/utils.ts
+++ b/src/renderer/entities/network/lib/common/utils.ts
@@ -39,5 +39,5 @@ export const isDisabled = (connection: Connection): boolean => {
 };
 
 export const isEnabled = (connection: Connection): boolean => {
-  return !isDisabled(connection);
+  return connection.connectionType !== ConnectionType.DISABLED;
 };
--- a/src/renderer/entities/network/lib/common/utils.ts
+++ b/src/renderer/entities/network/lib/common/utils.ts
@@ -37,3 +37,7 @@ export const getParachains = (chainMap: ChainMap, chainId: ChainId): Chain[] => 
 export const isDisabled = (connection: Connection): boolean => {
   return connection.connectionType === ConnectionType.DISABLED;
 };
+
+export const isEnabled = (connection: Connection): boolean => {
+  return !isDisabled(connection);
+};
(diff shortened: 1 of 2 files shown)


Asmadek:
Task linked: [CU-8693dgjpu Network doesn't save state after disable](https://app.clickup.com/t/8693dgjpu)

egor0798 (approved):

src/renderer/entities/network/lib/common/utils.ts
```diff
@@ -37,3 +37,7 @@ export const getParachains = (chainMap: ChainMap, chainId: ChainId): Chain[] =>
 export const isDisabled = (connection: Connection): boolean => {
   return connection.connectionType === ConnectionType.DISABLED;
 };
+
+export const isEnabled = (connection: Connection): boolean => {
+  return !isDisabled(connection);
```
tuul-wq, line 42: This is ok, but I'm a bit afraid of this dependency
What if one day `isDisabled` will rely on something different than `ConnectionType.DISABLED`?
At the same time `isEnabled` still should rely on `ConnectionType.DISABLED` - it's gonna break.
The bulletproof solution is to check `connection.connectionType !== ConnectionType.DISABLED` 😉

@valentunn Want to hear your opinion, if you don't mind

Asmadek: I think can be only two states, enabled and disabled, and one is just revers of another, checking in both may lead to errors or double code rewriting

valentunn: I would split such helper `bool` functions on enum into two categories:

1. A simple check for a certain value:
Example: `isDisabled` function 
2.  Some extra property we derive from the enum, for example `shouldTryConnecting(connectionState)`

Regardless of the category, I usually use conditions that won't break the overall logic of the helper function if I forget to consider this function OR construct a function in a such a way that adding new enum will result in compile error if forgot to modify the function

Specifically, 
for the 1 kind of functions I usually do it in the similar way it is already implemented
for the 2 kind I usually use exhaustive when so that I am forced to consider new enum value for this function due to compile error otherwise

In our case looks like isEnabled is of the first type, so it just checks whether the value of enum matches or not. So I would actually suggest `state == ENABLED` because `!is disabled` and `connection.connectionType !== ConnectionType.DISABLED` will result in true if we add an additional enum state which might break our expectations of what `isEnabled` is supposed to do

valentunn: I agree with Yaroslav, since we actually have 4 connection states it is better to handle them explicitly in such a helper functions so they won't depend on each other

tuul-wq (approved):

PR #1353, opened 2024-01-09
fix: bifrost xcm types 

Asmadek:
* get types for xcm versioned from call arguments

### DIFF
--- a/src/renderer/shared/lib/utils/substrate.ts
+++ b/src/renderer/shared/lib/utils/substrate.ts
@@ -4,9 +4,8 @@ import { isHex, hexToU8a, bnMin, BN_TWO, BN } from '@polkadot/util';
 import { blake2AsHex } from '@polkadot/util-crypto';
 
 import { Address, CallData, CallHash, XcmPallets } from '@shared/core';
-// TODO: Incorrect import
-import { DEFAULT_TIME, ONE_DAY, THRESHOLD } from '@entities/network';
 import { XcmTransferType } from '../../api/xcm';
+import { DEFAULT_TIME, ONE_DAY, THRESHOLD } from './constants';
 
 const V3_LABEL = 'V3';
 const UNUSED_LABEL = 'unused';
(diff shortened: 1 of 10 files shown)


Asmadek:
Task linked: [CU-8693fvkg5 Add new XCM type - StagingXcmVersionedMultiAsset](https://app.clickup.com/t/8693fvkg5)

src/renderer/shared/lib/utils/substrate.ts
```diff
@@ -3,8 +3,10 @@ import { BaseTxInfo, getRegistry, GetRegistryOpts, OptionsWithMeta, TypeRegistry
 import { isHex, hexToU8a, bnMin, BN_TWO, BN } from '@polkadot/util';
 import { blake2AsHex } from '@polkadot/util-crypto';
 
-import { Address, CallData, CallHash } from '@shared/core';
+import { Address, CallData, CallHash, XcmPallets } from '@shared/core';
+// TODO: Incorrect import
```
tuul-wq, line 7: All of these constants are used only in `shared/utils/substrate` we can safely move to `shared/utils/constants`
```javascript
export const THRESHOLD = BN_THOUSAND.div(BN_TWO);
export const DEFAULT_TIME = new BN(6_000);
export const ONE_DAY = new BN(24 * 60 * 60 * 1000);
```

src/renderer/shared/lib/utils/substrate.ts
```diff
@@ -128,3 +130,35 @@ export const getTypeVersions = (api: ApiPromise, typeName: string): string[] =>
...
+  api: ApiPromise,
+  transferType: XcmTransferType,
+): { pallet: XcmPallets; call: string } => {
+  if (transferType === 'xtokens') {
+    return { pallet: 'xTokens', call: 'transferMultiasset' };
+  }
+
+  const pallet = api.tx.xcmPallet ? 'xcmPallet' : 'polkadotXcm';
+
+  if (transferType === 'xcmpallet') {
```
tuul-wq, line 154: `const enum XcmTransferType` would be a nice enhancement, what do you think?
To remove string conditions

tuul-wq (commented):
Few comments

tuul-wq (approved):

stepanLav (approved):

PR #1371, opened 2024-01-16
fix: not create proxied account with non-zero delay

Asmadek:
* add check on delay for proxied account creation

### DIFF
--- a/src/renderer/features/proxies/lib/worker-utils.ts
+++ b/src/renderer/features/proxies/lib/worker-utils.ts
@@ -20,6 +20,7 @@ export const proxyWorkerUtils = {
   isSameProxied,
   isProxiedAccount,
   isApiConnected,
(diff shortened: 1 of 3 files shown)


egor0798 (approved):

src/renderer/features/proxies/lib/worker-utils.ts
```diff
@@ -61,6 +62,10 @@ function isApiConnected(apis: Record<ChainId, ApiPromise>, chainId: ChainId): bo
   return Boolean(api?.isConnected);
 }
 
+function isAvailableProxiedAccount(proxy: NoID<ProxyAccount>): boolean {
```
tuul-wq, line 65: Function compares 2 values, no proxieds, no accounts 🤔 
`isDelayedProxy` seems to be the most obvious name

PR #1373, opened 2024-01-17
fix: avoid create proxieds for watch only

Asmadek:
* use two arrays for accounts to add proxy and proxied accounts

### DIFF
--- a/src/renderer/entities/wallet/model/wallet-model.ts
+++ b/src/renderer/entities/wallet/model/wallet-model.ts
@@ -1,6 +1,5 @@
 import { combine, createEffect, createEvent, createStore, sample } from 'effector';
 import { spread } from 'patronum';
-import keyBy from 'lodash/keyBy';
 
 import type {
   Account,
@@ -11,7 +10,6 @@ import type {
   NoID,
   ProxiedAccount,
   Wallet,
-  WalletsMap,
 } from '@shared/core';
 import { WalletConnectAccount } from '@shared/core';
 import { storageService } from '@shared/api/storage';
@@ -37,15 +35,6 @@ const $activeWallet = combine(
     return wallets.find((wallet) => wallet.isActive);
   },
   { skipVoid: false },
-);
-const $walletsMap = combine(
-  $wallets,
-  (wallets): WalletsMap | undefined => {
-    return keyBy(wallets, 'id');
-  },
-  {
-    skipVoid: false,
-  },
 );
 
 const $accounts = createStore<Account[]>([]);
@@ -254,7 +243,6 @@ sample({
 export const walletModel = {
   $wallets,
   $activeWallet,
-  $walletsMap,
   $accounts,
   $activeAccounts,
   $isLoadingWallets: fetchAllWalletsFx.pending,
--- a/src/renderer/entities/wallet/model/wallet-model.ts
+++ b/src/renderer/entities/wallet/model/wallet-model.ts
@@ -38,9 +38,15 @@ const $activeWallet = combine(
   },
   { skipVoid: false },
 );
-const $walletsMap = combine($wallets, (wallets): WalletsMap | undefined => keyBy(wallets, 'id'), {
-  skipVoid: false,
-});
+const $walletsMap = combine(
+  $wallets,
+  (wallets): WalletsMap | undefined => {
+    return keyBy(wallets, 'id');
+  },
+  {
+    skipVoid: false,
+  },
+);
 
 const $accounts = createStore<Account[]>([]);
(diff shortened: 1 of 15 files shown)


Asmadek:
Task linked: [CU-8693j0p5y Proxy. Proxied account should not be added for WOW](https://app.clickup.com/t/8693j0p5y)

src/renderer/features/proxies/lib/__tests__/proxies-utils.test.ts
```diff
@@ -22,4 +22,37 @@ describe('features/proxies/lib/proxies-utils', () => {
 
     expect(result).toEqual(false);
   });
+
+  test('should check if proxy for account is available (true for chain account)', () => {
```
tuul-wq, line 26: I don't like that first half of test name is kind of meaningless and only second part is actually states what test does
Try this one: "should be true if proxy is available for chain account" and so on 😉

src/renderer/entities/wallet/model/wallet-model.ts
```diff
@@ -36,6 +38,9 @@ const $activeWallet = combine(
   },
   { skipVoid: false },
 );
+const $walletsMap = combine($wallets, (wallets): WalletsMap | undefined => keyBy(wallets, 'id'), {
```
tuul-wq, line 41: Let's not abuse oneliners for combine
```javascript
const $walletsMap = combine(
  $wallets,
  (wallets): WalletsMap | undefined => {
    return keyBy(wallets, 'id');
  },
  { skipVoid: false },
);
```

src/renderer/features/proxies/lib/__tests__/worker-utils.test.ts
```diff
@@ -157,4 +157,28 @@ describe('features/proxies/lib/worker-utils', () => {
 
     expect(result).toBeUndefined();
   });
+
+  test('should check proxy is delayed (true for delay > 0)', () => {
```
tuul-wq, line 161: Same here
Is it AI generated test names?

src/renderer/features/proxies/model/proxies-model.ts
```diff
@@ -75,14 +77,26 @@ type GetProxiesResult = {
...
     );
 
-    return endpoint.call.getProxies(chainId, nonProxiedAccounts, proxiedAccounts, proxies) as Promise<GetProxiesResult>;
+    return endpoint.call.getProxies(
+      chainId,
+      accountsForProxy,
+      accountsForProxied,
+      proxiedAccounts,
+      proxies,
+      wallets,
```
tuul-wq, line 98: `wallets,` is not used inside `getProxies`

src/renderer/features/proxies/workers/proxy-worker.ts
```diff
@@ -97,7 +97,8 @@ async function disconnect(chainId: ChainId) {
 // TODO: Refactor this code
 async function getProxies(
   chainId: ChainId,
```
tuul-wq, line 99: Suggest to use object when pass 5 parameters

src/renderer/features/proxies/model/proxies-model.ts
```diff
@@ -75,14 +77,26 @@ type GetProxiesResult = {
...
     const proxiedAccounts = accounts.filter((a) => accountUtils.isProxiedAccount(a));
-    const nonProxiedAccounts = keyBy(
-      accounts.filter((a) => !accountUtils.isProxiedAccount(a)),
+
+    const accountsForProxy = keyBy(
+      accounts.filter((a) => proxiesUtils.isProxyAvailable(a)),
+      'accountId',
+    );
+    const accountsForProxied = keyBy(
+      accounts.filter((a) => proxiesUtils.isProxiedAvailable(a, wallets[a.walletId])),
```
tuul-wq, line 88: Not sure that we require `walletsMap` because `getProxiesFx` can conveniently use it
It's ok to pass wallets array, build map from it and use just right here

tuul-wq (changes requested):
Bunch of suggestions

src/renderer/features/proxies/lib/__tests__/proxies-utils.test.ts
Asmadek: I will rename it, but I think we should discuss naming convention for tests

src/renderer/features/proxies/model/proxies-model.ts
Asmadek: done, but we already calculate it in several places, and it may be useful to have it just in one place

tuul-wq (approved):

PR #1385, opened 2024-01-24
fix:  proxy creation through db

Asmadek:
* proxy creation through db
* collect proxies for newly created proxied account

### DIFF
--- a/src/renderer/features/proxies/workers/__tests__/proxy-worker.test.ts
+++ b/src/renderer/features/proxies/workers/__tests__/proxy-worker.test.ts
@@ -377,7 +377,15 @@ describe('features/proxies/workers/proxy-worker', () => {
       proxies,
     });
 
-    expect(result.proxiesToAdd).toEqual([]);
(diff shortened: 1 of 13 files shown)


Asmadek:
Task linked: [CU-8693m099g Proxy. Delegated authorities. Proxies are shown only in session where it was added](https://app.clickup.com/t/8693m099g)

src/renderer/features/proxies/workers/__tests__/proxy-worker.test.ts
```diff
@@ -377,7 +377,15 @@ describe('features/proxies/workers/proxy-worker', () => {
       proxies,
     });
 
-    expect(result.proxiesToAdd).toEqual([]);
+    expect(result.proxiesToAdd).toEqual([
```
tuul-wq, line 380: Maybe extract it into a variable?

tuul-wq (approved):

egor0798 (approved):

PR #1387, opened 2024-01-25
fix: proxies in details

Asmadek:
* fix: proxies in details
* fix: add multisig when proxied already exists


src/renderer/widgets/CreateWallet/ui/MultisigWallet/components/WalletForm.tsx
```diff
@@ -80,7 +80,9 @@ export const WalletForm = ({ signatories, onContinue, isActive, isLoading, onGoB
     );
   });
 
-  const accountAlreadyExists = accounts.some((a) => a.accountId === multisigAccountId);
```
tuul-wq, line 83: Can't we do `.some(a => !accountUtils.isProxiedAccount(a) && a.accountId === multisigAccountId)`?

tuul-wq (approved):

PR #1389, opened 2024-01-25
Fix: Multisig creation warning

### LINKED ISSUE #1384: If user has multisig as WOW, then he can't create new one.
If user has multisig as WOW, then he can't create new one.

**Case:**
Add multisig wallet as a WOW
Try to create that multisig wallet

**Actual result:**
Can't create multisig wallet, trigger error that wallet already exiist

**Expected result:**
That validation should be triggered only if it's a multisig wallet
User can't create multisig wallet only if another multisig with the same address exist.

**Additional info:**
Error should show which wallet user already has in the app

Tbaut:
closes #1384 

A first shot at a simple problem. It's not ideal and I'm looking for comments.

The issue says:
>  Error should show which wallet user already has in the app

But it's clear what account already exist isn't it? It's the one that we create.

>  User can't create multisig wallet only if another multisig with the same address exist.

The code only checks whether another account exist, be it a multisig or not. The face that we are creating a multisig is so that if the account exists, then it must be a multisig.

See comments with some questions


src/renderer/widgets/CreateWallet/ui/MultisigWallet/components/WalletForm.tsx
```diff
@@ -69,8 +69,7 @@ export const WalletForm = ({ signatories, onContinue, isActive, isLoading, onGoB
     onSubmit(name, threshold.value, creator.accountId);
   };
 
-  const hasNoAccounts =
-    wallets.filter((wallet) => !walletUtils.isWatchOnly(wallet) || !walletUtils.isMultisig(wallet)).length === 0;
+  const hasNoAccount = !wallets.some((wallet) => !walletUtils.isWatchOnly(wallet) || !walletUtils.isMultisig(wallet));
```
Tbaut, line 72: I took the liberty to change a check for `filter().length === 0` by a `some()` because it's more efficient as it will stop iterating as soon as it's false.

src/renderer/widgets/CreateWallet/ui/MultisigWallet/components/WalletForm.tsx
```diff
@@ -80,7 +79,14 @@ export const WalletForm = ({ signatories, onContinue, isActive, isLoading, onGoB
     );
   });
 
-  const accountAlreadyExists = accounts.some((a) => a.accountId === multisigAccountId);
+  const accountAlreadyExists = wallets.some((wallet) => {
+    return (
+      !walletUtils.isWatchOnly(wallet) &&
```
Tbaut, line 84: This approach seems quite overkill. Am I right that there is no easy way or util currently to go from an account to its wallet, beside filtering by `account.walletId`?

I would ideally check for an `account === multisigAccountId` and then only check whether this account is part of a wallet that is `watchOnly`. But I couldn't find an easy way to do this.

src/renderer/widgets/CreateWallet/ui/MultisigWallet/components/WalletForm.tsx
```diff
@@ -80,7 +79,14 @@ export const WalletForm = ({ signatories, onContinue, isActive, isLoading, onGoB
...
+      accounts.some((account) => {
+        return account.accountId === multisigAccountId && account.walletId === (wallet as Wallet).id;
```
Tbaut, line 86: this `wallet as Wallet` is very weird to me. For some reason, I got a TS error here saying that "Property 'id' does not exist on type 'never'". I really don't understand. Let me know if you have an idea, otherwise I'll research more to remove this type casting.

src/renderer/entities/wallet/model/wallet-model.ts
```diff
@@ -117,6 +117,13 @@ const multishardCreatedFx = createEffect(
   },
 );
 
+// TODO: Move wallet creation to its own feature
+const proxiedWalletsCreatedFx = createEffect(
```
tuul-wq, line 121: Serves some purpose? or just a typo?

Tbaut: Oops conflict resolution gone wrong, I'll remove.

src/renderer/widgets/CreateWallet/ui/MultisigWallet/components/WalletForm.tsx
Tbaut: even better, thank you

Tbaut:
Thanks a lot for the help and answers. It should be better now :)

tuul-wq (approved):

Asmadek (approved):

PR #1393, opened 2024-01-29
fix: show assets for available chains

Asmadek:
* check is chain id match for active accounts on assets page

### DIFF
--- a/src/renderer/entities/wallet/lib/account-utils.ts
+++ b/src/renderer/entities/wallet/lib/account-utils.ts
@@ -20,7 +20,7 @@ export const accountUtils = {
   isBaseAccount,
   isChainAccount,
   isMultisigAccount,
-  isAllChainsSupport,
+  isChainDependant,
   isChainIdMatch,
   isWalletConnectAccount,
   isProxiedAccount,
@@ -58,12 +58,12 @@ function isAccountWithShards(accounts: Pick<Account, 'type'> | ShardAccount[]): 
   return Array.isArray(accounts) && isShardAccount(accounts[0]);
 }
 
-function isAllChainsSupport(account: Pick<Account, 'type'>): boolean {
-  return isBaseAccount(account) || isMultisigAccount(account);
+function isChainDependant(account: Pick<Account, 'type'>): boolean {
+  return !isBaseAccount(account) && !isMultisigAccount(account);
 }
 
 function isChainIdMatch(account: Pick<Account, 'type'>, chainId: ChainId): boolean {
-  if (isAllChainsSupport(account)) return true;
+  if (!isChainDependant(account)) return true;
 
   const chainAccountMatch = isChainAccount(account) && account.chainId === chainId;
   const shardAccountMatch = isShardAccount(account) && account.chainId === chainId;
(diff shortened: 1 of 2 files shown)


Asmadek:
Task linked: [CU-8693j5t6f Proxy. Assets page. Does not match with mockups](https://app.clickup.com/t/8693j5t6f)

src/renderer/pages/Assets/Assets/components/AssetsList/AssetsList.tsx
```diff
@@ -41,7 +42,9 @@ export const AssetsList = () => {
       const isDisabled = networkUtils.isDisabledConnection(connections[c.chainId]);
       const hasMultiPallet = !isMultisig || networkUtils.isMultisigSupported(c.options);
 
-      return !isDisabled && hasMultiPallet;
+      const hasChainAccount = activeAccounts.some((a) => accountUtils.isChainIdMatch(a, c.chainId));
```
tuul-wq, line 45: Suggest to create a map `{ [chainId]: true / false }` and check it inside `filteredChains` for each chain
For instance PV wallet with 300 accs would be looked up `chains x 300` times

tuul-wq (changes requested):

src/renderer/pages/Assets/Assets/components/AssetsList/AssetsList.tsx
```diff
@@ -38,11 +38,23 @@ export const AssetsList = () => {
...
+      ? new Set(Object.keys(chains) as ChainId[])
+      : activeAccounts.reduce<Set<ChainId>>((acc, account) => {
+          const chainId = (account as ChainAccount).chainId;
+
+          if (chainId) {
+            acc.add(chainId);
+          }
+
+          return acc;
+        }, new Set<ChainId>());
```
Tbaut, line 51: May I suggest to use a `filter` here for readability, something like
```suggestion
      : new Set(activeAccounts.filter((account) => !!account.chainId))
```

src/renderer/entities/wallet/lib/account-utils.ts
```diff
@@ -57,8 +58,12 @@ function isAccountWithShards(accounts: Pick<Account, 'type'> | ShardAccount[]):
   return Array.isArray(accounts) && isShardAccount(accounts[0]);
 }
 
+function isAllChainsSupport(account: Pick<Account, 'type'>): boolean {
```
tuul-wq, line 61: I would suggest account-centric name like `isChainDependant`

```javascript
function isChainDependant(account: Pick<Account, 'type'>): boolean {
  return !isBaseAccount(account) && !isMultisigAccount(account);
}
```

src/renderer/pages/Assets/Assets/components/AssetsList/AssetsList.tsx
tuul-wq: Thibaut's suggestion is nice and clean, `reduce` makes no calculations inside just filtering
```javascript
new Set(activeAccounts.filter((account) => Boolean(account.chainId)))
```

egor0798 (approved):

tuul-wq (approved):

PR #1394, opened 2024-01-29
fix: don't switch to assets after add proxied

Asmadek:
* check that user already have wallets and don't need to redirect to assets


src/renderer/app/App.tsx
```diff
@@ -27,14 +28,17 @@ export const App = () => {
   const wallets = useUnit(walletModel.$wallets);
   const isLoadingWallets = useUnit(walletModel.$isLoadingWallets);
 
+  const previousWallets = usePrevious(wallets);
```
tuul-wq, line 31: Can we store previous `wallets.length`?

egor0798 (approved):

tuul-wq (approved):

PR #1399, opened 2024-01-30
Fix: Notifications sort

tuul-wq:
Fixed notification sort inside groups - the most recent at top


tuul-wq:
Task linked: [CU-8693jcwf9 Notifications. Wrong sorting rules inside date group](https://app.clickup.com/t/8693jcwf9)

egor0798 (approved):

PR #1401, opened 2024-01-30
Fix: Use transferKeepAlive instead of transferAllowDeath

### LINKED ISSUE #1024: Use `transfer_keep_alive` as (default) option for transfers
## Description

Presently `transfer_allow_death` is the call made with the send button (at least on Polkadot for a TX I tried recently).
I suggest the _default_ call be made to `transfer_keep_alive` : 

https://github.com/paritytech/substrate/blob/0cdf7029017b70b7c83c21a4dc0aa1020e7914f6/frame/balances/src/lib.rs#L625-L632

## Motivation

Help people not footgun that don't know about or understand the existential deposit model.

## Alternatives Considered

Nice to have: a way for power users to select a default, and at tx construction time in the UX override this to what they want.

Tbaut:
closes #1024

### DIFF
--- a/src/renderer/app/index.tsx
+++ b/src/renderer/app/index.tsx
@@ -20,7 +20,7 @@ log.variables.env = process.env.NODE_ENV;
 log.transports.console.format = '{y}/{m}/{d} {h}:{i}:{s}.{ms} [{env}#{version}]-{processType} [{level}] > {text}';
 log.transports.console.useStyles = true;
 
-Object.assign(console, log.functions);
+// Object.assign(console, log.functions);
 log.errorHandler.startCatching({
   showDialog: false,
   onError({ createIssue, error, processType, versions }) {
--- a/src/renderer/app/index.tsx
+++ b/src/renderer/app/index.tsx
@@ -20,7 +20,7 @@ log.variables.env = process.env.NODE_ENV;
 log.transports.console.format = '{y}/{m}/{d} {h}:{i}:{s}.{ms} [{env}#{version}]-{processType} [{level}] > {text}';
 log.transports.console.useStyles = true;
 
-// Object.assign(console, log.functions);
+Object.assign(console, log.functions);
 log.errorHandler.startCatching({
   showDialog: false,
   onError({ createIssue, error, processType, versions }) {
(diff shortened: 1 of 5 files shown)


src/renderer/app/index.tsx
```diff
@@ -20,7 +20,6 @@ log.variables.env = process.env.NODE_ENV;
 log.transports.console.format = '{y}/{m}/{d} {h}:{i}:{s}.{ms} [{env}#{version}]-{processType} [{level}] > {text}';
 log.transports.console.useStyles = true;
 
-Object.assign(console, log.functions);
```
tuul-wq, line 23: Please revert this change
For dev purposes you can comment this line but don't forget to put it back

Tbaut: Oh sorry, ofc

tuul-wq (approved):

Asmadek (commented):
I think we should update validation for `TransferForm` and signing step to show not enough balance error ASAP before sending transaction.

Tbaut:
Agreed, that's super important. Do you want to see this happen in this PR? Otherwise I'd create an issue and tackle it in a separate one.

pgolovkin:
> Agreed, that's super important. Do you want to see this happen in this PR? Otherwise I'd create an issue and tackle it in a separate one.

It's better to make it in this PR.
Also if the `transfer_keepAlive` is being used there should be ED check when the user makes a transfer.

tuul-wq:
@Tbaut We're about to make a huge update to transfer page (during upcoming weeks), that involves refactoring to effector model (not final but better than current solution). Knowing that I would suggest you to postpone this issue for a while. Otherwise you'll make a change that might be overwritten.
Moving from `transferAllowDeath` to `transferKeepAlive` is a valuable feature and we're definitely going to implement it, just need time and space around transfer page to make the transition painless and maintainable.

cc: @pgolovkin

Tbaut:
Thanks for the context @tuul-wq Is this huge update already in progress?
If not yet, since this PR is relatively well scoped, and simple (and tackles https://github.com/novasamatech/nova-spektr/issues/1024 not more not less), I'd be in favor of merging, and add anything that is relevant, but not actually part of the issue, in a separated issue. In general, I think that having tiny and to the point PRs is preferable to big ones touching many areas of the code. I'm talking about this one, not the big update you mention, sometimes there's no choice :)

Adding warning regarding ED etc is certainly important, but isn't more important before or after this PR, so I'm happy to create a specific issue for it and I can tackle it after the big update.

pgolovkin (approved):

Asmadek (approved):

Tbaut:
Thanks y'all, I'll create an issue for what we described.

Issue #1397, opened 2024-01-29
Prompted to "sign with Polkadot Vault" although I only have WC and Nova wallets


Tbaut:
## Description
For a multisig Tx, I was prompted the following

![image](https://github.com/novasamatech/nova-spektr/assets/33178835/cb69e490-6737-4bca-add6-d1badc9e7adc)

## Steps to Reproduce

1. Make sure you have only WC/WOW/Nova wallets
2. Make a multisig tx
3. See the above modal

## Expected Behavior

To have "Sign" if there are different types of wallet, or that the right type is selected

## Screenshots

See above

Tbaut:
The problem comes from this line https://github.com/novasamatech/nova-spektr/blob/bbcecd9592ca473e3f379682cc088209958c9b87/src/renderer/pages/Operations/components/ActionSteps/Confirmation.tsx#L97 that defaults to `WalletType.SINGLE_PARITY_SIGNER` when we don't know what wallet type the user have.

There are a couple solutions to this:
1. we can check what wallet types are available, for the signatories, and default to the first type
2. introduce a new `WalletText` when the type is unknown, and instead of having the button say "Sign with XYZ", it'd just say "Sign" without an icon either. Clicking on it subsequently shows the signatory selection modal.

Is there an approach you think is the best, or do you see any other way to make this better?

PR #1428, opened 2024-02-07
fix: Allow unkown wallet type

Tbaut:
closes https://github.com/novasamatech/nova-spektr/issues/1397

I went for the second solution from my comment https://github.com/novasamatech/nova-spektr/issues/1397#issuecomment-1931928078 because I thought it would be cleaner.


pgolovkin:
@Tbaut for addressing the text in localization file `t(operation.sign.unknown)` should be used.

Asmadek (approved):

Tbaut:
Thanks @pgolovkin, I had a type in the key of the text :man_facepalming: It's fixed now :)

pgolovkin (approved):

Tbaut:
Merging as discussed with Pavel and it's really small :)

PR #1474, opened 2024-02-16
Feat: Rewrite Networks List with FSD + Effector

### LINKED ISSUE #1455: Rewrite Networks List with FSD + Effector
### Description
In order to increase maintainability and testability of Networks list, we must rewrite that part of code according to FSD with help of effector.

What needs to be done:
- [x] Networks list feature (or widget)
- [x] Support 2 lists: active / inactive networks
- [x] Persist existing filtration mechanism
- [x] Tests

### LINKED ISSUE #1544: Network list on Settings should be expanded by default
Network list on Settings should be expanded by default.

DOD:
1. When the user opens the Networks setting network lists (enabled and disabled) should be expanded by default.
2. When the user makes a search then then network lists should be expanded

Tbaut:
closes #1455 
closes https://github.com/novasamatech/nova-spektr/issues/1544


src/renderer/pages/Settings/Networks/components/NetworkList/NetworkList.tsx
```diff
@@ -1,42 +1,54 @@
...
+  const filterQuery = useUnit(filterModel.$filterQuery);
+
+  console.log('filterQuery', filterQuery);
+  console.log('isDefaultOpen', isDefaultOpen);
+  console.log('isListOpen', isListOpen);
 
   useEffect(() => {
-    if (query) {
-      setIsListOpen(Boolean(query));
+    if (filterQuery) {
```
Tbaut, line 24: While I've kept the exact same logic, this actually doesn't work and from my tests, it wasn't working before either, as in:
- deactivate "Polkadot"
- make sure the accordion is closed
- type "Polka" in the search
- the Accordion of disabled networks doesn't open


The reason seems to be that the `Disclosure` component from `headlessui` doesn't change the behavior when this props changes.

Tbaut:
Although it's a WIP, I'd love to get an early review of this PR as my last commit unexpectedly broke the rendering. The `pnpm start:renderer:dev` command doesn't display any terminal error, neither does `pnpm build` but visiting `https://localhost:3000/` displays an error that I don't get
> TypeError
_model_network_model__WEBPACK_IMPORTED_MODULE_2__ is undefined
Call Stack
 networkModel
  renderer-10314b5427f1f882d3a6.js:376234:65
 ./src/renderer/features/networks/NetworkFilter/model/network-filter.ts
  renderer-10314b5427f1f882d3a6.js:401732:5

Can't use `import { networkModel } from '@entities/network';`?
Should this be somewhere different? Any other hint?

Thank you in advance.

src/renderer/entities/network/model/network-model.ts
```diff
@@ -39,6 +40,53 @@ const $apis = createStore<Record<ChainId, ApiPromise>>({});
 const $connections = createStore<Record<ChainId, Connection>>({});
 const $connectionStatuses = createStore<Record<ChainId, ConnectionStatus>>({});
 
+// not sure where this should live
+// network-utils did not seem like the best place
```
Tbaut, line 44: @tuul-wq what you you think?

Tbaut:
While this PR is still a draft, and contains some `console.log`, I'd be very helpful if you could skim though it, and tell me if there's any major flaw/issue in the logic.

Tbaut:
While this PR is still a draft, and contains some console.log, I'd be very helpful if you could skim though it, and tell me if there's any major flaw/issue in the logic @tuul-wq.

src/renderer/entities/network/model/network-model.ts
```diff
@@ -39,6 +40,53 @@ const $apis = createStore<Record<ChainId, ApiPromise>>({});
...
+      connectionStatus: connectionStatuses[chain.chainId],
+    };
+
+    acc.push(extendedChain);
+
+    return acc;
+  }, []);
+};
+
+const $activeChainsSorted: Store<ExtendedChain[]> = combine(
```
tuul-wq, line 64: These stores should be located in a dedicated feature - `src/features/network/NetworksList`

src/renderer/features/networks/NetworkFilter/model/network-filter.ts
tuul-wq, line 1: Better to place all filtering logic inside `NetworksList` because it's mostly the part of that feature

src/renderer/features/networks/NetworkFilter/model/network-filter.ts
```diff
@@ -0,0 +1,38 @@
+import { combine, createEvent, createStore, sample } from 'effector';
+
+import { includes } from '@shared/lib/utils';
+import { networkModel } from '@entities/network';
+
+const componentMounted = createEvent();
+
+const $filterQuery = createStore<string>('');
+const queryChanged = createEvent<string>();
```
tuul-wq, line 9: Events first, stores second, effects third

src/renderer/features/networks/NetworkFilter/model/network-filter.ts
```diff
@@ -0,0 +1,38 @@
...
+const queryReset = createEvent();
+
+$filterQuery.on(queryChanged, (_, query) => query).reset(queryReset);
+
+sample({
+  clock: componentMounted,
+  target: queryReset,
+});
+
+const $networksFiltered = combine(
```
tuul-wq, line 19: I'd rather stick to this one
```suggestion
const $filteredNetworks = combine(
```

src/renderer/features/networks/NetworkFilter/model/network-filter.ts
```diff
@@ -0,0 +1,38 @@
...
+import { includes } from '@shared/lib/utils';
+import { networkModel } from '@entities/network';
+
+const componentMounted = createEvent();
+
+const $filterQuery = createStore<string>('');
+const queryChanged = createEvent<string>();
+const queryReset = createEvent();
+
+$filterQuery.on(queryChanged, (_, query) => query).reset(queryReset);
```
tuul-wq, line 12: We used to declare unit connections through `sample`

src/renderer/features/networks/NetworkFilter/model/network-filter.ts
```diff
@@ -0,0 +1,38 @@
...
+
+sample({
+  clock: componentMounted,
+  target: queryReset,
+});
```
tuul-wq, line 17: `reinit` is a built-in reset event
`formInitiated` is preferable name, in future this event will be called even before React mounts the component 😉  (see [reflect](https://github.com/effector/reflect))
```suggestion
sample({
  clock: formInitiated,
  target: $filterQuery.reinit,
});
```

src/renderer/pages/Settings/Networks/components/NetworkList/NetworkList.tsx
```diff
@@ -1,42 +1,54 @@
...
+      setIsListOpen(true);
     } else {
       setIsListOpen(isDefaultOpen);
     }
-  }, [query]);
+  }, [filterQuery]);
 
   if (networkList.length === 0) return null;
 
   const { success, connecting, error } = networkList.reduce(
```
tuul-wq, line 33: Should be moved inside model

src/renderer/entities/network/model/network-model.ts
tuul-wq: `features/network/NetworksList/lib/networks-list-utils.ts`

tuul-wq (commented):
Good start, now let's separate code to dedicated features and continue with tests

src/renderer/pages/Settings/Networks/components/NetworkList/NetworkList.tsx
tuul-wq: I'll revisit Accordion behaviour for `NetworksList` and let you know

src/renderer/pages/Settings/Networks/components/NetworkList/NetworkList.tsx
Tbaut: I moved it inside the `network-list-utils`, because it's linked to a specific network list (either active or inactive) and I think it's probably an even better place. Let me know if you think otherwise.

src/renderer/entities/network/model/network-model.ts
Tbaut: Makes sense, thanks.

src/renderer/features/networks/NetworkFilter/model/network-filter.ts
tuul-wq: Contact form was the initial place where effector was integrated, so it contains outdated code style 😄

src/renderer/features/networks/NetworkList/networks-list-utils.ts
tuul-wq, line 1: All `-utils.ts` files should be inside `/lib` according to FSD
To reduce amount of exports from `utils` we use aggregating variable
```javascript
export const networksListUtils = {
  getSomething,
};

function getSomething() { ... }
```

src/renderer/features/networks/NetworkList/networks-list-utils.ts
```diff
@@ -0,0 +1,39 @@
...
+      connectionStatus: connectionStatuses[chain.chainId],
+    };
+
+    acc.push(extendedChain);
+
+    return acc;
+  }, []);
+};
+
+export const getMetrics = (networkList: ExtendedChain[]) =>
```
tuul-wq, line 23: Return type

src/renderer/pages/Settings/Networks/components/NetworkList/NetworkList.tsx
tuul-wq: The behaviour should be like this:
- disabled section closed by default / active section opened be default
- if filterQuery exists open both sections
- if filterQuery cleared with _cross button_ then collapse disabled section once again
- if filterQuery cleared with `backspace/delete` then disabled section stays opened until user closes it

Tbaut: > if filterQuery exists open both sections

This is the problem that I mentioned in my first comment. This is not the case in the `dev` branch, and it's also not the case in this branch either. This could be seen as a bug. However, I looked into it and I don't think there's a non-hacky way to achieve this behavior because the `defaultOpen` prop, is not changing the behavior of the `Disclosure` component. It is a **default** state, and it is only commanded by the button click after the first render.

> if filterQuery cleared with cross button then collapse disabled section once again

There isn't any cross button on Firefox, I wonder if the behavior here is similar to `Ctrl-a` + `Delete`

tuul-wq:
@Tbaut pushed 1 commit that separates code into distinct features, please take a look.
We can have a call so I'll explain why it was divided this way. 
The only thing that's left is fix tests in `Networks.test.ts`

cc: @Asmadek  please review

src/renderer/features/matrix/MatrixLogin/ui/LoginForm.test.tsx
```diff
@@ -20,7 +20,7 @@ describe('pages/Settings/Matrix/LoginForm', () => {
     await act(async () => {
       render(
         <Provider value={fork()}>
-          <LoginForm />
+          <LoginForm redirectStep="step" />
```
Tbaut, line 23: It's unrelated to this PR right?

src/renderer/features/network/NetworksList/model/inactive-networks-model.ts
```diff
@@ -0,0 +1,31 @@
+import { createEvent, sample, createStore, combine } from 'effector';
```
Tbaut: I wonder if this in particular is what you mentioned in a conversation we had in the past regarding code duplication, that we shouldn't be afraid of.

To me, active and inactive networks are very much related, and I'd have never thought of duplicating this somewhat simple code.

src/renderer/features/network/NetworksList/ui/ActiveNetwork.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <div className="spektr-waiting">
+        <FootnoteText className="text-text-tertiary">{t('settings.networks.connectingStatusLabel')}</FootnoteText>
+      </div>
+    ),
+  },
+  [ConnectionStatus.DISCONNECTED]: {
+    variant: 'waiting',
+    title: (t: TFunction) => (
+      <div className="spektr-waiting">
+        <FootnoteText className="text-text-tertiary">{t('settings.networks.connectingStatusLabel')}</FootnoteText>
```
Tbaut: I'm surprised to see the `connectingStatusLabel` together with a `DISCONNECTED` state. Wouldn't this make users think something may still happen (we'll be connected eventually) while it's actually not the case?

src/renderer/features/network/NetworksList/ui/ActiveNetwork.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+const Status = {
+  [ConnectionStatus.CONNECTING]: {
+    variant: 'waiting',
+    title: (t: TFunction) => (
+      <div className="spektr-waiting">
+        <FootnoteText className="text-text-tertiary">{t('settings.networks.connectingStatusLabel')}</FootnoteText>
+      </div>
+    ),
+  },
```
Tbaut: To simplify things a bit, what do you think about using the `<Trans>`?
I just tested it successfully. This would allow to remove the `{ t } = useI18n()` and there's then no need to pass `t` to the title function. It'd look like this.

```suggestion
            title: (
      <div className="spektr-waiting">
        <FootnoteText className="text-text-tertiary">
          <Trans>settings.networks.connectingStatusLabel</Trans>
        </FootnoteText>
      </div>
    ),
```

Happy to do the change if you're ok with it.

src/renderer/pages/Settings/Networks/Networks.tsx
```diff
@@ -42,29 +49,10 @@ export const Networks = () => {
...
-
-  const [inactive, active] = partition(extendedChains, ({ connection }) =>
-    networkUtils.isDisabledConnection(connection),
-  );
-
   const confirmRemoveCustomNode = (name: string): Promise<boolean> => {
     return confirm({
       title: t('settings.networks.confirmModal.removeTitle'),
-      message: <Trans t={t} i18nKey="settings.networks.confirmModal.removeLabel" values={{ node: name }} />,
+      message: <Trans t={t} i18nKey="settings.networkss.confirmModal.removeLabel" values={{ node: name }} />,
```
Tbaut: There's a typo here and below. I'll fix it, commenting here to remember
```suggestion
      message: <Trans t={t} i18nKey="settings.networks.confirmModal.removeLabel" values={{ node: name }} />,
```

src/renderer/features/network/NetworksList/ui/EmptyNetworks.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+import { Icon, BodyText } from '@shared/ui';
+import { activeNetworksModel } from '../model/active-networks-model';
+import { inactiveNetworksModel } from '../model/inactive-networks-model';
+import { useI18n } from '@app/providers';
+
+export const EmptyNetworks = () => {
+  const { t } = useI18n();
+
+  const [activeNetworks, inactiveNetworks] = useUnit([
```
Tbaut: this is what makes me think these 2 models should actually be together, we do group them in our head, but they come from 2 different models. Anyways, it's not a constructive comment, I'm just trying to prove my point.

Tbaut (commented):
Thanks for the help @tuul-wq 

While I understand where this comes from, I have to mention that I do not think that it is worth it to separate active and inactive networks model in relation to the added complexity. 
Effector on its own is already hard enough, having to add things like `networks-overview-model.ts` really feels like one more barrier in front of new contributions.

However, I do agree with splitting the UI components for enabled/disabled, even though these are related and result in code duplication, I do see the up side here.

I left a couple questions and comments. I'll push the obvious changes such as the typo, and take a look at the tests.

src/renderer/features/network/NetworksList/model/active-networks-model.ts
```diff
@@ -0,0 +1,31 @@
...
+      .getExtendedChain(networks, connections, connectionStatuses)
+      .filter((o) => networkUtils.isEnabledConnection(o.connection));
+  },
+);
+
+sample({ clock: networksChanged, target: $networks });
+
+export const activeNetworksModel = {
+  $activeNetworks,
+  events: {
```
Asmadek: In some places we have empty line before `events`, I think we should have one way for it

src/renderer/features/network/NetworksList/ui/ActiveNetwork.tsx
Asmadek: Is it works correctly with changed location? Now we don't have it (but we have language switcher component), but I think we should test it too. And what about `i18next/no-literal-string` validation? If all this parts okay, I think it's good idea. But it should be another PR if you want to do this for hole project.

src/renderer/features/network/NetworksList/model/inactive-networks-model.ts
```diff
@@ -0,0 +1,31 @@
...
+      .getExtendedChain(networks, connections, connectionStatuses)
+      .filter((o) => networkUtils.isDisabledConnection(o.connection));
+  },
+);
+
+sample({ clock: networksChanged, target: $networks });
+
+export const inactiveNetworksModel = {
+  $inactiveNetworks,
+  events: {
```
Asmadek: same about blank line

src/renderer/features/network/NetworksList/lib/networks-list-utils.ts
```diff
@@ -0,0 +1,37 @@
...
+    return {
+      ...chain,
+      connection: connections[chain.chainId],
+      connectionStatus: connectionStatuses[chain.chainId],
+    };
+  }, []);
+}
+
+type Metrics = Record<'success' | 'connecting' | 'error', number>;
+function getMetrics(networkList: ExtendedChain[]): Metrics {
```
Asmadek: Are you sure that `Metrics` is a good name? Maybe add smth specific for it? `getStatusMetrics` or smth else

src/renderer/features/network/NetworksList/ui/NetworksList.tsx
```diff
@@ -1,44 +1,23 @@
...
-
-  useEffect(() => {
-    if (query) {
-      setIsListOpen(Boolean(query));
-    } else {
-      setIsListOpen(isDefaultOpen);
-    }
-  }, [query]);
-
+export const NetworksList = ({ title, isDefaultOpen, networkList, children }: Props) => {
```
Asmadek: I think we should have a convention for naming. Because I prefer `NetworkList` for this

Asmadek (approved):

src/renderer/features/network/NetworksList/model/active-networks-model.ts
Tbaut: Thanks, this one I can probably remember in the future, it has actually triggered me to see this empty line :sweat_smile:. It'd be great to have eslint rules for this.. but I know it's a rabbit hole.

src/renderer/features/network/NetworksList/lib/networks-list-utils.ts
Tbaut: `getStatusMetrics` sounds better I agree, I'll change

src/renderer/features/network/NetworksList/ui/NetworksList.tsx
Tbaut: agreed, I'll change

src/renderer/features/network/NetworksList/ui/ActiveNetwork.tsx
Tbaut: I think `<Trans>` and `t(` achieve exactly the same thing, and are interchangeable, but one is more for when you're actually having components, and the other inside some logic. If it translates the key into the english string now, I don't see why there would be a problem with language switcher. 

> And what about i18next/no-literal-string

Good point, this is accounted for according to the doc https://github.com/edvardchen/eslint-plugin-i18next/blob/HEAD/docs/rules/no-literal-string.md#selector-options

> But it should be another PR if you want to do this for hole project.

I don't intend to change the whole project, just to allow `<Trans>` to live alongside `t(` and use whichever feels cleaner depending on the context. What do you think?

src/renderer/pages/Settings/Networks/Networks.tsx
```diff
@@ -201,7 +202,11 @@ export const Networks = () => {
           )}
         </NetworkList>
 
-        <NetworkList isDefaultOpen title={t('settings.networks.activeNetworksLabel')} networkList={activeNetworks}>
+        <NetworkList
+          query={filterQuery}
```
Tbaut, line 208: Unless I missed something, there's no need to pass this as prop here, it can be fetched directly in the `NetworkList` component with `useUnit`

src/renderer/features/matrix/MatrixLogin/ui/LoginForm.test.tsx
tuul-wq: Yeah, but it caused errors
As I remember it was added in some other PR, that's ok

src/renderer/pages/Settings/Networks/Networks.tsx
tuul-wq: `NetworkList` and `networkFilterModel` are both features and cannot be imported one inside another

src/renderer/features/network/NetworksList/ui/EmptyNetworks.tsx
tuul-wq: They are grouped in a single feature - `NetworkList`
Explicit division inside feature is more flexible and maintainable
One day `activeNetworkModel`/`inactiveNetworkModel` might grow and lead to the big mess if they were aggregated in a single file/model

src/renderer/features/network/NetworksList/ui/ActiveNetwork.tsx
tuul-wq: @Asmadek @Tbaut Take a look into `StatusLabel`
```javascript
{typeof title === 'string' ? <FootnoteText className={TitleStyles[variant]}>{title}</FootnoteText> : title}
```
If we switch to `<Trans>` we have to go further and duplicate `<FootnoteText className={TitleStyles[variant]}>{title}</FootnoteText>` which is clearly not needed in `ActiveNetwork.tsx`
So I reverted back to `t(...)`

src/renderer/features/network/NetworksList/model/inactive-networks-model.ts
tuul-wq: > are very much related

But not the same. After I've discovered fsd & effector I've became very suspicious about so called "duplication" and urge to combine everything that looks or behaves similar.
These models serves different purposes

src/renderer/features/network/NetworksList/ui/ActiveNetwork.tsx
tuul-wq: If you find a good solution how to make TS happy and remove this redundant status form `ActiveNetwork`, that be awesome 🤩

src/renderer/features/network/NetworksList/model/inactive-networks-model.ts
tuul-wq: > I wonder if this in particular is what you mentioned in a conversation we had in the past regarding code duplication, that we shouldn't be afraid of.

Yes, it's exactly that

src/renderer/pages/Settings/Networks/Networks.tsx
Tbaut: Re-reading the FSD doc, I understand what you mean thanks.

src/renderer/features/network/NetworksList/ui/ActiveNetwork.tsx
Tbaut: Thanks, I hadn't seen this logic `StatusLabel`

src/renderer/features/network/NetworksList/ui/ActiveNetwork.tsx
Tbaut: What I meant is more: why do we display "connecting" when it's actually disabled. Isn't this misleading users?

Tbaut:
I got the test to at least not crash with weird errors anymore. However, the most basic test to render the Network component still fails because we expect the network list to show, but the list is empty.

This is to me an unexpected effector behavior. I've watched the `filteredNetworks` inside `src/renderer/pages/Settings/Networks/model/networks-overview-model.ts` and it is empty. logging it in the `Networks.tsx` it does print the expected chains. All other stores seem to have the right state.

I'd love to get your eyes on this @Asmadek or @tuul-wq when you have the chance, it may be obvious to you.

src/renderer/pages/Settings/Networks/Networks.tsx
```diff
@@ -38,6 +40,10 @@ export const Networks = () => {
   const inactiveNetworks = useUnit(inactiveNetworksModel.$inactiveNetworks);
   const connections = useUnit(networkModel.$connections);
   const filterQuery = useUnit(networksFilterModel.$filterQuery);
+  const filteredNetworks = useUnit(networksFilterModel.$filteredNetworks);
+
+  console.log('filteredNetworks', filteredNetworks);
```
Tbaut: This log is printing the expected chains, but the `sample` in `src/renderer/pages/Settings/Networks/model/networks-overview-model.ts` may not be triggered. Or at least, watching the `$filteredNetworks` prints an empty array.

Jess25051980 (approved):

tuul-wq (approved):
We made it 🚀

PR #1522, opened 2024-02-28
fix: Contact-filter effector refactor and tests

### LINKED ISSUE #1504: Refactor Contact form with Effector best practices
As per https://github.com/novasamatech/nova-spektr/pull/1474#discussion_r1502985897 the contact form model should be refactored a little to use effector best practices
- use `sample` instead of `$store.on`
- use `formInitiated` instead of `.reset`

Tbaut:
closes https://github.com/novasamatech/nova-spektr/issues/1504


src/renderer/features/contacts/ContactFilter/model/__tests__/contact-filter.test.tsx
```diff
@@ -0,0 +1,64 @@
+import { fork } from 'effector';
+
+import { filterModel } from '@features/contacts';
```
tuul-wq: Better to use relative import `from '../contact-filter-model'`;

src/renderer/features/contacts/ContactFilter/model/__tests__/contact-filter.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+};
+const getScopeByQuery = (query: string) => {
+  return fork({
+    values: new Map().set(contactModel.$contacts, [mockContact1, mockContact2]).set(filterModel.$filterQuery, query),
+  });
+};
+
+describe('features/contacts/model/contact-filter-model', () => {
+  afterEach(() => {
+    jest.clearAllMocks();
```
tuul-wq: `jest.clearAllMocks` is redundant here (this seems to be a mistake, because we should use `jest.restoreAllMocks` to restore spies - [doc](https://jestjs.io/docs/jest-object#jestrestoreallmocks))
Following code doesn't contain any spies or whatever mocks

src/renderer/features/contacts/ContactFilter/model/__tests__/contact-filter.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+    expect(scope.getState(filterModel.$contactsFiltered)[0].id).toBe(mockContact1.id);
+  });
+
+  test('should search by matrix id', async () => {
+    const scope = getScopeByQuery(mockContact2.matrixId);
+
+    expect(scope.getState(filterModel.$contactsFiltered)).toHaveLength(1);
+    expect(scope.getState(filterModel.$contactsFiltered)[0].id).toBe(mockContact2.id);
+  });
+});
```
tuul-wq, line 71: `await allSettled(...)` is a mandatory action to make sure all events/effects settle down - [doc](https://effector.dev/en/api/effector/allsettled/)
I see that you pass concrete query into the scope, that's not quite correct
In real life we call event `queryChanged` and hope for a change to happen, and test must do same stuff
- preset default values if needed
- call event(s)
- wait for allSettled
- assert the value

src/renderer/features/contacts/ContactFilter/model/__tests__/contact-filter.test.tsx
```diff
@@ -0,0 +1,64 @@
+import { fork } from 'effector';
+
+import { filterModel } from '@features/contacts';
+import { contactModel } from '@entities/contact';
+
+const mockContact1 = {
```
tuul-wq: I'd rather stick with a single variable
```javascript
const contacts = [ ... ];
```

src/renderer/features/contacts/ContactFilter/model/__tests__/contact-filter.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+};
+
+const mockContact2 = {
+  id: 5,
+  name: 'name 6',
+  address: '777',
+  accountId: '0x888',
+  matrixId: '@999',
+};
+const getScopeByQuery = (query: string) => {
```
tuul-wq: It's ok to declare a scope in each `test case`

src/renderer/features/contacts/ContactFilter/model/contact-filter.ts
```diff
@@ -3,17 +3,24 @@ import { combine, createEvent, createStore, sample } from 'effector';
...
 const queryChanged = createEvent<string>();
-const queryReset = createEvent();
 
-$filterQuery.on(queryChanged, (_, query) => query).reset(queryReset);
+sample({
+  clock: queryChanged,
+  target: $filterQuery,
+});
+
+sample({
```
tuul-wq: Better to swap these `samples` so you read the code like a book starting from `formInitiated` event

tuul-wq (commented):
Few suggestions to increase code quality

Asmadek (approved):

src/renderer/features/contacts/ContactFilter/model/__tests__/contact-filter.test.tsx
Tbaut: Thanks, that's the result of copy/pasting from other tests I found in the code.. and I can't say I understand 100% :)

src/renderer/features/contacts/ContactFilter/model/__tests__/contact-filter.test.tsx
Tbaut: I did this so that I can use specific contact details in the query. Look at the last 2 tests, they are IMHO much cleaner by doing `getScopeByQuery(mockContact1.address);` rather than `getScopeByQuery("222")` and then in the "expect" we use `mockContact1` rather than some value that aren't obvious.

TL;DR to me it's cleaner to read 

- search contact1.address
- expect contact1.id

rather than
- search "222"
- expect id "5"


Also I explicitly set all the field value to different numbers to make the test more robust, and make sure the only possible reason a contact is displayed, if because a specific field match.

src/renderer/features/contacts/ContactFilter/model/__tests__/contact-filter.test.tsx
Tbaut: edit: I got you, I declared 1 scope now that it's not depending on the `filterQuery` any more, no need to call it every time then.

src/renderer/features/contacts/ContactFilter/model/__tests__/contact-filter.test.tsx
Tbaut: Thank you so much for the detailed explanation, this clears things up a lot.

src/renderer/features/contacts/ContactFilter/model/__tests__/contact-filter.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+    expect(scope.getState(filterModel.$contactsFiltered)).toHaveLength(0);
+  });
+
+  test('should search by name', async () => {
+    allSettled(filterModel.events.formInitiated, { scope });
+    allSettled(filterModel.events.queryChanged, { scope, params: mockContact1.name });
+
+    expect(scope.getState(filterModel.$contactsFiltered)).toHaveLength(1);
+    expect(scope.getState(filterModel.$contactsFiltered)[0].id).toBe(mockContact1.id);
```
tuul-wq: always use `toEqual` instead of `toBe`

src/renderer/features/contacts/ContactFilter/model/__tests__/contact-filter.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+const mockContact2 = {
+  id: 5,
+  name: 'name 6',
+  address: '777',
+  accountId: '0x888',
+  matrixId: '@999',
+};
+
+describe('features/contacts/model/contact-filter-model', () => {
+  const scope = fork({
```
tuul-wq: `scope` must be unique in every test
The whole point of `fork` - it creates isolated scope

> Creates an isolated instance of application. Primary purposes of this method are SSR and testing.
[From Doc](https://effector.dev/en/api/effector/fork/)

src/renderer/features/contacts/ContactFilter/model/__tests__/contact-filter.test.tsx
tuul-wq:
``` javascript
const contacts = [ ... ];
// in tests
allSettled(filterModel.events.queryChanged, { scope, params: contacts[0].matrixId });
```

src/renderer/features/contacts/ContactFilter/model/__tests__/contact-filter.test.tsx
tuul-wq: This way there is no need to call `allSettled(filterModel.events.formInitiated, { scope });` in every test (especially without `await` which is required)

tuul-wq (changes requested):

src/renderer/features/contacts/ContactFilter/model/__tests__/contact-filter.test.tsx
Tbaut: Makes sense thanks. I find the duplication of the exact same `fork` somewhat harder to read, because it feels like the scope would be different between tests. But I understand it may help convey the idea that all test is based on an entirely new one.

src/renderer/features/contacts/ContactFilter/model/__tests__/contact-filter.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+  },
+];
+
+describe('features/contacts/model/contact-filter-model', () => {
+  test('should return all contacts if no search query', async () => {
+    const scope = fork({
+      values: new Map().set(contactModel.$contacts, contacts),
+    });
+
+    expect(scope.getState(filterModel.$contactsFiltered)).toHaveLength(2);
```
tuul-wq: To not depend on specific value
```suggestion
    expect(scope.getState(filterModel.$contactsFiltered)).toHaveLength(contacts.length);
```

src/renderer/features/contacts/ContactFilter/model/__tests__/contact-filter.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+  });
+
+  test('should search by name', async () => {
+    const scope = fork({
+      values: new Map().set(contactModel.$contacts, contacts),
+    });
+
+    await allSettled(filterModel.events.queryChanged, { scope, params: contacts[0].name });
+
+    expect(scope.getState(filterModel.$contactsFiltered)).toHaveLength(1);
```
tuul-wq:
```suggestion
    expect(scope.getState(filterModel.$contactsFiltered)).toEqual(contacts[0]);
```

src/renderer/features/contacts/ContactFilter/model/__tests__/contact-filter.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  test('should search by address', async () => {
+    const scope = fork({
+      values: new Map().set(contactModel.$contacts, contacts),
+    });
+
+    await allSettled(filterModel.events.queryChanged, { scope, params: contacts[0].address });
+
+    expect(scope.getState(filterModel.$contactsFiltered)).toHaveLength(1);
+    expect(scope.getState(filterModel.$contactsFiltered)[0].id).toEqual(contacts[0].id);
```
tuul-wq:
```suggestion
    expect(scope.getState(filterModel.$contactsFiltered)).toEqual(contacts[0]);
```

src/renderer/features/contacts/ContactFilter/model/__tests__/contact-filter.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+  });
+
+  test('should search by matrix id', async () => {
+    const scope = fork({
+      values: new Map().set(contactModel.$contacts, contacts),
+    });
+
+    await allSettled(filterModel.events.queryChanged, { scope, params: contacts[1].matrixId });
+
+    expect(scope.getState(filterModel.$contactsFiltered)).toHaveLength(1);
```
tuul-wq:
```suggestion
    expect(scope.getState(filterModel.$contactsFiltered)).toEqual(contacts[1]);
```

tuul-wq (commented):
Last pack of suggestions 🚀

src/renderer/features/contacts/ContactFilter/model/__tests__/contact-filter.test.tsx
Tbaut: oh, if that object/array equality works that's amazing.

Tbaut:
Thanks for the suggestions, I went for removing most of the `length` tests, and rather verified `$contactsFiltered` array equality

src/renderer/features/contacts/ContactFilter/model/__tests__/contact-filter.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+    id: 5,
+    name: 'name 6',
+    address: '777',
+    accountId: '0x888',
+    matrixId: '@999',
+  },
+];
+
+describe('features/contacts/model/contact-filter-model', () => {
+  test('should return all contacts if no search query', async () => {
```
tuul-wq:
```suggestion
  test('should return all contacts if no search query', () => {
```

tuul-wq (approved):
Now this PR looks awesome 🤩

PR #1523, opened 2024-02-29
Fix: Error overlay

tuul-wq:
- Removed noisy error overlay in dev mode


tuul-wq:
@Tbaut thanks for pointing to the existing overlay 👍

Tbaut (approved):
Amazing, the overlays are gone! Thank you so much

Issue #1445, opened 2024-02-12
Create proxy operation for add/remove proxy


pgolovkin:
Create proxy operations:
1. Create/remove proxy 
2. Show validation error if proxy has no funds for network fee
3. The account schema is: my account ➝ (any, nonTransfer) proxied account ➝ add a new proxy for delegated to you wallet

DOD:
1. Update confirmation screens according [doc](https://docs.google.com/document/d/1EHcncdvGo4tpA6Gne6ioyjv69H_l48m3DUm8UIKH2WE/edit#heading=h.v4te3y8ejv18)

stepanLav:
Status verification: ✅

What has been verified:
- Create proxy operation
- Remove proxy operation
- Check that balance for deposit is enough
- Check that balance for fee is enough
- Check that this proxy already created

PR #1547, opened 2024-03-06
Feat: Create proxy

### LINKED ISSUE #1442: Create regular proxy account (delegated authority)
Create proxy screen
1. Build list of available proxy accounts 
2. Build list of available proxy type
3. Calculate proxy deposit
4. Validation
5. Confirm create proxy screen
6. Sign operation
7. If transaction executed successfully then save new proxy and update deposit in the database

Doc [link](https://docs.google.com/document/d/1xwzz-geiEvTxIPRPeAh_5R3PENuYjefzE5C-sRtaVtY/edit#heading=h.a2mbaii0vl7e)
Operation confirmation screens [link](https://docs.google.com/document/d/1EHcncdvGo4tpA6Gne6ioyjv69H_l48m3DUm8UIKH2WE/edit#heading=h.e2rr9hgusek3)
Schema with adding the Proxy account for Multisig account on the image below:

Image

tuul-wq:
- Add delegated authority
- 4 steps flow full Effector + FSD

image
image

closes #1442


src/renderer/app/index.tsx
```diff
@@ -20,7 +20,7 @@ log.variables.env = process.env.NODE_ENV;
 log.transports.console.format = '{y}/{m}/{d} {h}:{i}:{s}.{ms} [{env}#{version}]-{processType} [{level}] > {text}';
 log.transports.console.useStyles = true;
 
-Object.assign(console, log.functions);
+// Object.assign(console, log.functions);
```
pgolovkin: don't comment it!

src/renderer/entities/transaction/lib/common/types.ts
```diff
@@ -30,7 +30,7 @@ export type ITransactionService = {
   getTransactionFee: (transaction: Transaction, api: ApiPromise) => Promise<string>;
   getExtrinsicWeight: (extrinsic: SubmittableExtrinsic<'promise'>) => Promise<Weight>;
   getTxWeight: (transaction: Transaction, api: ApiPromise) => Promise<Weight>;
-  getTransactionDeposit: (threshold: Threshold, api: ApiPromise) => string;
+  getMultisigDeposit: (threshold: Threshold, api: ApiPromise) => string;
```
pgolovkin, line 33: Great! It's really meaningful!

src/renderer/entities/transaction/lib/extrinsicService.ts
```diff
@@ -454,3 +454,23 @@ export const wrapAsMulti = (
     },
   };
 };
+
+// TODO: finish in later tasks
```
pgolovkin: which task?

src/renderer/widgets/AddProxyModal/lib/add-proxy-utils.ts
```diff
@@ -0,0 +1,35 @@
...
+  isInitStep,
+  isConfirmStep,
+  isSignStep,
+  isSubmitStep,
+
+  hasMultisig,
+  hasProxy,
+};
+
+function isInitStep(step: Step): boolean {
```
pgolovkin, line 26: seems that it's not only proxy steps.

src/renderer/widgets/AddProxyModal/lib/submit-proxy-utils.ts
```diff
@@ -0,0 +1,19 @@
+import { SubmitStep } from './types';
+
+export const submitProxyUtils = {
+  isLoadingStep,
+  isSuccessStep,
+  isErrorStep,
+};
+
+function isLoadingStep(step: SubmitStep): boolean {
```
pgolovkin, line 9: seems it's not only proxy submitting steps.

tuul-wq: this file contains utility functions only for `Submit` step

src/renderer/widgets/AddProxyModal/lib/add-proxy-utils.ts
tuul-wq: this file contains utility functions only for `AddProxyModal` manager

src/renderer/entities/transaction/lib/extrinsicService.ts
tuul-wq: Create proxy operation for add/remove proxy #1445 
I will update the comment

src/renderer/entities/network/lib/hooks.ts
```diff
@@ -13,7 +13,7 @@ type NetworkData = {
   extendedChain: ExtendedChain;
 };
 
-export const useNetworkData = (chainId: ChainId): NetworkData => {
+export const useNetworkData = (chainId = '0x00' as ChainId): NetworkData => {
```
Tbaut, line 16: This default `chainId` is invalid and makes me think it's an error. I guess if we expect some undefined chainId, we could rather return early?

src/renderer/entities/transaction/lib/transactionService.ts
```diff
@@ -28,21 +28,107 @@ export type TxWrappers = WrapAsMulti | WrapAsProxy;
...
+  api: ApiPromise,
+  callback: (executed: boolean, params: ExtrinsicResultParams | string) => void,
+) => {
+  let extrinsicCalls = 0;
+
+  const callIndex = api.createType('Call', unsigned.method).callIndex;
+  const { method, section } = api.registry.findMetaCall(callIndex);
+
+  api.rpc.author
+    .submitAndWatchExtrinsic(tx, async (result) => {
```
Tbaut, line 62: Can you tell me more about the reason behind using `.submitAndWatch`? Knowing very little about how Walletconnect would handle that, I'd have expected that we could use `.send` that then gives the Tx events etc.

src/renderer/shared/ui/Dropdowns/Combobox/Combobox.tsx
```diff
@@ -14,40 +14,40 @@ import {
...
   disabled,
   position = 'down',
   theme = 'light',
+  onInput,
   onChange,
   ...inputProps
 }: Props) => {
-  const [query, setQuery] = useState('');
-
-  const filteredOptions = query
```
tuul-wq, line 37: Filtering must be done by component user

src/renderer/shared/ui/Typography/common/TextBase.ts
```diff
@@ -3,8 +3,8 @@ import { createElement } from 'react';
 import { cnTw } from '@shared/lib/utils';
 import { TypographyProps } from './types';
 
-const TextBase = ({ as = 'p', align = 'left', className, children }: TypographyProps) => {
+export const TextBase = ({ as = 'p', align = 'left', className, children }: TypographyProps) => {
+  if (!children) return null;
```
tuul-wq, line 7: Instead of doing this
```javascript
{myValue && <FootnoteText>{myValue}</FootnoteText>}
```
can do this with much cleaner code (if `myValue` is falsy it won't render):
```javascript
<FootnoteText>{myValue}</FootnoteText>
```

src/renderer/widgets/AddProxyModal/model/add-proxy-model.ts
tuul-wq, line 1: Model that is responsible for management of data flow from step to step (from feature to feature)
This way each step is fully isolated

src/renderer/widgets/AddProxyModal/model/confirm-model.ts
```diff
@@ -0,0 +1,71 @@
...
+type Input = {
+  chain: Chain;
+  account: Account;
+  signatory?: Account;
+  proxyType: ProxyType;
+  delegate: Address;
+  description: string;
+  transaction: Transaction;
+
+  oldProxyDeposit: string;
```
tuul-wq, line 17: Because `Fee / ProxyDeposit` are being calculated from UI, this data have to be stored
In future all fee calculations must be done in model (as instance with help of services)

src/renderer/widgets/AddProxyModal/ui/SubmitProxy.tsx
```diff
@@ -0,0 +1,56 @@
...
+  onClose: () => void;
+};
+
+export const SubmitProxy = ({ isOpen, onClose }: Props) => {
+  const { t } = useI18n();
+
+  const submitStore = useUnit(submitModel.$submitStore);
+  const { step, message } = useUnit(submitModel.$submitStep);
+
+  const { addTask } = useTaskQueue();
```
tuul-wq, line 26: Had to leave these hooks in UI with a bridge to the model, because they have `useState` inside 😔

src/renderer/widgets/RemoveProxy/ui/Confirmation.tsx
```diff
@@ -39,9 +39,7 @@ export const Confirmation = ({
   return (
     <div className="flex flex-col items-center pt-4 gap-y-4 pb-4 pl-5 pr-3">
       <div className="flex flex-col items-center gap-y-3 mb-2">
-        <div className="flex items-center justify-center shrink-0 w-15 h-15 box-border rounded-full border-[2.5px] border-icon-default">
-          <Icon name="proxied" size={42} />
-        </div>
+        <Icon name="proxyConfirm" size={60} />
```
tuul-wq, line 42: Fixed icon

webpack/webpack.renderer.dev.ts
```diff
@@ -29,10 +29,6 @@ const config = merge<WpConfig & WdsConfig>(sharedConfig, {
     client: {
       overlay: false,
     },
-
-    devMiddleware: {
```
tuul-wq: Rolled back webpack option to write build files to the disk during development

src/renderer/features/balances/model/balance-subscription-model.ts
Asmadek, line 1: do we need this empty file?

src/renderer/widgets/AddProxyModal/model/add-proxy-model.ts
```diff
@@ -0,0 +1,272 @@
...
+  target: $txWrappers,
+});
+
+sample({
+  clock: formModel.output.formSubmitted,
+  source: {
+    txWrappers: $txWrappers,
+    apis: networkModel.$apis,
+  },
+  fn: ({ txWrappers, apis }, formData) => {
```
Asmadek, line 89:
```suggestion
  fn: ({ txWrappers, apis }, { chain, account, signatory, delegate, proxyType } => {
```

src/renderer/widgets/AddProxyModal/model/add-proxy-model.ts
```diff
@@ -0,0 +1,272 @@
...
+  filter: (addProxyStore: AddProxyStore | null): addProxyStore is AddProxyStore => Boolean(addProxyStore),
+  fn: (addProxyStore) => [
+    {
+      accountId: toAccountId(addProxyStore.delegate),
+      proxiedAccountId: addProxyStore.account.accountId,
+      chainId: addProxyStore.chain.chainId,
+      proxyType: addProxyStore.proxyType,
+      delay: 0,
+    },
+  ],
```
Asmadek, line 191:
```suggestion
  fn: (addProxyStore) => {
    const newProxy = {
      accountId: toAccountId(addProxyStore.delegate),
      proxiedAccountId: addProxyStore.account.accountId,
      chainId: addProxyStore.chain.chainId,
      proxyType: addProxyStore.proxyType,
      delay: 0,
    };
    
    return [newProxy];
  },
```

src/renderer/widgets/AddProxyModal/lib/add-proxy-utils.ts
```diff
@@ -0,0 +1,40 @@
...
+  return step === Step.SUBMIT;
+}
+
+function hasMultisig(txWrappers: TxWrappers): boolean {
+  return txWrappers.includes('multisig');
+}
+
+function hasProxy(txWrappers: TxWrappers): boolean {
+  return txWrappers.includes('proxy');
+}
```
Asmadek, line 48: This functions and `TxWrappers` type looks like really reusable for all other types of transactions

src/renderer/widgets/AddProxyModal/model/add-proxy-model.ts
```diff
@@ -0,0 +1,272 @@
...
+    const { chain, account, signatory, delegate, proxyType } = formData;
+
+    const transaction: Transaction = {
+      chainId: chain.chainId,
+      address: toAddress(account.accountId, { prefix: chain.addressPrefix }),
+      type: TransactionType.ADD_PROXY,
+      args: { delegate, proxyType, delay: 0 },
+    };
+
+    return txWrappers.reduce<{ transaction: Transaction; multisigTx: Transaction | null }>(
```
Asmadek: Maybe move it to variable for better readability?

src/renderer/widgets/AddProxyModal/model/add-proxy-model.ts
```diff
@@ -0,0 +1,272 @@
...
+import { submitModel } from './submit-model';
+
+const stepChanged = createEvent<Step>();
+
+const flowStarted = createEvent();
+const flowFinished = createEvent();
+
+const $step = createStore<Step>(Step.NONE);
+
+const $addProxyStore = createStore<AddProxyStore | null>(null);
```
Asmadek, line 26: what do you think about `addProxyState` or smth else, because `addProxyStore` sounds really wide

src/renderer/widgets/AddProxyModal/model/add-proxy-model.ts
```diff
@@ -0,0 +1,272 @@
...
+
+sample({
+  clock: flowFinished,
+  fn: () => Step.NONE,
+  target: stepChanged,
+});
+
+export const addProxyModel = {
+  $step,
+  $chain: $addProxyStore.map((store) => store?.chain, { skipVoid: false }),
```
Asmadek, line 249: New method for me 👍

src/renderer/widgets/AddProxyModal/model/submit-model.ts
```diff
@@ -0,0 +1,235 @@
...
+
+type SignedExtrinsicParams = {
+  api: ApiPromise;
+  signature: HexString;
+  unsignedTx: UnsignedTransaction;
+};
+const getSignedExtrinsicFx = createEffect(({ api, signature, unsignedTx }: SignedExtrinsicParams): Promise<string> => {
+  return getSignedExtrinsic(unsignedTx, signature, api);
+});
+type SubmitExtrinsicParams = {
```
Asmadek: add empty line here

src/renderer/widgets/AddProxyModal/ui/AddProxyForm.tsx
```diff
@@ -0,0 +1,378 @@
...
+        <SignatorySelector />
+        <ProxyCombobox />
+        <ProxyTypeSelector />
+        <DescriptionInput />
+      </form>
+      <div className="flex flex-col gap-y-6 pt-6 pb-4">
+        <FeeSection />
+        <FeeError />
+      </div>
+      <ButtonsSection onGoBack={onGoBack} />
```
Asmadek: What do you think about `ActionSection`, or `ActionButtons`?

src/renderer/widgets/AddProxyModal/ui/SubmitProxy.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const { addMultisigTx } = useMultisigTx({ addTask });
+  const { addEventWithQueue } = useMultisigEvent({ addTask });
+
+  useEffect(() => {
+    submitModel.events.hooksApiChanged({ addMultisigTx, addEventWithQueue });
+  }, [addMultisigTx, addEventWithQueue]);
```
Asmadek, line 32: Am I right that this is the same as `gate` or is there some difference?

src/renderer/widgets/AddProxyModal/ui/AddProxyForm.tsx
```diff
@@ -0,0 +1,378 @@
...
+        onChange={({ value }) => signatory.onChange(value)}
+      />
+      <InputHint variant="error" active={signatory.hasError()}>
+        {t(signatory.errorText())}
+      </InputHint>
+    </div>
+  );
+};
+
+const ProxyCombobox = () => {
```
Asmadek: I don't clearly understand why this is a combobox instead of an input?

src/renderer/widgets/AddProxyModal/model/form-model.ts
```diff
@@ -0,0 +1,590 @@
...
+    balances: balanceModel.$balances,
+  },
+  ({ wallet, accounts, chain, balances }) => {
+    if (!wallet || !chain) return [];
+
+    const isPolkadotVault = walletUtils.isPolkadotVault(wallet);
+    const walletAccounts = accountUtils.getWalletAccounts(wallet.id, accounts).filter((account) => {
+      if (isPolkadotVault && accountUtils.isBaseAccount(account)) return false;
+
+      return accountUtils.isChainIdMatch(account, chain.chainId);
```
Asmadek: I think crypto type should be checked too

src/renderer/widgets/AddProxyModal/model/form-model.ts
```diff
@@ -0,0 +1,590 @@
...
+    const signers = dictionary(account.signatories, 'accountId', () => true);
+
+    return wallets.reduce<{ signer: Account; balance: string }[]>((acc, wallet) => {
+      const walletAccounts = accountUtils.getWalletAccounts(wallet.id, accounts);
+      const isAvailable = permissionUtils.canCreateMultisigTx(wallet, walletAccounts);
+
+      if (!isAvailable) return acc;
+
+      const signer = walletAccounts.find((a) => {
+        return signers[a.accountId] && accountUtils.isChainIdMatch(a, chain.chainId);
```
Asmadek: I think crypto type should be checked too

src/renderer/widgets/AddProxyModal/model/form-model.ts
```diff
@@ -0,0 +1,590 @@
...
+      });
+
+      if (signer) {
+        const balance = balances.find((balance) => {
+          return (
+            balance.chainId === chain.chainId &&
+            balance.accountId === signer.accountId &&
+            balance.assetId === chain.assets[0].assetId.toString()
+          );
+        });
```
Asmadek: looks like candidate for util function

Asmadek (changes requested):
some comments

src/renderer/entities/network/lib/hooks.ts
tuul-wq: Right now it's mostly for old pages without `effector`
In future this hook won't be needed
Also '0x00' was just moved into the hook from components where chainId could be `undefined`

src/renderer/entities/transaction/lib/transactionService.ts
tuul-wq: WalletConnect signing is not connected with `submitAndWatchExtrinsic`
`submitAndWatchExtrinsic` happens after signing was already made

src/renderer/widgets/AddProxyModal/model/add-proxy-model.ts
tuul-wq: I have intentionally avoided destructuring to reduce the mess in params

src/renderer/widgets/AddProxyModal/model/add-proxy-model.ts
tuul-wq: I'd rather not introduce additional variable that's being immediately returned
This patters better works for conditional variables

src/renderer/widgets/AddProxyModal/model/add-proxy-model.ts
tuul-wq: For me `state` is something temporary and closer to UI components, store is something more permanent and a bit better aligns with `effector` terminology 
I agree that `store` sounds wide, don't really know what is a better option for a structure that just holds a bunch of data relevant for that specific model 🤔

src/renderer/widgets/AddProxyModal/model/add-proxy-model.ts
tuul-wq: Could have introduced `combine` store or `.map` store like this one
But because it's not used anywhere else, decided to declare it here

src/renderer/widgets/AddProxyModal/lib/add-proxy-utils.ts
tuul-wq: Agree, but I suggest to copy paste for a while
When we have time, we should think about appropriate place for these functions as well as factories for operation forms

src/renderer/widgets/AddProxyModal/ui/SubmitProxy.tsx
tuul-wq: I assume it's the same as Gate (not a big fan of it, another react specific function)
Event style is just more generic
Also we used to do `apiChanged` in other components, good to be consistent

src/renderer/entities/transaction/lib/transactionService.ts
Tbaut: Ok, so do you know why we're using `submitAndWatchExtrinsic` rather than [`.send`](https://github.com/polkadot-js/api/blob/390610127a350a8ebb5bac2b8c6645c50060a7c6/packages/api-base/src/types/submittable.ts#L65)? the send allows to pass a callback, and will give you all the status, events right away, without the need to subscribe.

Asmadek (approved):

src/renderer/entities/transaction/lib/transactionService.ts
tuul-wq: @Asmadek need your help with Thibaut's question

Asmadek: I don't remember well the reasons for using `submitAndWatchExtrinsic`, as this decision was made over two years ago during the development of the PoC. But I am not deeply familiar with this method. @Tbaut can you describe pros of usage `send`? because for first look it pass very similar callback as for `submitAndWatchExtrinsic`

Tbaut: The idea is that you woulnd't need to do all the following
```ts
      const signedBlock = await api.rpc.chain.getBlock();
      const blockHeight = signedBlock.block.header.number.toNumber();
      const apiAt = await api.at(signedBlock.block.header.hash);
      const allRecords = await apiAt.query.system.events();
```

But rather directly access the status and event as explained [here](https://polkadot.js.org/docs/api/start/api.tx.subs#transaction-events), or [here](https://polkadot.js.org/docs/api/cookbook/tx#how-do-i-get-the-decoded-enum-for-an-extrinsicfailed-event) e.g:

```ts
.send(sender, ({ status, events }) => {
    if (status.isInBlock || status.isFinalized) {
      events
        // find/filter for failed events
        .filter(({ event }) =>
          api.events.system.ExtrinsicFailed.is(event)
        )
        // we know that data for system.ExtrinsicFailed is
        // (DispatchError, DispatchInfo)
        .forEach(({ event: { data: [error, info] } }) => {
          if (error.isModule) {
            // for module errors, we have the section indexed, lookup
            const decoded = api.registry.findMetaError(error.asModule);
            const { docs, method, section } = decoded;

            console.log(`${section}.${method}: ${docs.join(' ')}`);
          } else {
            // Other, CannotLookup, BadOrigin, no extra info
            console.log(error.toString());
          }
        });
    }
```

Asmadek: Looks reasonable, maybe we can create task to refactor it, @tuul-wq what do you think?

tuul-wq: @Asmadek @Tbaut I agree, will create task for it
Thanks for such a valuable feedback!
update: [task](https://github.com/novasamatech/nova-spektr/issues/1585)

PR #1548, opened 2024-03-06
feat: check proxy variant on add

### LINKED ISSUE #1483: Distinguish pure proxy and regular proxy

Asmadek:
* use subquery to check proxied variant on add

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,6 +106,7 @@
     "electron-updater": "^6.1.1",
     "electron-window-state": "^5.0.3",
     "graphql": "^16.8.1",
+    "graphql-request": "^6.1.0",
     "i18next": "^21.8.13",
     "lodash": "^4.17.21",
     "lottie-react": "^2.4.0",
--- a/src/renderer/entities/proxy/api/graphql/queries/pureProxies.ts
+++ b/src/renderer/entities/proxy/api/graphql/queries/pureProxies.ts
@@ -1,0 +1,11 @@
+import { gql } from 'graphql-request';
+
+export const CHECK_PURE_PROXIES = gql`
+  query PureProxies($accountIds: [String!]) {
+    pureProxies(filter: { id: { in: $accountIds } }) {
+      nodes {
+        id
+      }
+    }
+  }
+`;
--- a/src/renderer/entities/proxy/api/graphql/queries/pureProxies.ts
+++ b/src/renderer/entities/proxy/api/graphql/queries/pureProxies.ts
@@ -1,6 +1,6 @@
 import { gql } from 'graphql-request';
 
-export const CHECK_PURE_PROXIES = gql`
+export const FILTER_PURE_PROXIED_ACCOUNT_IDS = gql`
   query PureProxies($accountIds: [String!]) {
     pureProxies(filter: { id: { in: $accountIds } }) {
       nodes {
--- a/src/renderer/entities/proxy/api/pureProxiesService.ts
+++ b/src/renderer/entities/proxy/api/pureProxiesService.ts
@@ -1,0 +1,14 @@
+import { GraphQLClient } from 'graphql-request';
+
+import { CHECK_PURE_PROXIES } from './graphql/queries/pureProxies';
+import type { AccountId } from '@shared/core';
+
+export const checkPureProxies = async (client: GraphQLClient, accountIds: AccountId[]): Promise<AccountId[]> => {
+  const data = await client.request(CHECK_PURE_PROXIES, {
+    accountIds,
+  });
+
+  const checkedProxies = (data as any)?.pureProxies?.nodes?.map((node: any) => node.id) || [];
+
+  return checkedProxies;
+};
--- a/src/renderer/entities/proxy/api/pureProxiesService.ts
+++ b/src/renderer/entities/proxy/api/pureProxiesService.ts
@@ -1,14 +1,16 @@
 import { GraphQLClient } from 'graphql-request';
 
-import { CHECK_PURE_PROXIES } from './graphql/queries/pureProxies';
+import { FILTER_PURE_PROXIED_ACCOUNT_IDS } from './graphql/queries/pureProxies';
 import type { AccountId } from '@shared/core';
 
-export const checkPureProxies = async (client: GraphQLClient, accountIds: AccountId[]): Promise<AccountId[]> => {
-  const data = await client.request(CHECK_PURE_PROXIES, {
-    accountIds,
-  });
-
-  const checkedProxies = (data as any)?.pureProxies?.nodes?.map((node: any) => node.id) || [];
-
-  return checkedProxies;
+export const pureProxiesService = {
+  filterPureProxiedAccountIds,
 };
+
+async function filterPureProxiedAccountIds(client: GraphQLClient, accountIds: AccountId[]): Promise<AccountId[]> {
+  const data = await client.request(FILTER_PURE_PROXIED_ACCOUNT_IDS, { accountIds });
+
+  const filteredAccountIds = (data as any)?.pureProxies?.nodes?.map((node: any) => node.id);
+
+  return filteredAccountIds || [];
+}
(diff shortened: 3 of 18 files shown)


package.json
```diff
@@ -106,6 +106,7 @@
     "electron-updater": "^6.1.1",
     "electron-window-state": "^5.0.3",
     "graphql": "^16.8.1",
+    "graphql-request": "^6.1.0",
```
pgolovkin, line 109: As I remember the we're using graphql for getting staking onformation. Why this additional import needed?

Asmadek: Now we use library for react, but here we need plain js solution

src/renderer/entities/proxy/api/pureProxiesService.ts
```diff
@@ -0,0 +1,14 @@
+import { GraphQLClient } from 'graphql-request';
+
+import { CHECK_PURE_PROXIES } from './graphql/queries/pureProxies';
+import type { AccountId } from '@shared/core';
+
+export const checkPureProxies = async (client: GraphQLClient, accountIds: AccountId[]): Promise<AccountId[]> => {
+  const data = await client.request(CHECK_PURE_PROXIES, {
```
tuul-wq: No need for multiline 
```suggestion
  const data = await client.request(CHECK_PURE_PROXIES, { accountIds });
```

src/renderer/entities/proxy/api/pureProxiesService.ts
```diff
@@ -0,0 +1,14 @@
...
+    accountIds,
+  });
+
+  const checkedProxies = (data as any)?.pureProxies?.nodes?.map((node: any) => node.id) || [];
+
+  return checkedProxies;
```
tuul-wq:
```suggestion
  const checkedProxies = (data as any)?.pureProxies?.nodes?.map((node: any) => node.id);

  return checkedProxies || [];
```

src/renderer/entities/proxy/api/pureProxiesService.ts
```diff
@@ -0,0 +1,14 @@
+import { GraphQLClient } from 'graphql-request';
+
+import { CHECK_PURE_PROXIES } from './graphql/queries/pureProxies';
+import type { AccountId } from '@shared/core';
+
+export const checkPureProxies = async (client: GraphQLClient, accountIds: AccountId[]): Promise<AccountId[]> => {
```
tuul-wq: Function name is misleading, should be renamed
It actually requests data and returns `ids` but name `check` tells us nothing
Something like: `getAccountIdsWithPureProxies` don't quite like it, can discuss other variants

src/renderer/entities/proxy/api/graphql/queries/pureProxies.ts
```diff
@@ -0,0 +1,11 @@
+import { gql } from 'graphql-request';
+
+export const CHECK_PURE_PROXIES = gql`
```
tuul-wq: Same for the constant name

src/renderer/entities/proxy/lib/proxy-utils.ts
```diff
@@ -49,8 +49,18 @@ function isSameProxyGroup(oldGroup: NoID<ProxyGroup>, newGroup: NoID<ProxyGroup>
...
+  proxyType: ProxyType,
+  proxyVariant: ProxyVariant,
+  addressPrefix?: number,
+): string {
+  const address = toAddress(accountId, {
+    prefix: addressPrefix,
+    chunk: 6,
+  });
+
+  return `${proxyType} ${proxyVariant === ProxyVariant.PURE ? 'for pure' : 'for'} ${address}`;
```
tuul-wq: I guess having a variable for `proxyVarian` is a good idea, like you did for `address`

src/renderer/features/proxies/lib/proxies-utils.ts
```diff
@@ -42,7 +42,12 @@ function getNotification({
       proxyWalletName: accountsWalletsMap[proxied.proxyAccountId].name,
       proxyWalletType: accountsWalletsMap[proxied.proxyAccountId].type,
       proxiedAccountId: proxied.accountId,
-      proxiedWalletName: proxyUtils.getProxiedName(proxied.accountId, proxied.proxyType, addressPrefix),
+      proxiedWalletName: proxyUtils.getProxiedName(
```
tuul-wq: For  `proxyUtils.getProxiedName()` If we change first 3 arguments to a single one - `account: ProxiedAccount`
wouldn't that be more convenient? Or at least `account: Pick<ProxiedAccount, 'accountId' | 'proxyType' | 'proxyVariant'>`

src/renderer/shared/core/types/chain.ts
```diff
@@ -15,7 +15,7 @@ export type Chain = {
   options?: ChainOptions[];
 };
 
-export type ChainOptions = 'testnet' | 'crowdloans' | 'multisig' | 'regular_proxy' | 'ethereum_based';
+export type ChainOptions = 'testnet' | 'crowdloans' | 'multisig' | 'regular_proxy' | 'pure_proxy' | 'ethereum_based';
```
tuul-wq: This `type` has gotten so many options, nice to have enum instead of endless options of strings 😎
Could you create an issue and link it here with TODO?

src/renderer/shared/core/types/chain.ts
```diff
@@ -35,5 +35,5 @@ type ExternalValue = {
   url: string;
 };
 
-type ExternalType = 'history' | 'staking' | 'crowdloans';
+type ExternalType = 'history' | 'staking' | 'crowdloans' | 'proxy';
```
tuul-wq: Let's convert `ExternalType` to enum as well within `ChainOptions` task I've mention above

src/renderer/features/proxies/workers/proxy-worker.ts
```diff
@@ -228,6 +232,21 @@ async function getProxies({
   });
   console.log(`proxy-worker ${api.genesisHash}: 🟣 proxied accounts to remove: `, proxiedAccountsToRemove);
 
+  if (proxyUrl && proxiedAccountsToAdd.length) {
```
tuul-wq: This should not be done inside the worker
Much more cleaner way is to call QGL inside `getProxiesFx` after we receive data from `endpoint.call.getProxies`
enrich `proxiedAccountsToAdd` if needed and return the data

src/renderer/features/proxies/workers/proxy-worker.ts
```diff
@@ -98,6 +98,7 @@ async function disconnect(chainId: ChainId) {
 
 type GetProxiesParams = {
   chainId: ChainId;
+  proxyUrl: string;
```
tuul-wq: Isn't used anywhere inside `getProxies`

src/renderer/shared/core/types/chain.ts
```diff
@@ -35,5 +42,11 @@ type ExternalValue = {
   url: string;
 };
 
-type ExternalType = 'history' | 'staking' | 'crowdloans';
+// TODO: make it an enum
+export enum ExternalType {
```
tuul-wq:
```suggestion
export const enum ExternalType {
```

src/renderer/shared/core/types/chain.ts
```diff
@@ -15,7 +15,14 @@ export type Chain = {
   options?: ChainOptions[];
 };
 
-export type ChainOptions = 'testnet' | 'crowdloans' | 'multisig' | 'regular_proxy' | 'ethereum_based';
+export enum ChainOptions {
```
tuul-wq:
```suggestion
export const enum ChainOptions {
```

tuul-wq (commented):
Few suggestions

tuul-wq (approved):

pgolovkin (approved):

PR #1571, opened 2024-03-14
Feat: refactor custom rpc node creation/edition with FSD + Effector

### LINKED ISSUE #1454: Rewrite creation and removal of custom RPC with FSD + Effector
### Description
In order to increase maintainability and testability of managing custom RPC nodes, we must rewrite that part of code according to FSD with help of `effector`.

What needs to be done:
- [ ] Add new RPC node feature
- [ ] Edit RPC node feature
- [ ] Use `effector-forms` instead of `react-hook-form`
- [ ] Persist existing form validation
- [ ] Tests for add rpc
- [ ] Tests for edit rpc

Reference model: `features/../CurrencyForm` & `features/../CreateContactForm`

### LINKED ISSUE #1542: Can't add a custom RPC the modal closes
## Description
The modal to add a custom RPC closes itself automatically when switching field, or just clicking on it (because the network modal itself is closing)

https://github.com/novasamatech/nova-spektr/assets/33178835/4f944ed2-522c-4881-b515-dea081efe6ee

## Steps to Reproduce
See video

I'll fix it together with https://github.com/novasamatech/nova-spektr/issues/1454

Tbaut:
closes https://github.com/novasamatech/nova-spektr/issues/1542
closes #1454 

--- update: 17.04.2024 ---
- Add / edit / remove custom node moved to distinct feature
- Updated Networks page & model to become a mediator between network-selector & manage-custom-rpc
- Removed rpc validation from service
- Made node switch more robust, based on events not delay(500ms)


Tbaut:
I am still flying blind in terms of tests, because it crashes with a memory leak more often than not, so I'll continue working on this and make sure they all pass. In the mean time I would love to have your comments on the current state of this PR.

src/renderer/pages/Settings/Networks/components/index.ts
```diff
@@ -1,2 +1,2 @@
-export { CustomRpcModal } from './CustomRpcModal/CustomRpcModal';
+export { AddCustomRpcModal } from '../../../../features/network/CustomRpcForm/ui/AddCustomRpcModal';
```
Asmadek:
```suggestion
export { AddCustomRpcModal } from '@features/network';
```

src/renderer/pages/Settings/Networks/components/NetworkSelector/NetworkSelector.tsx
```diff
@@ -38,10 +39,14 @@ export const NetworkSelector = ({
   onDisconnect,
   onConnect,
   onRemoveCustomNode,
-  onChangeCustomNode,
 }: Props) => {
   const { t } = useI18n();
   const [ref, scroll] = useScrollTo<HTMLDivElement>(TRANSITION_DURATION);
+  const openAddCustomNodeModal = useUnit(addCustomRpcModel.events.processStarted);
```
Asmadek: Not needed to use `useUnit` for events, just call `addCustomRpcModel.events.processStarted()`

src/renderer/pages/Settings/Networks/components/CustomRpcModal/AddCustomRpcModal.tsx
Asmadek, line 1: Do you need this file? It looks unused

src/renderer/features/network/CustomRpcForm/lib/utils.ts
```diff
@@ -0,0 +1,16 @@
+import { validateWsAddress } from '@shared/lib/utils';
+
+export const fieldRules = {
+  name: [
+    { name: 'required', errorText: 'settings.networks.requiredNameError', validator: Boolean },
+    {
+      name: 'minMaxLength',
+      errorText: 'settings.networks.maxLengthNameError',
+      validator: (val: string) => val.length <= 50 && val.length >= 3,
```
Asmadek: I suggest using constants instead of numbers

src/renderer/features/network/CustomRpcForm/model/add-custom-rpc-model.ts
```diff
@@ -0,0 +1,179 @@
...
+const $selectedNetwork = createStore<ExtendedChain | null>(null);
+const $rpcConnectivityResult = createStore<RpcCheckResult>(RpcCheckResult.INIT);
+const $isNodeExist = createStore<boolean>(false);
+const $isProcessStarted = createStore<boolean>(false);
+
+const checkRpcNodeFx = createEffect(
+  async ({ network, url }: { network: ExtendedChain | null; url: string }): Promise<RpcCheckResult | null> => {
+    if (!network) return null;
+
+    try {
```
Asmadek: Better to use `checkRpcNodeFx.fail` instead of try catch inside effect

src/renderer/features/network/CustomRpcForm/model/add-custom-rpc-model.ts
```diff
@@ -0,0 +1,179 @@
...
+      rpcConnectivityChecked(RpcCheckResult.LOADING);
```
Asmadek: Better to avoid use of events inside effect. better to cover this behaviour with `sample`

src/renderer/features/network/CustomRpcForm/model/add-custom-rpc-model.ts
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  const defaultNodes = network.nodes;
+  const customNodes = network.connection.customNodes || [];
+
+  const result = defaultNodes.some(({ url: u }) => u === url) || customNodes.some(({ url: u }) => u === url);
+
+  nodeExistChecked(result);
+
+  return result;
```
Asmadek: better to use sample

```typescript
sample({ clock: isNodeExistFx.doneData, target: nodeExistChecked })
```

src/renderer/features/network/CustomRpcForm/model/add-custom-rpc-model.ts
```diff
@@ -0,0 +1,179 @@
...
+
+const resetFormFx = createEffect(() => {
+  resetRpcValidationFx();
+  $addCustomRpcForm.reset;
+});
+
+const resetRpcValidationFx = createEffect(() => {
+  nodeExistChecked(false);
+  rpcConnectivityChecked(RpcCheckResult.INIT);
+});
```
Asmadek: better to avoid such effects which contain only events. Use `sample` instead

src/renderer/features/network/CustomRpcForm/model/add-custom-rpc-model.ts
```diff
@@ -0,0 +1,179 @@
...
+const processStarted = createEvent<boolean>();
+
+const $selectedNetwork = createStore<ExtendedChain | null>(null);
+const $rpcConnectivityResult = createStore<RpcCheckResult>(RpcCheckResult.INIT);
+const $isNodeExist = createStore<boolean>(false);
+const $isProcessStarted = createStore<boolean>(false);
+
+const checkRpcNodeFx = createEffect(
+  async ({ network, url }: { network: ExtendedChain | null; url: string }): Promise<RpcCheckResult | null> => {
+    if (!network) return null;
```
Asmadek: may be covered by `filter` in `sample`

src/renderer/features/network/CustomRpcForm/model/edit-custom-rpc-model.ts
Asmadek, line 1: same comments as for `add-custom-rpc-model.ts`

Asmadek (changes requested):

src/renderer/features/network/CustomRpcForm/index.ts
tuul-wq, line 1: I would have called this feature `ManageCustomRpcNode`
Otherwise `CustomRpcForm` contains `Add__Modal`/`Edit__Modal` seems awkward

src/renderer/features/network/CustomRpcForm/lib/types.ts
```diff
@@ -0,0 +1,16 @@
...
+  'INVALID',
+  'WRONG_NETWORK',
+}
+
+export type CustomRpcForm = {
+  name: string;
+  url: string;
+};
+
+export type NodeExistParam = { network: ExtendedChain | null; url: string };
```
tuul-wq:
```suggestion
export type NodeExistParam = {
  network: ExtendedChain | null;
  url: string;
};
```

src/renderer/features/network/CustomRpcForm/lib/utils.ts
tuul-wq:
```suggestion
      validator: (value: string) => value.length <= 50 && value.length >= 3,
```

src/renderer/features/network/CustomRpcForm/lib/utils.ts
tuul-wq, line 1: `utils` should contain functions
Would be right to create `constants.ts` and move the following code there
`export const FIELD_RULES = { ... };`

src/renderer/features/network/CustomRpcForm/lib/utils.ts
```diff
@@ -0,0 +1,16 @@
+import { validateWsAddress } from '@shared/lib/utils';
```
tuul-wq: Because `validateWsAddress` is not used anywhere else, I'd recommend to use it's source code inside `validator`

src/renderer/features/network/CustomRpcForm/model/__test__/add-custom-rpc-model.test.ts
```diff
@@ -0,0 +1,106 @@
+import { fork, allSettled } from 'effector';
+
+import { addCustomRpcModel } from '../add-custom-rpc-model';
+import { ConnectionStatus } from '@shared/core';
+import { ExtendedChain } from '@entities/network';
+import { RpcCheckResult } from '../../lib/types';
+
+describe('features/network/CustomRpcForm/add-custom-rpc-model', () => {
+  test('should have error for wrong node url', async () => {
+    const scope = fork({});
```
tuul-wq: This is also a valid expression 😉
```suggestion
    const scope = fork();
```

src/renderer/features/network/CustomRpcForm/model/__test__/add-custom-rpc-model.test.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    const { name, url } = addCustomRpcModel.$addCustomRpcForm.fields;
+    await allSettled(name.onChange, { scope, params: 'some name' });
```
tuul-wq: Would be great to create a helper function right after `describe` block
that accepts `name & url` call `allSetteled` for those fields and submit
(reference: `constructor-model.test.ts`)
```javascript
await allSettled(addCustomRpcModel.$addCustomRpcForm.submit, { scope });
```

src/renderer/features/network/CustomRpcForm/model/__test__/add-custom-rpc-model.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    await allSettled(url.onChange, { scope, params: 'wrong url' });
+    await allSettled(addCustomRpcModel.$addCustomRpcForm.validate, { scope });
+
+    expect(scope.getState(name.$errors)).toEqual([]);
+    expect(scope.getState(url.$errors)[0].rule).toEqual('wsAddressValidation');
```
tuul-wq: Nice 👍

src/renderer/features/network/CustomRpcForm/model/__test__/add-custom-rpc-model.test.ts
```diff
@@ -0,0 +1,106 @@
...
+    const scope = fork({});
+
+    const network = {
+      chainId: '0x01',
+      connection: { chainId: '0x01', id: '0x01' },
+      connectionStatus: ConnectionStatus.CONNECTED,
+      nodes: [{ url: 'wss://some-rpc.com', name: 'node' }],
+    };
+
+    expect(scope.getState(addCustomRpcModel.$isNodeExist)).toEqual(false);
```
tuul-wq: No need to check the default state

src/renderer/features/network/CustomRpcForm/model/__test__/add-custom-rpc-model.test.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    expect(scope.getState(addCustomRpcModel.$isNodeExist)).toEqual(true);
+  });
+
+  test('should have the wrong network state', async () => {
+    const scope = fork({});
+
+    const network = {
+      specName: 'polkadot',
+      chainId: '0x91b171bb158e2d3848fa23a9f1c25182fb8e20313b2c1eb49219da7a70ce90c3',
```
tuul-wq: We have stub values for these cases `TEST_CHAIN_ID`
Sometimes they're also used when we require fake transaction to calculate the fee

src/renderer/features/network/CustomRpcForm/model/__test__/edit-custom-rpc-model.test.ts
```diff
@@ -0,0 +1,110 @@
...
+  test('should have the form pre-filled on launch', async () => {
+    const scope = fork({});
+    const mockNode = { name: 'some name', url: 'some url' } as RpcNode;
+
+    await allSettled(editCustomRpcModel.events.nodeSelected, { scope, params: mockNode });
+
+    expect(scope.getState(editCustomRpcModel.$editCustomRpcForm.$values)).toEqual(mockNode);
+  });
+
+  test('should have error for wrong node url', async () => {
```
tuul-wq: Let's safely remove duplicated tests
When we provide a factory for `add/edit rpc` then tests will cover it including common validation rules

src/renderer/pages/Settings/Networks/components/index.ts
tuul-wq: No need to export `AddCustomRpcModal` from 'components/index', after `NetworkSelector` moving to features folder, components folder will disappear
`AddCustomRpcModal` could be imported directly into `Networks`

src/renderer/pages/Settings/Networks/components/NetworkSelector/NetworkSelector.tsx
```diff
@@ -11,6 +12,7 @@ import { CommonInputStyles, CommonInputStylesTheme } from '@shared/ui/Inputs/com
 import { ConnectionType } from '@shared/core';
 import type { Theme } from '@shared/ui/types';
 import type { RpcNode } from '@shared/core';
+import { addCustomRpcModel, editCustomRpcModel } from '@features/network';
```
tuul-wq: As soon as `NetworkSelector` moves to `features` this is going to be an invalid cross import
`NetworkSelector` should be responsible only for activating available options eg.: Light client, rpc node, disable.
And provide outputs for: `onEditNode`, `onRemoveNode`, `onAddNode`
For now these outputs can be declared as props, in future they will become model's outputs
```javascript
export const myModel = {
  output: {
    editRequested,
    removeRequested,
    additionRequested,
  },
};
```

This way things like `useUnit(addCustomRpcModel.events.processStarted)` must be called inside controller component, in our case `Networks`

src/renderer/pages/Settings/Networks/Networks.tsx
```diff
@@ -40,13 +42,17 @@ export const Networks = () => {
   const inactiveNetworks = useUnit(inactiveNetworksModel.$inactiveNetworks);
   const connections = useUnit(networkModel.$connections);
   const filterQuery = useUnit(networksFilterModel.$filterQuery);
+  const [isAddRpcModalOpen, setAddRpcModalOpen] = useUnit([
+    addCustomRpcModel.$isProcessStarted,
+    addCustomRpcModel.events.processStarted,
```
tuul-wq: As @Asmadek mentioned, events can be used without `useUnit` 
`useUnit` is required for events only in SSR - not to lose the scope

src/renderer/features/network/CustomRpcForm/ui/EditCustomRpcModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+    isValid,
+  } = useForm(editCustomRpcModel.$editCustomRpcForm);
+
+  const onSubmit = (event: FormEvent) => {
+    event.preventDefault();
+
+    submit();
+  };
+
+  const isLoading = rpcCheckResult === RpcCheckResult.LOADING;
```
tuul-wq: A simple question: "How UI component knows how to check `rpcCheckResult`?"
It only knows what and how to render 😉
There are 2 options:
- **okish**: this condition must be declared as a util function in `edit-custom-rpc-utils.ts`
- **best**: store inside a model (it might use above util as well)

src/renderer/features/network/CustomRpcForm/ui/EditCustomRpcModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+            </InputHint>
+            <InputHint variant="error" active={url?.hasError()}>
+              {t(url?.errorText())}
+            </InputHint>
+          </div>
+          <InputHint active={rpcCheckResult === RpcCheckResult.VALID} variant="success">
+            {t('settings.networks.addressConnected')}
+          </InputHint>
+          <Alert
+            active={rpcCheckResult === RpcCheckResult.INVALID}
```
tuul-wq: Same UI conditions here

src/renderer/features/network/CustomRpcForm/ui/EditCustomRpcModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+            <Input
+              label={t('settings.networks.nameLabel')}
+              placeholder={t('settings.networks.namePlaceholder')}
+              invalid={name.hasError()}
+              disabled={isLoading}
+              value={name.value}
+              onChange={name.onChange}
+            />
+            <InputHint variant="error" active={name?.hasError()}>
+              {t(name?.errorText())}
```
tuul-wq:
```suggestion
            <InputHint variant="error" active={name.hasError()}>
              {t(name.errorText())}
```

src/renderer/features/network/CustomRpcForm/ui/EditCustomRpcModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+              value={url.value}
+              invalid={url.hasError()}
+              disabled={isLoading}
+              onChange={url.onChange}
+            />
+            <InputHint active variant="hint">
+              {t('settings.networks.addressHint')}
+            </InputHint>
+            <InputHint variant="error" active={url?.hasError()}>
+              {t(url?.errorText())}
```
tuul-wq:
```suggestion
            <InputHint variant="error" active={url.hasError()}>
              {t(url.errorText())}
```

src/renderer/features/network/CustomRpcForm/ui/EditCustomRpcModal.test.tsx
tuul-wq, line 1: I suggest to completely remove UI tests
The only thing they do is check existence of elements which is much less important than business logic

Test `should disable submit button during submission` even tries to reproduce BL behaviour to check if button is disabled or not

tuul-wq (commented):
Part 1 of my review

src/renderer/features/network/CustomRpcForm/lib/utils.ts
Tbaut: Thanks, I moved `validateWsAddress` to utils, and this rules to `constants.ts`

src/renderer/features/network/CustomRpcForm/model/__test__/add-custom-rpc-model.test.ts
Tbaut: Thanks, ran a replace it across the whole code base :)

src/renderer/features/network/CustomRpcForm/model/__test__/edit-custom-rpc-model.test.ts
Tbaut: The thing is that it's not because the rule is set in the form, that the UI will correctly implement and show the error. To me the model can be perfectly implemented, but the UI could miss the displaying of the error, which is a pretty annoying UX issue, as in users won't know what's up.

src/renderer/features/network/CustomRpcForm/model/add-custom-rpc-model.ts
Tbaut: This is covered already, but the typing inferred isn't aware of it, I personaly prefer to have these 2 ifs, rather than a `!`. Also if I change the types of the effect, then I get another TS issue that the effect should extend the types of the source. Let me know what outcome you'd prefer.

src/renderer/features/network/CustomRpcForm/model/add-custom-rpc-model.ts
Tbaut: I've managed to remove the others, but in this case I'm not sure what you'd recommend, since this is a temporary state, to show the loading button. I created something like the following
```ts
const connectivityCheckStarted = createEvent<boolean>()
const $isValid = createStore(false)
```

which is arguably nicer, however this `connectivityCheckStarted` still needs to be called in the effect unless I miss something. I don't know a way to call an event with a `sample`, do you mean it should be done in a `fn`?

src/renderer/features/network/CustomRpcForm/model/add-custom-rpc-model.ts
Tbaut: I must miss something, but to me, there's no simple way to call an effect with a `sample`. I went through the code a bit, and the `target` of a `sample` is always either an effect, or a store. The only place I saw effects being called was using a `spread` together with a `fn`. To me, this makes it hard to read an to maintain. But if you insist I should do this, I'll just do it :shrug: I just tend to prefer readability, also because as a newcomer to `effector` I find it extremely hard to review with those many `sample`s.

src/renderer/features/network/CustomRpcForm/ui/EditCustomRpcModal.test.tsx
Tbaut: I think I wrote this above, but to me, these tests are important, because if users don't have errors displayed, they don't know what's going on. The RPC check for instance or the "node already exist" check are done upon submit and depending on the result some errors should be displayed or not. Maybe some tests aren't relevant, but some are in my opinion.

src/renderer/features/network/CustomRpcForm/ui/EditCustomRpcModal.tsx
Tbaut: Sure, I went for the store and no util. I know you're using a lot of utils across the code base, to me these are very repetitive since I find the enum already quite clear. lmk if you think I should add a util

src/renderer/features/network/CustomRpcForm/ui/EditCustomRpcModal.tsx
Tbaut: I didn't understand this comment. What do you mean exactly?

src/renderer/pages/Settings/Networks/components/CustomRpcModal/AddCustomRpcModal.tsx
Tbaut: you're right, it's not needed, I forgot to push my changes.

src/renderer/pages/Settings/Networks/Networks.tsx
Tbaut: Is this an issue though? for readability, I think it's nicer to have the `isXYZ` and `setXYZ` like this, but maybe that's too much of a React state management habit. Let me know if I should remove them from this `useUnit`

src/renderer/pages/Settings/Networks/components/NetworkSelector/NetworkSelector.tsx
Tbaut: Thanks I got it. I've moved all these as props, and called the events from the `Networks` then.

src/renderer/pages/Settings/Networks/components/index.ts
Tbaut: Indeed, this was a stray change that's not needed.

Tbaut:
Thank you so much for the review. I've gone through each of your comments, and either :+1: meaning that I have done the change, or eventually commented if I didn't understand or asked for clarification. It turns out the `blur`  for `onValidate` of `effector-form` doesn't work.. or at least not the way I expected. The tests are still failing because I didn't take a look at it again, I will do it carefully tomorrow and wanted to first focus on addressing your comments.

src/renderer/features/network/CustomRpcForm/model/add-custom-rpc-model.ts
Asmadek: `clock` and `target` in sample are universal and you can use any from `store`, `effect`, or `event` for it

src/renderer/features/network/CustomRpcForm/model/add-custom-rpc-model.ts
Asmadek: You can use effect as target in sample.

And for loading button you can use [pending property](https://effector.dev/en/api/effector/effect/#properties-pending) for `effect`. Other properties are useful too. `Done`, `donedata`, `fail` we use a lot

src/renderer/features/network/CustomRpcForm/model/add-custom-rpc-model.ts
Asmadek: You can use `network: ExtendedChain` type without null, and filter it in sample. 

Example
```typescript
sample({
    clock: getConfigFx.doneData,
    filter: (config: XcmConfig | null): config is XcmConfig => Boolean(config),
    target: calculateFinalConfigFx,
});
```

src/renderer/features/network/CustomRpcForm/model/add-custom-rpc-model.ts
Tbaut: with the `pending`, this makes total sense yes, thank you.

src/renderer/features/network/CustomRpcForm/model/add-custom-rpc-model.ts
tuul-wq: In other words "any kind of Unit" and Unit is store/event/effect 😉

src/renderer/features/network/CustomRpcForm/model/add-custom-rpc-model.ts
tuul-wq: `sample` is doing one thing and do it really well, so it's the best place to place filtering (with `filter`), data mapping (with `fn`).
And if we managed to reach the following `effect` (in `target`) it must receive all required data to proceed and not fast return on the first line, otherwise it's a red flag and better to step back and design more robust data flow.

Tbaut: Thanks for the pointers, after fighting a bit more, I ended up having TS happy. I'd say it's pretty hard to read though :sweat: 

```ts
sample({
  clock: $addCustomRpcForm.submit,
  source: { network: $selectedNetwork, url: $addCustomRpcForm.fields.url.$value },
  filter: (params: { network: ExtendedChain | null; url: string }): params is CheckRpcNodeFxParams => !!params.network,
  target: [checkRpcNodeFx, isNodeExistFx],
});
```

src/renderer/features/network/CustomRpcForm/model/add-custom-rpc-model.ts
Tbaut: But is something like this ok? This looks wrong to me :)

```ts
sample({
  clock: saveRpcNodeFx.done,
  fn: () => false,
  target: processStarted,
});
```

src/renderer/pages/Settings/Networks/Networks.tsx
tuul-wq: Yeah, better to remove event from `useUnit`
As more framework agnostic the code is, with more ease new features/bug fixes/ui changes would be added

src/renderer/features/network/CustomRpcForm/ui/EditCustomRpcModal.tsx
tuul-wq: It's about business rules in UI
As long as `isLoading` comes from the store (ideally) and it's boolean we're fine, not bothering ourselves how it's being calculated (business rules)
Once business rules leak to UI we have a problem
utils function that checks enums/states/or whatever not that complicated, is just a compromise and a way to avoid introducing yet another simple boolean store
The ideal solution is utils function with business rule + store that has it's value set via that utility function (+ tests for all utility functions)

src/renderer/features/network/CustomRpcForm/ui/EditCustomRpcModal.tsx
tuul-wq: I ment this [coment](https://github.com/novasamatech/nova-spektr/pull/1571#discussion_r1530229074)

src/renderer/features/network/ManageCustomRpcNode/lib/constants.ts
```diff
@@ -0,0 +1,27 @@
+import { RpcCheckResult } from './types';
+import { validateWsAddress } from './utils';
+import { RpcValidation } from '@shared/api/network';
+
+const RPC_NAME_MAX_LENGTH = 50;
+const RPC_NAME_MIN_LENGTH = 3;
+
+export const fieldRules = {
```
tuul-wq: We use PascalCase for object constants
`export const FieldRules`

src/renderer/features/network/CustomRpcForm/ui/EditCustomRpcModal.tsx
Tbaut: understood, thank you, I'll change it.

src/renderer/features/network/CustomRpcForm/ui/EditCustomRpcModal.tsx
Tbaut: Sounds good, I'll add a util then.

src/renderer/features/network/ManageCustomRpcNode/lib/utils.ts
tuul-wq, line 1: If you try to search specific utils file in the project you'll stumble upon many-many `utils.ts` files.
To avoid this and to emphasise features names we call utils files based on the feature it's located in.
For this file it would be something like `custom-rpc-utils.ts`
No need to keep every word from the feature name if it becomes inconveniently long like `manage-custom-rpc-node-utils.ts`

tuul-wq: Also to reduce amount of exported functions from utils we use this approach:
```javascript
export const xcmUtils = {
  getEstimatedWeight,
  getEstimatedFeeFromConfig,
  // other functions to export
};
```
This way functions must be declared via function declaration, not function expression

src/renderer/features/network/ManageCustomRpcNode/model/add-custom-rpc-model.ts
```diff
@@ -0,0 +1,178 @@
...
+    name: {
+      init: '',
+      rules: fieldRules.name,
+    },
+    url: {
+      init: '',
+      rules: fieldRules.url,
+    },
+  },
+  validateOn: ['change'],
```
tuul-wq: In all new forms (with effector) we move to `validateOn: ['submit']`

src/renderer/features/network/ManageCustomRpcNode/ui/AddCustomRpcModal.test.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+    const name = screen.getByPlaceholderText('settings.networks.namePlaceholder');
+    const address = screen.getByPlaceholderText('settings.networks.addressPlaceholder');
+    const submit = screen.getByRole('button', { name: 'settings.networks.addNodeButton' });
+    expect(name).toBeInTheDocument();
+    expect(address).toBeInTheDocument();
+    expect(submit).toBeInTheDocument();
+  });
+
+  test('should disable submit button during submission', async () => {
```
tuul-wq: UI tests must test only UI appearance, because UI is just a visual state of data.

Provide `isLoading: true` inside the `scope` and check disabled submit button
Same for `isValid` state - put errors/invalid form state into the `scope` and check the button
If you're trying (which you are) to invoke invalid UI state filling inputs and pressing buttons, then you're mimicking business logic of the model
The rule is to "provide required state - check UI"
This way all UI tests become really really dumb, which they should be indeed
That's why I suggest to just remove them at least in this point in time 😉

src/renderer/features/network/CustomRpcForm/model/add-custom-rpc-model.ts
tuul-wq: That's why using `!` is a valid compromise after you've check everything in filter
By the way new TS release is going to have [upgraded type guards](https://github.com/microsoft/TypeScript/pull/57465), so nasty types for filter will eventually go away (people in effector chat say so)

src/renderer/features/network/ManageCustomRpcNode/model/add-custom-rpc-model.ts
```diff
@@ -0,0 +1,178 @@
...
+const nodeExistChecked = createEvent<boolean>();
+const processStarted = createEvent<boolean>();
+
+const $selectedNetwork = createStore<ExtendedChain | null>(null);
+const $rpcConnectivityResult = createStore<RpcCheckResult>(RpcCheckResult.INIT);
+const $isNodeExist = createStore<boolean>(false);
+const $isProcessStarted = createStore<boolean>(false);
+const $isLoading = createStore<boolean>(false);
+
+const checkRpcNodeFx = createEffect(async ({ network, url }: CheckRpcNodeFxParams): Promise<RpcCheckResult> => {
```
tuul-wq: `check` is a common word, we want to be as specific as possible
inside the effect we fire validation, so why not to call it `validateRpcNodeFx`? 😉

src/renderer/features/network/ManageCustomRpcNode/model/add-custom-rpc-model.ts
```diff
@@ -0,0 +1,178 @@
...
+  const validationResult = await networkService.validateRpcNode(network.chainId, url);
+  const result = RpcValidationMapping[validationResult];
+
+  return result;
+});
```
tuul-wq: Better pass `chainId` instead of `network` if that's the only thing we need
Also renamed params/result types - a bit simpler naming
```suggestion
const validateRpcNodeFx = createEffect(async ({ chainId, url }: ValidateRpcParams): Promise<ValidateRpcResult> => {
  const validationResult = await networkService.validateRpcNode(chainId, url);
  
  return RpcValidationMapping[validationResult];
});
```

src/renderer/features/network/ManageCustomRpcNode/model/add-custom-rpc-model.ts
```diff
@@ -0,0 +1,178 @@
...
+
+const saveRpcNodeFx = createEffect(async ({ network, form }: SaveRpcNodeFxParams) => {
+  manageNetworkModel.events.rpcNodeAdded({
```
tuul-wq: These events (and everything related to them) must be moved here
Previously they were moved from the network model, as a first step of becoming a self-sufficient feature
Now time has come to bring them to the final destination 😎
```javascript
rpcNodeAdded
rpcNodeUpdated
rpcNodeRemoved
```

src/renderer/features/network/ManageCustomRpcNode/model/add-custom-rpc-model.ts
```diff
@@ -0,0 +1,178 @@
...
+    chainId: network?.chainId,
+    rpcNode: {
+      name: form.name,
+      url: form.url,
+    },
+  });
+});
+
+const isNodeExistFx = createEffect(({ network, url }: NodeExistParam): boolean => {
```
tuul-wq: `effect` is an action so verbs must be used in the first place
Here `check` is a nice candidate - `checkIsNodeExistFx`

tuul-wq: By the way this `effect` doesn't involve any kind of side effect
Every part of this code could be placed in `fn` for appropriate `sample`

src/renderer/features/network/ManageCustomRpcNode/ui/AddCustomRpcModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+    submit,
+    isValid,
+    isDirty,
+  } = useForm(addCustomRpcModel.$addCustomRpcForm);
+
+  const onSubmit = (event: FormEvent) => {
+    event.preventDefault();
+
+    submit();
+  };
```
tuul-wq:
```suggestion
  const onSubmit = (event: FormEvent) => {
    event.preventDefault();
    submit();
  };
```

src/renderer/features/network/ManageCustomRpcNode/ui/AddCustomRpcModal.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  useEffect(() => {
+    addCustomRpcModel.events.formInitiated();
+  }, []);
+
+  const {
+    fields: { name, url },
+    submit,
+    isValid,
+    isDirty,
```
tuul-wq: Do we really need `isDirty`?

src/renderer/features/network/index.ts
```diff
@@ -1,3 +1,5 @@
+export { addCustomRpcModel, AddCustomRpcModal, editCustomRpcModel, EditCustomRpcModal } from './ManageCustomRpcNode';
+export { RpcCheckResult } from './ManageCustomRpcNode/lib/types';
```
tuul-wq: Don't see any usage outside of Rpc feature 🤔

src/renderer/pages/Settings/Networks/Networks.tsx
```diff
@@ -195,10 +190,14 @@ export const Networks = () => {
             <InactiveNetwork networkItem={network}>
               <NetworkSelector
                 networkItem={network}
+                openEditRpcModal={() => editCustomRpcModel.events.processStarted(true)}
```
tuul-wq: Seems awkward that we need to pass a boolean flag to the event that claims process start but requires something besides void 🤔
`editCustomRpcModel.events.processStarted(true)` 

> Summary of 'Clean code' by Robert C. Martin
> Function rules
> 6. Don't use flag arguments. Split method into several independent methods that can be called from the client without the flag.

src/renderer/pages/Settings/Networks/Networks.tsx
```diff
@@ -150,27 +157,15 @@ export const Networks = () => {
...
   };
 
-  const changeCustomNode = (network: ExtendedChain) => {
-    return (node?: RpcNode) => {
-      setNodeToEdit(node);
-      setNetwork(network);
-
-      toggleCustomRpc();
-    };
+  const closeAddCustomRpcModal = () => {
```
tuul-wq: This might be changed by `useModalClose` but not sure
At least it incapsulates a delay value and provides `closeModal` function
We use it from time to time, at least worth checking 😉

src/renderer/pages/Settings/Networks/components/NetworkSelector/NetworkSelector.tsx
```diff
@@ -29,7 +30,11 @@ type Props = {
   onDisconnect: () => void;
   onConnect: (type: ConnectionType, node?: RpcNode) => void;
   onRemoveCustomNode: (node: RpcNode) => void;
-  onChangeCustomNode: (node?: RpcNode) => void;
+  onAddRpcNetworkChange: EventCallable<ExtendedChain>;
+  onEditRpcNetworkChange: EventCallable<ExtendedChain>;
```
tuul-wq:
```suggestion
  onAddRpcNetworkChange: (_required_param_) => void;
  onEditRpcNetworkChange: (_required_param_) => void;
```

src/renderer/features/network/CustomRpcForm/model/add-custom-rpc-model.ts
Tbaut: Answering my own question: It's probably not ok, the trick is to use `.reinit` on the resulting store, there's basically never a case in the code where a target is pointing at an event.

src/renderer/features/network/ManageCustomRpcNode/model/add-custom-rpc-model.ts
Tbaut: I made the change to pass the result directly, but making the change as you suggest with the `chainId` completely messes up with the sample I have in place, that combines different effects, such as the one with checking node existence, where more than the `chainId` is used.

```ts
sample({
  clock: $addCustomRpcForm.submit,
  source: { network: $selectedNetwork, url: $addCustomRpcForm.fields.url.$value },
  filter: (params: { network: ExtendedChain | null; url: string }): params is VerifyRpcConnectivityFxParams =>
    !!params.network,
  target: [verifyRpcConnectivityFx, isNodeExistFx],
});
```

I know you're saying we shouldn't be scared about duplicating, but given the filtering here, I think this is not worth it. Let me know.

src/renderer/features/network/ManageCustomRpcNode/model/add-custom-rpc-model.ts
Tbaut: Wow you're dragging me in a slippery slope here, expanding the scope of this PR quite a bit... 
I gave it a go, and moved the `rpcNodeAdded` and `rpcNodeUpdated` to see what this would entail. While this seems to be ok, as in I moved some functions, the UI works, but some side effect such as the following are needed and are still in the manage-network-model. Unless I move these too (not sure how tbh) then when you update a node or add one, you won't connect to it right away.

```ts
sample({
  clock: updateConnectionFx.doneData,
  source: networkModel.$providers,
  filter: (_, connection) => {
    return Boolean(connection) && networkUtils.isEnabledConnection(connection!);
  },
  fn: (providers, connection) => {
    const chainId = connection!.chainId;

    return providers[chainId] ? { reconnect: { chainId, providers } } : { start: chainId };
  },
  target: spread({
    start: networkModel.events.chainConnected,
    reconnect: reconnectProviderFx,
  }),
});
```

Anyway, I'll have to push back on this change. This will be for another PR, that I'm happy to take, after this one is merged first. I like to keep PRs as focused as possible and do baby steps to start with.

src/renderer/features/network/ManageCustomRpcNode/ui/AddCustomRpcModal.tsx
Tbaut: for me yes, so that the button is disabled unless users changed any field.

src/renderer/features/network/index.ts
Tbaut: Not with the utils any more indeed.

src/renderer/pages/Settings/Networks/Networks.tsx
Tbaut: Agreed, but this is the event that manages whether the modal is open or not. I had no better idea, maybe we could simply have a better name.

Tbaut:
Thanks for the reviews! I've not touched the tests yet, simply bc it's a nightmarish task for me still. I'll most probably ditch the UI ones as Yaroslav advocates :)

I should have addressed or commented back for every or your comments.

src/renderer/pages/Settings/Networks/Networks.tsx
Tbaut: This is something I wasn't sure about and I've asked earlier. The problem with managing the state of the modal in React (with `useModal` for instance), is that I need somehow to pass a callback to the effector model so that I close the modal e.g, when the new rpc node has been added/updated. So either I have:
- React state and I need to pass and `onClose` callback to the model
- Effector state and I need to play with some events in React to open/close the modal

At least that what I've been thinking about. Maybe there's something better, let me know.

tuul-wq: In "AddProxy" I use this:
```javascript
const [isModalOpen, closeModal] = useModalClose(!addProxyUtils.isNoneStep(step), addProxyModel.output.flowFinished);
```
But because I've got steps this is really handy, not sure about your situation

Tbaut: Oh I see, sure, I guess I can make it work with it too. Let me check.

src/renderer/pages/Settings/Networks/Networks.tsx
Tbaut: I just saw some of your code that was merged recently. I would have thought this isn't ok.. but if it is, then I can do such a construct here too and have a nice event name to call.

```ts
sample({
  clock: flowStarted,
  fn: () => Step.INIT,
  target: stepChanged,
});
```

Tbaut: changed it now, with things I originally thought were not ok, with sample and `fn: () => false`

src/renderer/pages/Settings/Networks/Networks.tsx
Tbaut: done

src/renderer/features/network/CustomRpcForm/ui/EditCustomRpcModal.test.tsx
Tbaut: aaand it's gone

src/renderer/features/network/CustomRpcForm/model/__test__/edit-custom-rpc-model.test.ts
Tbaut: There was a misunderstanding there, I wasn't finished with tests actually. This is fixed now

Tbaut:
Yay tests are now passing, and I believe that I've addressed (or commented) on all your comments.

src/renderer/features/network/ManageCustomRpcNode/model/edit-custom-rpc-model.ts
```diff
@@ -0,0 +1,175 @@
...
+  manageNetworkModel.events.rpcNodeUpdated({
+    chainId: network.chainId,
+    oldNode: nodeToEdit,
+    rpcNode: { url: form.url, name: form.name },
+  });
+});
+
+const updateInitialValuesFx = createEffect(({ url, name }: RpcNode) => {
+  $editCustomRpcForm.fields.url.onChange(url);
+  $editCustomRpcForm.fields.name.onChange(name);
```
Asmadek: I am not sure that use store directly in effect is good idea, better to provide it through params. 

@tuul-wq correct me if I'm wrong

Asmadek (approved):

Tbaut:
Can you take a look @tuul-wq please?

tuul-wq:
> Can you take a look @tuul-wq please?

Yeah, I even fetched the branch to my local machine to see the whole picture.
Going to provide commits from my side as soon as I'll be free from proxy staking
Don't worry about this PR it won't get lost 😉

src/renderer/features/network/ManageCustomRpcNode/lib/constants.ts
```diff
@@ -24,13 +24,8 @@ const FieldRules = {
...
+export const RpcValidationMapping = {
   [RpcValidation.INVALID]: RpcConnectivityResult.INVALID,
   [RpcValidation.VALID]: RpcConnectivityResult.VALID,
   [RpcValidation.WRONG_NETWORK]: RpcConnectivityResult.WRONG_NETWORK,
 };
-
-export const customRpcConstants = {
-  RpcValidationMapping,
-  FieldRules,
-};
```
Tbaut: What's the common practice here, don't we export an object for constants, unlike for other things like functions/effect..?

src/renderer/features/network/ManageCustomRpcNode/lib/custom-rpc-utils.ts
```diff
@@ -1,4 +1,4 @@
-import { RpcConnectivityResult } from './custom-rpc-types';
```
Tbaut: Similar question here. Is this a specific rule for types? I thought we need to give specific names, so as not to have many `types.ts`, just like we shouldn't have many `model.ts` etc.

src/renderer/features/network/ManageCustomRpcNode/model/add-custom-rpc-model.ts
```diff
@@ -1,171 +1,249 @@
...
-const flowStarted = createEvent();
-const flowFinished = createEvent();
+const $chainName = createStore<string | null>(null).reset(flowFinished);
+const $connection = createStore<Connection | null>(null).reset(flowFinished);
+const $existingNodes = createStore<RpcNode[]>([]).reset(flowFinished);
 
-const nodeExistVerified = createEvent<boolean>();
+const $provider = createStore<WsProvider | null>(null);
+const $rpcValidation = createStore<RpcValidation | null>(null).reset(flowFinished);
+const $isLoading = createStore<boolean>(false).reset(flowFinished);
```
Tbaut: I'm also surprised to see these `.reset` I thought we'd put as many things in `sample` as possible.

src/renderer/features/network/NetworkSelector/lib/network-selector-utils.ts
```diff
@@ -30,7 +29,3 @@ function getConnectionsList(networkItem: ExtendedChain): ConnectionItem[] {
 
   return [...actionNodes, ...nodesWithType, ...customs];
 }
-
-function canDeleteNode(nodeUrl: string, selectedConnectionUrl?: string): boolean {
```
tuul-wq, line 34: Active network can be removed with fallback to the 1st chain's default nodes

src/renderer/features/network/NetworkSelector/model/network-selector-model.ts
```diff
@@ -1,18 +1,20 @@
...
 
 const lightClientSelected = createEvent<ChainId>();
 const autoBalanceSelected = createEvent<ChainId>();
 const rpcNodeSelected = createEvent<{ chainId: ChainId; node: RpcNode }>();
-const chainDisabled = createEvent<ChainId>();
+const networkDisabled = createEvent<ChainId>();
 
-const updateConnectionFx = attach({ effect: networkModel.effects.updateConnectionFx });
-const disconnectProviderFx = attach({ effect: networkModel.effects.disconnectProviderFx });
+const $reconnectMap = createStore<Record<ChainId, boolean>>({});
```
tuul-wq, line 13: Map with chains that requires reconnect when changing from one active node to another

src/renderer/features/network/NetworkSelector/model/network-selector-model.ts
```diff
@@ -59,62 +61,72 @@ sample({
...
 });
 
 sample({
-  clock: [disconnectProviderFx.doneData, reconnectProviderFx.doneData],
-  source: networkModel.$providers,
-  fn: (providers, chainId) => {
-    const { [chainId]: _, ...rest } = providers;
-
-    return rest;
+  clock: networkModel.output.connectionStatusChanged,
```
tuul-wq, line 109: When connection status changed for a specific chain, we compare it with reconnect map and connect
Remove the chain from the map because it's handled

src/renderer/features/network/NetworkSelector/ui/NetworkSelector.tsx
```diff
@@ -95,43 +96,40 @@ export const NetworkSelector = ({
...
-                <Button
-                  size="sm"
-                  variant="text"
-                  className="w-full h-full justify-center"
-                  suffixElement={<Icon name="add" size={16} />}
-                >
-                  {t('settings.networks.addNodeButton')}
-                </Button>
-              </Listbox.Option>
+              <Button
```
tuul-wq, line 124: This way it's way easier to understand what callback we want to emit

src/renderer/pages/Settings/Networks/model/networks-overview-model.ts
```diff
@@ -39,6 +49,57 @@ sample({
   target: [activeNetworksModel.events.networksChanged, inactiveNetworksModel.events.networksChanged],
 });
 
+sample({
+  clock: networksFilterModel.$filteredNetworks,
+  target: [activeNetworksModel.events.networksChanged, inactiveNetworksModel.events.networksChanged],
+});
+
+sample({
```
tuul-wq, line 57: Following samples start node selection (reconnection) if we edit/remove active node or add node to the connected chain

src/renderer/features/network/NetworkSelector/model/network-selector-model.ts
Tbaut: > Don't forget to remove the chain from the map

is this a comment for me?

should we have this in the code for anyone reading this?

src/renderer/pages/Settings/Networks/model/networks-overview-model.ts
Tbaut: same here, I think it could be useful to have this comment in the code.

src/renderer/features/network/NetworkSelector/ui/NetworkSelector.tsx
```diff
@@ -95,43 +96,40 @@ export const NetworkSelector = ({
                           <FootnoteText className="text-text-secondary truncate">
                             {Title[type](t, node?.name)}
                           </FootnoteText>
-                          {node?.url && <HelpText className="text-text-tertiary truncate">{node.url}</HelpText>}
+                          <HelpText className="text-text-tertiary truncate">{node?.url}</HelpText>
```
sokolova-an, line 99: if there were no node or url, we can have 'undefined' in this HelpText block

src/renderer/features/network/NetworkSelector/model/network-selector-model.ts
tuul-wq: It's a part of my comment, nothing special
Rephrased a bit

src/renderer/features/network/NetworkSelector/ui/NetworkSelector.tsx
tuul-wq: `if (!children) return null;` inside `TextBase`

src/renderer/features/network/ManageCustomRpcNode/lib/constants.ts
tuul-wq: Right now constants related to a slice often don't leave the slice and not involved in public API
Utils and models - leave the slice through the API, so it's good to have them grouped with a meaningful name

src/renderer/features/network/ManageCustomRpcNode/model/add-custom-rpc-model.ts
tuul-wq: It's a nice shortcut, and we take it into account as soon as we discover stores in model file

src/renderer/pages/Settings/Networks/model/networks-overview-model.ts
tuul-wq: I would say you don't need comments if the code explicitly describes what it does or you're get used to read samples
1. read clock - `addCustomRpcModel.output.flowFinished`
2. read target - `networkSelectorModel.events.rpcNodeSelected`
Then if you need more details look into `filter / fn`
If code is messy or case is actually unique and requires human explanation - than yes, I up to leave a comment

Or even with such scheme of reading samples you'd better have comments?
It might end up having comments near every sample, which leads to questions: "is it a bad and unreadable code?" or "reading effector code is just a pure pain 😢" or ... "it's a matter of habit"
@Tbaut

tuul-wq (approved):

src/renderer/pages/Settings/Networks/model/networks-overview-model.ts
Tbaut: As someone is new to a code base, it's always easier to read plain text comments, I'm just thinking about any newcomer or outsider. Lowering the barrier to read code is often a good idea IMHO. While I agree that it's not absolutely needed for every sample, if you felt like it'd help me, or anyone reviewing this PR by adding this comment on Github, then I thought why not extend this help to any later reader.

tuul-wq: @sokolova-an suggests to create a document for contributors where we should describe vision of the project (not business value), how to interact with code in terms of FSD, how to read effector declarative operators an so on.
I'm strongly influenced by Robert Martin and his vision of a "clean code", so I'd better write a single document with guidelines instead of leaving comments here and there. And for sure continue code decoupling with the help of FSD.
My goal is to make code nice and tidy, comments are a good signal that code is too big to occupy just enough space in your brain.

PR #1603, opened 2024-03-27
fix: create proxy groups after pure proxied wallet created

Asmadek:
* fix: create proxy groups after pure proxied wallet created
* fix: pure proxy deposit calculation

### DIFF
--- a/src/renderer/features/proxies/model/proxies-model.ts
+++ b/src/renderer/features/proxies/model/proxies-model.ts
@@ -419,6 +419,6 @@ export const proxiesModel = {
   },
 
   output: {
-    proxiedWalletsCreationFinished: createProxiedWalletsFx.doneData,
+    walletsCreated: createProxiedWalletsFx.doneData,
   },
 };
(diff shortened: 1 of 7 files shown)


src/renderer/features/proxies/model/proxies-model.ts
```diff
@@ -417,4 +417,8 @@ export const proxiesModel = {
     workerStarted,
     proxiedWalletsCreated,
   },
+
+  output: {
+    proxiedWalletsCreationFinished: createProxiedWalletsFx.doneData,
```
tuul-wq: I would suggest a simpler name `walletsCreated`, we already know the domain - `proxies-model`
no need for such a specific name

src/renderer/widgets/AddPureProxiedModal/model/form-model.ts
```diff
@@ -412,71 +395,15 @@ sample({
   target: $proxyForm.fields.signatory.onChange,
 });
 
-sample({
```
tuul-wq, line 415: 👍

tuul-wq (approved):
Tiny comment

PR #1625, opened 2024-04-02
Feat: FSD refactor NetworkSelector

### LINKED ISSUE #1594: NetworkSelector should become a part of independent feature
Network screen is being decomposed into smaller features, which increases maintainability and testability.
`NetworkSelector` still holds business logic inside UI. When this task is completed `NetworkSelector` should be simple UI dropdown that invokes model's events, calls props for not feature related actions.
There are special conditions for LightClient which are resolved via `ConfirmModal`. They should be also addressed in this task.
Second part requires tinkering with design, I'll be glad to help if needed.

**Manage network feature** should consist of:
- [ ] plain UI `NetworkSelector` (without business logic)
- [ ] stateless `manage-network-model` that only handles connect/disconnect/change events
- [ ] call model's events in UI when trying to connect/disconnect/change
- [ ] Props for actions that do not belong to the feature: `onNewNode`, `onRemoveNode`, `onEditNode`

sokolova-an:
Move part of the logic from manage-network-modal to network-selector-modal
Split logic to utils 
Apply FSD for NetworkSelector component


src/renderer/features/network/NetworkSelector/lib/__tests__/network-selector-utils.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  test('should return correct connection options when canUseLightClient is true', () => {
+    const options = networkSelectorUtils.getConnectionOptions(defaultNetwork);
+    expect(options.availableNodes).toEqual([
+      { type: ConnectionType.AUTO_BALANCE },
+      { type: ConnectionType.DISABLED },
+      { type: ConnectionType.RPC_NODE, node: { url: 'wss://westend-rpc.polkadot.io', name: 'Parity node' } },
+      { type: ConnectionType.RPC_NODE, node: { url: 'wss://custom.io', name: 'node' } },
+    ]);
+    expect(options.selectedNode).toEqual({ type: ConnectionType.AUTO_BALANCE });
+    expect(options.isCustomNode('wss://custom.io')).toBe(true);
```
tuul-wq: Across the project we use `toEqual(...)` instead of `toBe(...)`

src/renderer/features/network/NetworkSelector/ui/NetworkSelector.test.tsx
tuul-wq, line 1: You can safely remove UI test for now, the only purpose for it to exist is to check correct rendering with specific state

src/renderer/features/network/NetworkSelector/ui/NetworkSelector.tsx
```diff
@@ -1,93 +1,60 @@
...
+const OptionsContainerStyle =
   'mt-1 absolute z-20 py-1 px-1 w-full border border-token-container-border rounded bg-input-background shadow-card-shadow';
 
 const TRANSITION_DURATION = 100;
 
-type SelectorPayload = {
-  type: ConnectionType;
-  title?: string;
-  node?: RpcNode;
+const Title = {
```
tuul-wq, line 22: 👍

src/renderer/features/network/NetworkSelector/ui/NetworkSelector.tsx
```diff
@@ -118,7 +86,7 @@ export const NetworkSelector = ({
             <div ref={ref} className={OptionsContainerStyle}>
               <Listbox.Options className="max-h-64 overflow-y-auto overscroll-contain">
                 {availableNodes.map((data) => {
```
tuul-wq: I suppose we can do this:
```javascript
 {availableNodes.map(({ type, node }) => {
```

sokolova-an: But we still use `data` itself for Listbox.Option value

src/renderer/features/network/NetworkSelector/lib/network-selector-utils.ts
```diff
@@ -0,0 +1,23 @@
...
+  const { canUseLightClient, customNodes } = connection;
+
+  const actionNodes = [
+    ...(canUseLightClient ? [{ type: ConnectionType.LIGHT_CLIENT }] : []),
+    { type: ConnectionType.AUTO_BALANCE },
+    { type: ConnectionType.DISABLED },
+  ];
+
+  const customs = (customNodes || []).map((node) => ({ node, type: ConnectionType.RPC_NODE, isCustom: true }));
+  const nodesWithType = nodes.map((node) => ({ node, type: ConnectionType.RPC_NODE, isCustom: false }));
```
tuul-wq: Even though I liked the initial variant the most, the suggested variant is a bit more readable 😉
```suggestion
  const customs = (customNodes || []).map((node) => ({
    node,
    type: ConnectionType.RPC_NODE,
    isCustom: true,
  }));
  const nodesWithType = nodes.map((node) => ({
    node,
    type: ConnectionType.RPC_NODE,
    isCustom: false,
  }));
```

src/renderer/features/network/NetworkSelector/model/__tests__/network-selector-model.test.ts
```diff
@@ -1,11 +1,11 @@
 import { fork, allSettled } from 'effector';
 
 import { networkModel } from '@entities/network';
-import { manageNetworkModel } from '../manage-network-model';
 import { ConnectionType, Connection, RpcNode } from '@shared/core';
 import { storageService } from '@shared/api/storage';
+import { networkSelectorModel } from '../network-selector-model';
 
-describe('pages/Settings/Networks/model/manage-network-model', () => {
+describe('features/network/NetworkSelector/model', () => {
```
tuul-wq:
```suggestion
describe('features/network/NetworkSelector/model/network-selector-model', () => {
```

src/renderer/features/network/NetworkSelector/model/network-selector-model.ts
```diff
@@ -0,0 +1,133 @@
+import { createEvent, sample, createEffect, attach } from 'effector';
+import { delay, spread } from 'patronum';
+
+import { ProviderWithMetadata } from '@/src/renderer/shared/api/network';
```
tuul-wq: I assume this import can be shortened

src/renderer/features/network/NetworkSelector/model/network-selector-model.ts
```diff
@@ -0,0 +1,133 @@
...
+import { networkModel, networkUtils } from '@entities/network';
+import { ChainId, ConnectionType, RpcNode } from '@shared/core';
+
+const lightClientSelected = createEvent<ChainId>();
+const autoBalanceSelected = createEvent<ChainId>();
+const rpcNodeSelected = createEvent<{ chainId: ChainId; node: RpcNode }>();
+const chainDisabled = createEvent<ChainId>();
+
+const updateConnectionFx = networkModel.effects.updateConnectionFx;
```
tuul-wq: If `networkModel.effects.updateConnectionFx` is never used anywhere else, it can be moved into this model
Otherwise you should use `attach` to wrap the initial effect
It's needed to have an isolated copy of the effect, so anybody who tracks original `updateConnectionFx` won't trigger accidentally. [doc](https://effector.dev/en/api/effector/attach/#methods-attach-effect-examples)

src/renderer/features/network/NetworkSelector/model/network-selector-model.ts
```diff
@@ -0,0 +1,133 @@
...
+  source: networkModel.$connections,
+  fn: (connections, { chainId, node }) => ({
+    ...connections[chainId],
+    connectionType: ConnectionType.RPC_NODE,
+    activeNode: node,
+  }),
+  target: updateConnectionFx,
+});
+
+sample({
```
tuul-wq, line 50: Looks like we need to set `activeNode: undefined` for disabled node

src/renderer/features/network/NetworkSelector/model/network-selector-model.ts
```diff
@@ -0,0 +1,133 @@
...
+  source: networkModel.$providers,
+  fn: (providers, chainId) => {
+    const { [chainId]: _, ...rest } = providers;
+
+    return rest;
+  },
+  target: networkModel.$providers,
+});
+
+delay({
```
tuul-wq, line 107: Oh, cannot remind myself why we introduced this delay...
Let it stay for a while, don't think it hurts someone
Hope to have time in future and reiterate this decision

src/renderer/features/network/NetworkSelector/ui/NetworkSelector.tsx
```diff
@@ -1,93 +1,60 @@
...
+    title: (t: TFunction) => t('settings.networks.selectorLightClient'),
+  },
+  [ConnectionType.RPC_NODE]: {
+    title: (t: TFunction, nodeName?: string) => nodeName,
+  },
 };
 
 type Props = {
-  networkItem: ExtendedChain;
+  nodesList: ConnectionList[];
```
tuul-wq: `List` already claims to be an array, according to types it should result in array of arrays 😉
I think `connectionList: ConnectionItem[]` conveys the meaning better
And `selectedConnection?: ConnectionItem;` as well
Could you remind me why it's optional? If connection is disabled `selectedConnection` should have disabled option

tuul-wq: All in all these props are brilliant, exactly what I wanted to see in simple UI component

src/renderer/features/network/NetworkSelector/ui/NetworkSelector.tsx
```diff
@@ -143,7 +113,7 @@ export const NetworkSelector = ({
                                 onChangeCustomNode(node);
                               }}
                             />
-                            {activeNode?.url !== node.url && (
+                            {selectedConnection?.node?.url !== node.url && (
```
tuul-wq: Let's add this comparison into utils `function canDeleteNode(...)`
So the UI won't bother comparing one url with another

tuul-wq (commented):
1st part of comments

src/renderer/pages/Settings/Networks/model/manage-network-model.ts
```diff
@@ -22,29 +14,11 @@ const rpcNodeAdded = createEvent<NodeEventParams>();
 const rpcNodeUpdated = createEvent<NodeEventParams>();
 const rpcNodeRemoved = createEvent<NodeEventParams>();
 
-const updateConnectionFx = createEffect((connection: Connection): Promise<Connection | undefined> => {
-  return storageService.connections.put(connection);
-});
+const updateConnectionFx = networkModel.effects.updateConnectionFx;
```
tuul-wq: Same comment about `attach`

src/renderer/pages/Settings/Networks/model/networks-overview-model.ts
```diff
@@ -1,8 +1,54 @@
...
+      Predicates[item.connection.connectionType]({ type: node.type, node: node.node }, item.connection.activeNode),
+    );
+
+    acc[item.chainId] = { nodes, selectedNode };
+
+    return acc;
+  }, {});
+});
+
+const $disabledNodeListMap = combine(inactiveNetworksModel.$inactiveNetworks, (list) => {
```
tuul-wq: Better to stick to $inactiveNetworks naming --> like `$inactiveConnectionsMap `
same for active list

src/renderer/pages/Settings/Networks/model/networks-overview-model.ts
```diff
@@ -1,8 +1,54 @@
...
+  networkSelectorUtils,
+} from '@features/network';
+import { ConnectionList } from '@/src/renderer/features/network/NetworkSelector';
 
 sample({
   clock: networksFilterModel.$filteredNetworks,
   target: [activeNetworksModel.events.networksChanged, inactiveNetworksModel.events.networksChanged],
 });
+
+const Predicates: Record<
```
tuul-wq: Nice to move it to `pages/Settings/Networks/lib/constants`

src/renderer/pages/Settings/Networks/model/networks-overview-model.ts
```diff
@@ -1,8 +1,54 @@
...
 
-import { networksFilterModel, activeNetworksModel, inactiveNetworksModel } from '@features/network';
+import { ChainId, ConnectionType, RpcNode } from '@shared/core';
+import {
+  networksFilterModel,
+  activeNetworksModel,
+  inactiveNetworksModel,
+  networkSelectorUtils,
+} from '@features/network';
+import { ConnectionList } from '@/src/renderer/features/network/NetworkSelector';
```
tuul-wq:
```suggestion
import type { XXX } from '@features/network/NetworkSelector';
```

tuul-wq (commented):
2nd part (final)

src/renderer/features/network/NetworkSelector/ui/NetworkSelector.tsx
sokolova-an: It's because of the typeScript - we use `nodes.find` so one of the possible outcomes is undefined. To avoid it we can add additional checks like `|| {disabled selectedConnection}` or if we absolutely sure that it should be a return result from our `find` function - we can do like this 
`const selectedNode = nodes.find((node) => node.type === item.connection.connectionType)!`
adding a `!` in the end will solve the issue with typeScript

tuul-wq: Alright, let's leave it as it, in the end `find` returning `undefined` is a totally valid result

src/renderer/features/network/NetworkSelector/lib/__tests__/network-selector-utils.test.ts
```diff
@@ -0,0 +1,83 @@
...
+        type: ConnectionType.RPC_NODE,
+        node: { url: 'wss://westend-rpc.polkadot.io', name: 'Parity node' },
+        isCustom: false,
+      },
+      { type: ConnectionType.RPC_NODE, node: { url: 'wss://custom.io', name: 'node' }, isCustom: true },
+    ];
+    expect(nodeList).toEqual(expected);
+  });
+
+  test('should return false for canDeleteNode if node is selected', () => {
```
tuul-wq, line 73: - Space between the tests
- Let's use variable to reduce nesting inside `expect` --> `const canDelete = networkSelectorUtils.canDeleteNode(...)`

src/renderer/features/network/NetworkSelector/model/__tests__/network-selector-model.test.ts
```diff
@@ -0,0 +1,82 @@
+import { fork, allSettled } from 'effector';
+
+import { networkModel } from '@entities/network';
+import { ConnectionType, Connection, RpcNode } from '@shared/core';
+import { storageService } from '@shared/api/storage';
+import { networkSelectorModel } from '../network-selector-model';
+
+describe('features/network/NetworkSelector/model/network-selector-model', () => {
+  const getMockConnection = (type: ConnectionType, activeNode?: RpcNode): Connection => ({
```
tuul-wq, line 9: 👍

src/renderer/pages/Settings/Networks/model/networks-overview-model.ts
```diff
@@ -1,8 +1,45 @@
...
+import type { ConnectionItem } from '@features/network/NetworkSelector';
+import {
+  networksFilterModel,
+  activeNetworksModel,
+  inactiveNetworksModel,
+  networkSelectorUtils,
+} from '@features/network';
+import { Predicates } from '../lib/constants';
 
 sample({
```
tuul-wq: Oh by the way, now we have computed stores, let's move `sample` down below as we do in other model
`events > stores > effects > combines > samples`

tuul-wq (approved):
Minor suggestions

src/renderer/entities/network/model/network-model.ts
```diff
@@ -97,6 +97,24 @@ const initConnectionsFx = createEffect((chains: Record<ChainId, Chain>) => {
...
+type DisconnectParams = {
+  chainId: ChainId;
+  providers: Record<ChainId, ProviderWithMetadata>;
+};
+const disconnectProviderFx = createEffect(async ({ chainId, providers }: DisconnectParams): Promise<ChainId> => {
+  await providers[chainId].disconnect();
+
+  providers[chainId].on('connected', () => undefined);
+  providers[chainId].on('disconnected', () => undefined);
+  providers[chainId].on('error', () => undefined);
```
Asmadek, line 113: meybe use `noop` function from lodash?

src/renderer/features/network/NetworkSelector/ui/NetworkSelector.tsx
```diff
@@ -1,93 +1,61 @@
...
 
 const TRANSITION_DURATION = 100;
 
-type SelectorPayload = {
-  type: ConnectionType;
-  title?: string;
-  node?: RpcNode;
+const Title = {
+  [ConnectionType.AUTO_BALANCE]: {
+    title: (t: TFunction) => t('settings.networks.selectorAutoBalance'),
```
Asmadek: Why not to use such approach? 

```ts
  [ConnectionType.AUTO_BALANCE]: (t: TFunction) => t('settings.networks.selectorAutoBalance'),
```

Asmadek (approved):

src/renderer/entities/network/model/network-model.ts
sokolova-an: I'd like to reduce lodash usage if it's possible

PR #1636, opened 2024-04-07
Feat: Bond proxy

tuul-wq:
- Proxy Bond and Nominate

Screenshot 2024-04-07 at 4 21 57 PM


src/renderer/entities/transaction/lib/transactionBuilder.ts
```diff
@@ -43,3 +49,114 @@ function buildTransfer({ chain, accountId, destination, asset, amount, xcmData }
...
+  const controller = destination ? toAddress(destination, { prefix: chain.addressPrefix }) : '';
+
+  return {
+    chainId: chain.chainId,
+    address: toAddress(accountId, { prefix: chain.addressPrefix }),
+    type: TransactionType.BOND,
+    args: {
+      value: formatAmount(amount, asset.precision),
+      controller,
+      payee: { Account: TEST_ADDRESS },
```
Asmadek, line 90: Are you sure about TEST_ADDRESS here? Maybe add comment why is it here

src/renderer/widgets/Staking/BondNominate/model/form-model.ts
```diff
@@ -0,0 +1,487 @@
...
+      return acc;
+    }, []);
+  },
+  target: $accountsBalances,
+});
+
+sample({
+  source: $accountsBalances,
+  fn: (accountsBalances) => {
+    if (accountsBalances.length === 0) return '0';
```
Asmadek: `ZERO_BALANCE` instead of '0'

Asmadek (approved):

PR #1658, opened 2024-04-09
fix: not save metadata if exists

Asmadek:
fix: remove metadata duplications
fix: not save metadata if exists

### DIFF
--- a/src/renderer/entities/network/model/network-model.ts
+++ b/src/renderer/entities/network/model/network-model.ts
@@ -384,6 +384,13 @@ sample({
 
 sample({
   clock: requestMetadataFx.doneData,
+  source: $metadata,
+  filter: (metadata, newMetadata) => {
+    return !metadata.some(
+      ({ chainId, version }) => chainId === newMetadata!.chainId && version === newMetadata!.version,
+    );
+  },
+  fn: (_, metadata) => metadata,
   target: saveMetadataFx,
 });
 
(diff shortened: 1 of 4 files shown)


pgolovkin (approved):

src/renderer/entities/network/model/network-model.ts
```diff
@@ -384,6 +384,13 @@ sample({
 
 sample({
   clock: requestMetadataFx.doneData,
+  source: $metadata,
+  filter: (metadata, newMetadata) => {
+    return !metadata.some(
```
tuul-wq: `newMetadata` can be null?
```javascript
sample({
  clock: requestMetadataFx.doneData,
  source: $metadata,
  filter: (metadata, newMetadata) => {
    if (!newMetadata) return false;

    return metadata.every(({ chainId, version }) => {
      return chainId !== newMetadata.chainId || version !== newMetadata!.version;
    });
  },
  fn: (_, metadata) => metadata,
  target: saveMetadataFx,
});
```

src/renderer/shared/api/storage/service/dexie.ts
```diff
@@ -69,6 +69,8 @@ class DexieStorage extends Dexie {
       metadata: '++id',
     });
 
+    this.version(22).upgrade(removeMetadataDuplications);
```
tuul-wq: What about deleting all metadata?
It updates from time to time anyway, and we need to decide how to remove old ones in future

tuul-wq (approved):

PR #1666, opened 2024-04-10
Fix: run proxy workers

### LINKED ISSUE #1604: Run proxy-worker when the user adds, removes the wallet or adds, removes account in the Polkadot Vault wallet
Run proxy-worker when the user adds, removes the wallet or adds, removes account in the Polkadot Vault wallet in order to update the list of delegated to you wallets and proxy accounts.

sokolova-an:
Run proxy worker on add/delete wallet or account and on add/delete delegated proxy

### DIFF
--- a/src/renderer/entities/proxy/model/proxy-model.ts
+++ b/src/renderer/entities/proxy/model/proxy-model.ts
@@ -1,4 +1,4 @@
-import { combine, createEffect, createEvent, createStore, sample } from 'effector';
+import { combine, createEffect, createEvent, createStore, merge, sample } from 'effector';
 import groupBy from 'lodash/groupBy';
 
 import { type ProxyAccount, AccountId, ProxyGroup, NoID, ID } from '@shared/core';
@@ -190,5 +190,8 @@ export const proxyModel = {
     proxyGroupsAdded,
     proxyGroupsUpdated,
     proxyGroupsRemoved,
+  },
+  output: {
+    proxyStatusUpdated: merge([addProxiesFx, removeProxiesFx]),
   },
 };
--- a/src/renderer/entities/proxy/model/proxy-model.ts
+++ b/src/renderer/entities/proxy/model/proxy-model.ts
@@ -192,6 +192,6 @@ export const proxyModel = {
     proxyGroupsRemoved,
   },
   output: {
-    proxyStatusUpdated: merge([addProxiesFx, removeProxiesFx]),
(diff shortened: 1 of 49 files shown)


src/renderer/entities/wallet/model/wallet-model.ts
```diff
@@ -264,4 +259,5 @@ export const walletModel = {
     walletRemovedSuccess: removeWalletFx.done,
     walletsRemoved,
   },
+  output: { workerDataStarted: merge([walletCreatedFx.doneData, multishardCreatedFx.doneData]) },
```
tuul-wq:
```suggestion
  output: {
    workerDataStarted: merge([walletCreatedFx.doneData, multishardCreatedFx.doneData]),
  },
```

src/renderer/widgets/CreateWallet/ui/CreateWalletProvider.tsx
```diff
@@ -38,7 +39,14 @@ export const CreateWalletProvider = () => {
 
   const props: ModalProps = {
     onClose: walletPairingModel.events.walletTypeCleared,
-    onComplete: walletProviderModel.events.completed,
+    onComplete: () => {
+      if (walletType !== WalletType.WATCH_ONLY) {
```
tuul-wq, line 43: Nice workaround 👍

src/renderer/entities/proxy/model/proxy-model.ts
```diff
@@ -191,4 +191,7 @@ export const proxyModel = {
     proxyGroupsUpdated,
     proxyGroupsRemoved,
   },
+  output: {
+    proxyStatusUpdated: merge([addProxiesFx, removeProxiesFx]),
```
tuul-wq: I see handling for adding wallets/accounts + remove accounts but nothing for removing the wallet
This part of code is trying to cover that case of wallet deletion?
Explicit worker call after wallet deletion causes problems?

sokolova-an: This one should cover a case when we add or remove a proxy from an existing wallet. Otherwise, if we don't run a proxy worker after that - the list of proxies won't change

tuul-wq: Actually when we add proxy through details we manually add it into the store/db
This was done to avoid running worker
Maybe it's bugged, I'll take a look

Asmadek (approved):

tuul-wq (approved):

PR #1708, opened 2024-04-19
Fix: Identities

### LINKED ISSUE #1707: Validator list is not displayed on Kusama
Validator list is not displayed on Kusama.

The scenario:
1. Open the Staking page
2. Select nominator
3. Click "Show Validators"

Expected result:
1. The list of selected validator is shown

Actual result:
1. Empty list is shown

Reproduced for start staking, change validators flows.

Image

tuul-wq:
- Fixed Identities for runtime > 1.4.0
- Increased performance of validators mapping

closes #1707

### DIFF
--- a/src/renderer/entities/staking/api/validatorsService.ts
+++ b/src/renderer/entities/staking/api/validatorsService.ts
@@ -59,15 +59,19 @@ async function getValidatorsStake_OLD(api: ApiPromise, era: EraIndex): Promise<V
   const data = await api.query.staking.erasStakersClipped.entries(era);
   const maxNominatorRewarded = getMaxNominatorRewarded(api);
 
-  return data.reduce((acc, [storageKey, type]) => {
+  return data.reduce<ValidatorMap>((acc, [storageKey, type]) => {
     const address = storageKey.args[1].toString();
-    const totalStake = type.total.toString();
-    const oversubscribed = type.others.length >= maxNominatorRewarded;
     const nominators = type.others.map((n) => ({ who: n.who.toString(), value: n.value.toString() }));
 
-    const payload = { totalStake, address, oversubscribed, nominators, ownStake: type.own.toString() };
+    acc[address] = {
+      address,
+      nominators,
+      totalStake: type.total.toString(),
+      oversubscribed: type.others.length >= maxNominatorRewarded,
+      ownStake: type.own.toString(),
+    } as Validator;
 
-    return { ...acc, [address]: payload };
+    return acc;
   }, {});
 }
 
@@ -85,10 +89,8 @@ async function getValidatorsStake(api: ApiPromise, era: EraIndex): Promise<Recor
 
     acc[address] = {
       address,
-      // @ts-ignore
       totalStake: type.value.total.toString(),
       oversubscribed: false,
-      // @ts-ignore
       ownStake: type.value.own.toString(),
       nominators: [],
     };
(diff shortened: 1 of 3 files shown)


src/renderer/entities/staking/api/validatorsService.ts
```diff
@@ -111,12 +113,16 @@ async function getValidatorsStake(api: ApiPromise, era: EraIndex): Promise<Recor
...
-  return data.reduce((acc, [storageKey, type]) => {
+  return data.reduce<ValidatorMap>((acc, [storageKey, type]) => {
     const address = storageKey.args[1].toString();
-    const commission = parseFloat(type.commission.toHuman() as string);
-    const payload = { address, commission, blocked: type.blocked.toHuman() };
 
-    return { ...acc, [address]: payload };
+    acc[address] = {
+      address,
+      commission: parseFloat(type.commission.toHuman() as string),
```
Asmadek, line 121: why not use `toString` and not use `as string`?

src/renderer/entities/staking/api/validatorsService.ts
```diff
@@ -222,22 +243,24 @@ async function getParentIdentities(
...
+    const { display, web, email, twitter } = data.info; // { data also includes: judgements, deposit }
 
-    const payload: Identity = {
+    acc[sub] = {
       subName,
       email: email.isRaw ? u8aToString(email.asRaw) : email.value.toString(),
-      riot: riot.isRaw ? u8aToString(riot.asRaw) : riot.value.toString(),
       twitter: twitter.isRaw ? u8aToString(twitter.asRaw) : twitter.value.toString(),
       website: web.isRaw ? u8aToString(web.asRaw) : web.value.toString(),
+      // riot: riot.isRaw ? u8aToString(riot.asRaw) : riot.value.toString(),
```
Asmadek: what reason to comment this code and not to delete ?

src/renderer/entities/staking/api/validatorsService.ts
```diff
@@ -307,13 +330,14 @@ async function getSlashingSpans(
...
   }
 
-  return slashingSpans.reduce((acc, span, index) => {
+  return slashingSpans.reduce<Record<Address, { slashed: boolean }>>((acc, span, index) => {
     let validatorIsSlashed = false;
-    if (!span.isNone) {
-      const { lastNonzeroSlash } = span.unwrap();
-      validatorIsSlashed = era - lastNonzeroSlash.toNumber() < slashDeferDuration;
+    if (span.isSome) {
+      validatorIsSlashed = era - span.unwrap().lastNonzeroSlash.toNumber() < slashDeferDuration;
```
Asmadek, line 335: what do you think to use BN here?

src/renderer/shared/core/types/identity.ts
```diff
@@ -5,8 +5,8 @@ export type Identity = {
   email: string;
   website: string;
   twitter: string;
-  riot: string;
   parent: ParentIdentity;
+  // riot: string;
```
Asmadek: same comment as for another commented code

Asmadek (approved):

sokolova-an (approved):

src/renderer/entities/staking/api/validatorsService.ts
tuul-wq: Hm, I agree, we can safely remove it, because we don't show riot/matrix anyway

Tbaut: why not keep it, and actually show it? (I know it's not here, could be in another PR) I find it useful at times for actual real time messaging, rather than the other platforms.

tuul-wq: > why not keep it, and actually show it? (I know it's not here, could be in another PR) I find it useful at times for actual real time messaging, rather than the other platforms.

In current design we don't show `riot/matrix` and don't have any plans to do so

PR #1710, opened 2024-04-20
chore: fix some typos in comments

overallteach:


Tbaut:
This user seems to be an airdrop farmer based on their github history. I've had such PRs in the past. They do typo fixing on web3 repositories to be seen as a contributor and later hopefully get airdrops. While typo fixing can help, this should IMHO not be encouraged. I'd suggest to close.

pgolovkin:
> This user seems to be an airdrop farmer based on their github history. I've had such PRs in the past. They do typo fixing on web3 repositories to be seen as a contributor and later hopefully get airdrops. While typo fixing can help, this should IMHO not be encouraged. I'd suggest to close.

Thanks!

PR #1717, opened 2024-04-22
Improve proxy-worker performance

stepanLav:
That PR change logic for proxy-worker in network interaction. That use [entries](https://polkadot.js.org/docs/api/start/api.query.other#map-keys--entries) to receive data from the network.

Found a problem with number of messages, sent by worker's wss connection:

It's an initialisation of one worker - 1711 storage queries:
Screenshot 2024-04-22 at 17 20 10

The same worker with entries - 16 requests, by 1000 entities and pagination 
Screenshot 2024-04-22 at 17 38 52

<details>
  <summary>Some measures</summary>

approach with keys:
dev

approach with entries:
entries

</details>

### DIFF
--- a/src/renderer/features/proxies/workers/__tests__/proxy-worker.test.ts
+++ b/src/renderer/features/proxies/workers/__tests__/proxy-worker.test.ts
@@ -121,28 +121,26 @@ describe('features/proxies/workers/proxy-worker', () => {
       proxyType: 'Governance',
     };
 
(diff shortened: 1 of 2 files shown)


valentunn (approved):
`Proxy.proxies` no longer contain corrupted entry so its safe to use `entries` now

tuul-wq (approved):

PR #1725, opened 2024-04-24
Feat: Wallet accounts

### LINKED ISSUE #1529: Integrate accounts into wallet inside walletModel
Currently we save `wallets` & `accounts` into separate stores in `walletModel`.
But every time we need to get wallet's accounts we iterate over all accounts searching for a match with `walletId`.
Moreover type guards for wallet type doesn't provide appropriate type for wallet's accounts, thus we have to check their type as well which definitely should be unnecessary.

To greatly reduce the amount of work we have to do to map accounts to the specific wallet,
I suggest to map `accounts` into `wallet`.
```javascript
const wallet: Wallet = {
  // wallet properties
  accounts: [ ... ],
};
```

- [ ] update `Wallet` types
- [ ] update `walletModel`
- [ ] update code that relies on wallet & it's accounts
- [ ] update wallet/accounts/permission utils

tuul-wq:
- Put Accounts inside Wallet to have an easier access
- Updated Wallet/Account types
- Updated Wallet/Account type guards
- Implemented methods to iterate over all wallets and all accounts
- Removed MatrixContext

Three utility functions to conveniently find wallet(-s) with specific accounts:
`getAccountsBy`
`getWalletFilteredAccounts`
`getWalletsFilteredAccounts`

closes #1529


Tbaut:
Any reason to remove the Matrix context? This is one of the tasks of https://github.com/novasamatech/nova-spektr/pull/1729

tuul-wq:
> Any reason to remove the Matrix context? This is one of the tasks of #1729

Removed it just not to spend time fixing errors that are going to be removed by you anyway 😎

pgolovkin (approved):

pgolovkin (approved):

src/renderer/pages/Operations/components/modals/RejectTx.tsx
```diff
@@ -45,9 +44,9 @@ const AllSteps = [Step.CONFIRMATION, Step.SIGNING, Step.SUBMIT];
 
 const RejectTx = ({ tx, account, connection }: Props) => {
   const { t } = useI18n();
-  const accounts = useUnit(walletModel.$accounts);
-  const wallets = keyBy(useUnit(walletModel.$wallets), 'id');
+  // const wallets = keyBy(useUnit(walletModel.$wallets), 'id');
```
Asmadek: do you need it?

src/renderer/pages/Staking/ui/Actions.tsx
```diff
@@ -118,33 +70,33 @@ export const Actions = ({ canInteract, stakes, isStakingLoading, onNavigate }: P
...
 
-  const onClickAction = (action: StakeActions, path: PathType) => {
-    const incorrectType = getIncorrectAccountType(action);
+  const onClickAction = (operation: Operations, path: Operations) => {
+    const incorrectType = getIncorrectAccountType(operation);
 
     if (incorrectType) {
-      setActionType(action);
-      setActionPath(path);
+      // setOperationType(operation);
```
Asmadek: Do you need it?

src/renderer/entities/wallet/lib/account-utils.ts
```diff
@@ -151,53 +160,45 @@ function getBaseAccount(accounts: Account[], walletId?: ID): BaseAccount | undef
...
 }
 
 function isNonBaseVaultAccount(account: Account, wallet: Wallet): boolean {
   return !walletUtils.isPolkadotVault(wallet) || !accountUtils.isBaseAccount(account);
 }
 
-function isEthereumBased({ chainType }: Account): boolean {
-  return chainType === ChainType.ETHEREUM;
-}
+// function getPvAccounts(wallet: PolkadotVaultWallet): Array<ChainAccount | ShardAccount> {
```
Asmadek: Maybe add comment for this functions when it should be uncommented or remove it

Asmadek (approved):

pgolovkin (changes requested):
This PR can't be merged because of Matrix context removal. We can't remove Matrix before Multisig Discovery with indexer won't be implemented.

pgolovkin:
> > Any reason to remove the Matrix context? This is one of the tasks of #1729
> 
> Removed it just not to spend time fixing errors that are going to be removed by you anyway 😎

In that case how to release features before new Multisig discovery approach with indexer?

src/renderer/entities/signatory/lib/utils.ts
```diff
@@ -1,25 +1,17 @@
...
-
-    if (!accountIdMatch || !wallet) return;
+function getSignatoryWallet(wallets: Wallet[], accountId: AccountId): Wallet | undefined {
+  return wallets.find((wallet) => {
+    const hasMatch = wallet.accounts.some((account) => account.accountId === accountId);
 
     return (
-      walletUtils.isValidSignatory(wallet) || walletUtils.isPolkadotVault(wallet) || walletUtils.isMultiShard(wallet)
+      hasMatch &&
+      (walletUtils.isValidSignatory(wallet) || walletUtils.isPolkadotVault(wallet) || walletUtils.isMultiShard(wallet))
```
Tbaut: Random question as I read this code and ask myself. 
Why isn't `walletUtils.isValidSignatory(wallet)` sufficient here, shouldn't it take into account the `isPolkadotVault` and `isMultiShard`? Maybe @Asmadek you know.

From the function name `getSignatoryWallet` and then `isValidSignatory`, I wouldn't expect to have to check for anything else.

src/renderer/entities/wallet/lib/wallet-utils.ts
```diff
@@ -83,3 +93,63 @@ function isValidSignatory(wallet?: Pick<Wallet, 'type'>): boolean {
...
+function getWalletFilteredAccounts(
+  wallets: Wallet[],
+  predicates: {
+    walletFn?: (wallet: Wallet) => boolean;
+    accountFn?: (account: Account, wallet: Wallet) => boolean;
+  },
+): Wallet | undefined {
+  if (!predicates.walletFn && !predicates.accountFn) return undefined;
+
+  return wallets.reduce<Wallet | undefined>((acc, wallet) => {
```
Tbaut, line 138: Could we use `.find` instead of `reduce` to exit early and make this function a little easier to read?

I found it hard to understand what it does, some comments would be very welcome. I'd eventually rename it to `findWallet...`, what do you think?

pgolovkin (approved):

src/renderer/entities/wallet/lib/wallet-utils.ts
```diff
@@ -118,21 +119,20 @@ function getWalletFilteredAccounts(
 ): Wallet | undefined {
   if (!predicates.walletFn && !predicates.accountFn) return undefined;
 
-  return wallets.reduce<Wallet | undefined>((acc, wallet) => {
-    if (acc) return acc;
+  const wallet = wallets.find((wallet) => {
+    return !predicates.walletFn || predicates.walletFn(wallet);
```
Tbaut: I think `!predicates.walletFn` could be checked before, and we shouldn't even do the `.find` in case there's no wallet predicate. Something like

```ts
let walletsToSearch = wallets

if(predicates.walletFn){
  walletsToSearch = wallets.find((wallet) => predicates.walletFn(wallet)) || wallets
}
```

Asmadek: I am not sure it's needed for wallet, because we need only one wallet, and if walletFn is empty with `find` we get only the first wallet and calculation will be stoped. But I implement such approach for accounts where we use `filter`

Tbaut: > I am not sure it's needed for wallet, because we need only one wallet, and if walletFn is empty with find we get only the first wallet and calculation will be stoped.

Ah, that's why I think we need a comment for this function, since it's not clear what we really want to do.
From my understanding of this function originally, you would call it with a `walletFn` and `accountFn` and would get back a wallet that is truthy when you run both functions. Now, if there's no `walletFn` you would expect the function to return a wallet and account that fullfills only the `accountFn`. However, with this implementation, as you said, the first wallet would be selected, and then, if the `accountFn` isn't truthy for any account of this first wallet, the function will return `undefined` although there might be other wallets with an account that would match.

Said differently if I have the following
```ts
const wallets = [
  { 
    id: 1,
    accounts: [
      { accountId: 1 }
    ]
  },
  {
    id: 2,
    accounts: [
    { accountId : 2 }
    ]
  }
]
```

and you call `getWalletFilteredAccounts(wallets, {accountFn: (a) => a.accountId === 2})` then your function will return `undefiened` although it should have returned `{ id: 2, accounts: [ { accountId : 2 } ] }` isn't it?

Tbaut (approved):
Thanks, I guess this is safer with the comments, although hard to review. I don't want to hold this PR any longer.

Issue #1722, opened 2024-04-23
Portfolio view. Create portfolio view


pgolovkin:
1. Toggle for `Chain-centric` and `Token-centric` views
2. Implement `Token-centric` view itself

stepanLav:
Status verification: ⏳

What has been verified:
- Token-centric view
https://github.com/novasamatech/nova-spektr/issues/1834

Found during testing:
- https://github.com/novasamatech/nova-spektr/issues/1844

stepanLav:
Status verification: ✅

PR #1726, opened 2024-04-24
feat: change asset page structure

### LINKED ISSUE #1734: Portfolio view. Add view selector
Toggle for `Chain-centric` and `Token-centric` views

Part of  [1722](https://github.com/novasamatech/nova-spektr/issues/1722)

sokolova-an:
Change asset page structure in terms of preparation for portfolio view 

Toggle component to swish view will be in the asset page components as well as new components for send\receive;
The portfolio view will have the same structure as assetChainView;
Part of the asset modal moved to entities since we need to use store like `hideZeroBalance` in both views and in asset page


src/renderer/entities/asset/model/assets-model.ts
tuul-wq, line 1: Every store here is a part of a separate feature that can be easily extracted from the current Assets page
feature 1 - Hide zero balances
feature 2 - Shards selector (modal + button with shards amount)
feature 3 - Filter assets
And all the data they provide must be used on assets page for both views
Networks page is a good example: search & 2 lists being connected through `Networks.tsx`

src/renderer/shared/api/network/service/chainsService.ts
```diff
@@ -162,3 +162,24 @@ function sortChainsByBalance(
     sortBy(testnets.noBalance, 'name'),
   );
 }
+
+const sumValues = (firstValue?: string, secondValue?: string): string => {
```
tuul-wq, line 166: I find it pretty weird that both sum functions moved down in `network/chainsService`
Will they ever be used anywhere else besides Assets page?
If not then `pages/Assets/*-utils` is still a good place or the new one `AssetsChainView` 😉

src/renderer/features/assets/AssetsChainView/ui/NetworkFiatBalance/NetworkFiatBalance.tsx
tuul-wq, line 1: Don't hesitate to make folder structure more flat without additional folder nesting

src/renderer/pages/Assets/Assets/model/assets-view-model.ts
tuul-wq, line 1: Why not the old naming `assets-model`?

src/renderer/entities/asset/model/assets-model.ts
sokolova-an: But then we will need to use this data in other features like we need `hideZeroBalance` in pages to change it there and in both view (chain and portfolio) features. I moved it to entities to easily export it in features

src/renderer/shared/api/network/service/chainsService.ts
sokolova-an: I moved it there because we use `sumBalance` in chainsService and `sumValues` is a helper for `sumBalance`

src/renderer/pages/Assets/Assets/model/assets-view-model.ts
sokolova-an: I believe that `assets-model` is something more common like for entities while `assets-view-model` is responsible for the page and page view logic

Tbaut:
Can you link the original issue this is tackling please, e.g with `closes #[issue number]`, so that it gives more context?

tuul-wq:
@Tbaut it's a part of #1722

src/renderer/features/assets/AssetsSettings/model/assets-settings-modal.ts
```diff
@@ -0,0 +1,66 @@
+import { createEffect, createEvent, createStore, sample } from 'effector';
+
+import { localStorageService } from '@shared/api/local-storage';
+import { AssetsPageView } from '@entities/asset';
+import { ASSETS_PAGE_VIEW, HIDE_ZERO_BALANCES } from '../lib/constants';
+
+const $hideZeroBalances = createStore<boolean>(false);
+const $assetsView = createStore(AssetsPageView.TOKEN_CENTRIC);
```
tuul-wq:
```suggestion
const $assetsView = createStore<AssetsPageView>(AssetsPageView.TOKEN_CENTRIC);
```

src/renderer/features/assets/AssetsSettings/model/assets-settings-modal.ts
```diff
@@ -0,0 +1,66 @@
...
+
+const hideZeroBalancesChanged = createEvent<boolean>();
```
tuul-wq, line 7: Event > store > effect > combine > sample

src/renderer/features/assets/AssetsSettings/model/assets-settings-modal.ts
```diff
@@ -0,0 +1,66 @@
...
+const assetsViewChanged = createEvent<AssetsPageView>();
+const assetsStarted = createEvent();
+
+const getAssetsViewFx = createEffect((): AssetsPageView => {
+  return localStorageService.getFromStorage(ASSETS_PAGE_VIEW, AssetsPageView.TOKEN_CENTRIC);
+});
+
+const saveAssetsViewsFx = createEffect((value: AssetsPageView): AssetsPageView => {
```
tuul-wq:
```suggestion
const saveAssetsViewFx = createEffect((value: AssetsPageView): AssetsPageView => {
```

src/renderer/features/assets/AssetsSettings/model/assets-settings-modal.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export const assetsSettingsModel = {
+  $assetsView,
+  $hideZeroBalances,
+  events: {
+    hideZeroBalancesChanged,
+    assetsViewChanged,
+    assetsStarted,
+  },
+  outputs: {
```
tuul-wq: Right now feature consumer (in our case it's a page) can simply observe `$assetsView / $hideZB`
In more complicated cases where we need granular data to be observed `outputs` are a good choice
Or if we rely on some async action to be finished with specific data

src/renderer/features/assets/AssetsSettings/ui/AssetsSettings.tsx
```diff
@@ -0,0 +1,60 @@
...
+      element: <FootnoteText>{t('balances.tokenCentric')}</FootnoteText>,
+    },
+    {
+      id: AssetsPageView.CHAIN_CENTRIC.toString(),
+      value: AssetsPageView.CHAIN_CENTRIC,
+      element: <FootnoteText>{t('balances.chainCentric')}</FootnoteText>,
+    },
+  ];
+
+  const hideZeroBalances = useUnit(assetsSettingsModel.$hideZeroBalances);
```
tuul-wq: Grouping `useUnit` in one place is a good practice 😌

src/renderer/features/assets/AssetsSettings/ui/AssetsSettings.tsx
```diff
@@ -0,0 +1,60 @@
...
+            {t('balances.hideZeroBalancesLabel')}
+          </Switch>
+          <hr className="border-divider -mx-3 my-4" />
+          <Select
+            label={t('balances.pageView')}
+            selectedId={assetsView.toString()}
+            placeholder={t('settings.networks.selectorPlaceholder')}
+            options={options}
+            onChange={({ value }) => assetsSettingsModel.events.assetsViewChanged(value)}
+          />{' '}
```
tuul-wq: Serves some purpose?
```suggestion
          />{' '}
```

src/renderer/features/assets/index.ts
```diff
@@ -1 +1,4 @@
 export { AssetRouteGuard } from './AssetRouteGuard';
+export { AssetsChainView } from './AssetsChainView/ui/AssetsChainView';
```
tuul-wq: don't forget to import from `index.ts`

src/renderer/pages/Assets/Assets/Assets.tsx
tuul-wq, line 1: Nice 👍

src/renderer/entities/asset/lib/constants.ts
```diff
@@ -0,0 +1,4 @@
+export const enum AssetsPageView {
```
tuul-wq: It's even named `AssetsPage` it can't reside in `entities` 😉
I suggest to move into AssetsSettings (where it's actually used), rename it in order to make assets settings feature more versatile and not tied with Assets page that much (through the name)
Name options: `AssetsListView / AssetsView / AssetsGroup / AssetViewMode / AssetsDisplay`

tuul-wq: Also, we store const enums in `types`
If you're used to plain objects instead of const enums, please let me know
I'm eager to replace them but stumble upon appropriate typing 

(Paths object + type is actually made this way, but I don't like holding type in constant...)
`export type PathType = ObjectValues<typeof Paths>;`
And use it like this:
```javascript
const x: Record<PathType, string> = {
  [Paths.ROOT]: '1223',
};
console.log(x);
```

src/renderer/features/assets/AssetsSettings/model/assets-settings-modal.ts
sokolova-an: it was unused data, deleted it

src/renderer/entities/asset/lib/constants.ts
sokolova-an: I suppose I will need it later in other features, for example, to check in the modal if I should render the page based on `assetsView` param

Also, I saw in the project in some places an enum in constants files, is it a mistake then? Is it correct to place it in types if we use enum as values?

sokolova-an: I guess when we don't care about keys - an enum is an optimal solution, otherwise a plain object might be better but then we have to deal with types, which is not always convenient.

tuul-wq: > I suppose I will need it later in other features, for example, to check in the modal if I should render the page based on `assetsView` param
> 
> Also, I saw in the project in some places an enum in constants files, is it a mistake then? Is it correct to place it in types if we use enum as values?

That might be outdated, need to look into those files one by one

tuul-wq (approved):

PR #1740, opened 2024-05-02
fix: back to form for add proxy

### LINKED ISSUE #1712: Create proxy. Multisig. Back button lead to wrong widget state
## Description
When user go back with "back" button in multisig delegated authority case it lead to broken widget state.

<details>
  <summary>🍿 Example 🍿</summary>

https://github.com/novasamatech/nova-spektr/assets/40560660/678c6a68-810d-4e86-8a1a-4b16a1032cd6

</details>

## Steps to Reproduce

1. Go to multisig wallet details
2. Go to delegated authority
3. Click "Add delegated authority"
4. Fill all necessary fields
5. Go to Confirmation screen
6. Click "Back"

## Expected Behavior
Return to previous screen with all multisig data

## Actual Behavior
Will return to the broken widget state, where can go further

## Environment
https://github.com/novasamatech/nova-spektr/commit/34bc8d02b705548c958d21b2a6fd32d2ecb3a151

Asmadek:
* don't reset form if go back to form

### DIFF
--- a/src/renderer/widgets/AddProxyModal/model/form-model.ts
+++ b/src/renderer/widgets/AddProxyModal/model/form-model.ts
@@ -73,20 +73,20 @@ const feeChanged = createEvent<string>();
 const isFeeLoadingChanged = createEvent<boolean>();
 const isProxyDepositLoadingChanged = createEvent<boolean>();
 
-const $oldProxyDeposit = createStore<string>('0').reset(formSubmitted);
+const $oldProxyDeposit = createStore<string>('0');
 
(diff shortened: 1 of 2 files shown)


sokolova-an (approved):

Tbaut (commented):
Shouldn't it be reset on `formInitiated` though?

Asmadek:
> Shouldn't it be reset on `formInitiated` though?

@Tbaut We are not reset such stores in other flows, and I decided to do it in the same way

Tbaut:
ok, I was worried the data would be here still if you:
- submit
- launch the "add proxy" flow again.

PR #1744, opened 2024-05-03
feat: portfolio page

### LINKED ISSUE #1722: Portfolio view. Create portfolio view

### LINKED ISSUE #1792: Portfolio view. Sorting and filtering
Tokens sorting rules:
1. DOT, KSM at the top
2. Production tokens sorted by fiat balance `DESC`
3. Production tokens sorted by balance `DESC` if no fiat balance available
4. Test tokens sorted by balance `DESC`

Chains sorting rules:
1. Polkadot, Kusama at the top
2. Other networks sorted by balance `DESC` if balance is available
3. Other networks sorted by name `DESC`

sokolova-an:
https://github.com/novasamatech/nova-spektr/issues/1722
- Ui for Portfolio view page;
- Filtration based on active wallet, connected networks and query;
- Added balances;
 - Hide zero balance logic;
 - Sorting by balance;


pgolovkin:
@sokolova-an a few comments:
1. Wrong empty page state. I see table colums titles (screen is attached)
2. When I switch to token-centric view I alway get empty page 
image

src/renderer/features/assets/AssetsPortfolio/model/portfolio-model.ts
```diff
@@ -0,0 +1,174 @@
...
+
+const updateTokensFx = createEffect(
+  ({ activeWallet, chains }: { activeWallet?: Wallet; chains: Record<ChainId, Chain> }): TokenAsset[] => {
+    const tokens = tokensService.getTokensData();
+
+    return tokens.reduce((acc, token) => {
+      const filteredChains = token.chains.filter((chain) => {
+        return activeWallet?.accounts.some((account) => {
+          return (
+            accountUtils.isNonBaseVaultAccount(account, activeWallet) &&
```
Tbaut: Can you tell me why we filter these out? It's not immediately intuitive to me.

src/renderer/features/assets/AssetsPortfolio/model/portfolio-model.ts
```diff
@@ -0,0 +1,174 @@
...
+);
+
+const populateTokensBalanceFx = createEffect(
+  ({
+    activeTokens,
+    balances,
+    accounts,
+  }: {
+    activeTokens: TokenAsset[];
+    balances: any;
```
Tbaut: Is there a reason why this can't be typed `Balance[]`? I tried locally and seemed fine.

src/renderer/features/assets/AssetsPortfolio/model/portfolio-model.ts
```diff
@@ -0,0 +1,174 @@
...
+import { once } from 'patronum';
+
+import { Account, AccountId, Balance, Chain, ChainId, TokenAsset, TokenBalance, Wallet } from '@shared/core';
+import { networkModel, networkUtils } from '@entities/network';
+import { accountUtils, walletModel, walletUtils } from '@entities/wallet';
+import { AssetsListView } from '@entities/asset';
+import { balanceModel, balanceUtils } from '@entities/balance';
+import { sumTokenBalances, tokensService } from '../lib/tokensService';
+
+const setActiveView = createEvent<AssetsListView>();
```
Tbaut: I personally don't mind, but I thought we usually name the event in the past form for events. It would be `activeViewSet` and `accountsSet` in that case.

src/renderer/features/assets/AssetsPortfolio/model/portfolio-model.ts
```diff
@@ -0,0 +1,174 @@
...
+    activeWallet: walletModel.$activeWallet,
+    connections: networkModel.$connections,
+    chains: networkModel.$chains,
+    tokens: $tokens,
+  },
+  filter: ({ connections, activeWallet, activeView }) => {
+    return Boolean(activeView === AssetsListView.TOKEN_CENTRIC && Object.keys(connections).length && activeWallet);
+  },
+  fn: ({ connections, chains, tokens, activeWallet }): TokenAsset[] => {
+    const isMultisig = walletUtils.isMultisig(activeWallet);
```
Tbaut: I'm not sure I understand why we filter out multisigs here, and below with the `isMultisigSupported`? It seems weird to me that we filter out such chains, there are many.

Tbaut (commented):
Some small comments as I went through the code :)

src/renderer/features/assets/AssetsPortfolio/model/portfolio-model.ts
sokolova-an: good point

src/renderer/features/assets/AssetsPortfolio/model/portfolio-model.ts
sokolova-an: I took that part from the current chain filtration on the main page, I believe there was a point to add there in the first place 
Can't tell why it was added tbh

Tbaut: maybe @tuul-wq  or @Asmadek know?

src/renderer/features/assets/AssetsPortfolio/model/portfolio-model.ts
sokolova-an: Same here, took it from the original filtration don't want to break something in an unexpected way
Maybe @tuul-wq could tell - is it still necessary or if it's a rudiment

src/renderer/features/assets/AssetsPortfolio/model/portfolio-model.ts
pgolovkin: @Tbaut It's made in order to process the following case:
1. The user selects multisig wallet
2. The user opens Portfolio page
3. Nova Spektr shouldn't show chains where multisig pallet doesn't exist because it's redundant.

Tbaut: But this is not what this line does from what I understand if we look at the lines below:
```ts
        const hasMultiPallet = !isMultisig || networkUtils.isMultisigSupported(chains[c.chainId].options);

        return !isDisabled && hasMultiPallet;
```

I guess this `hasMultiPallet` is an unlucky name because it's mixing 2 things, the fact that the account is **not** a multisig and the fact that the chain has the multisig pallet. and it's not filtering out the `isMultisig && hasPallet`. I find this very confusing.

src/renderer/features/assets/AssetsPortfolio/model/portfolio-model.ts
pgolovkin: @Tbaut it's all about multishard and Polkadot Vault with dynamic derivations.
Multishard Polkadot Vault wallet or Polkadot Vault wallet with Dynamic derivations contains accounts that are linked to the chain (like in Polkadot Vault mobile app). So the token has to be shows only for the specific account on the specific chain.

Tbaut: thanks, I keep forgetting that the "Base" things have to do with multishard.

src/renderer/pages/Assets/Assets/model/assets-model.ts
```diff
@@ -1,12 +1,24 @@
...
 const activeShardsSet = createEvent<Account[]>();
 
 const $activeShards = createStore<Account[]>([]);
 
+sample({
+  clock: [assetsSettingsModel.$assetsView, once(assetsSettingsModel.events.assetsStarted)],
+  source: assetsSettingsModel.$assetsView,
+  target: portfolioModel.events.activeViewSet,
+});
+sample({
```
Asmadek: add empty line here

src/renderer/shared/core/types/asset.ts
```diff
@@ -30,3 +33,21 @@ export type OrmlExtras = {
   existentialDeposit: string;
   transfersEnabled?: boolean;
 };
+
+export type TokenAsset = {
```
Asmadek: I think we should discuss naming, because for me token and asset are synonyms. But I see the idea that it's means `asset type for token centric view`

src/renderer/pages/Assets/Assets/Assets.tsx
```diff
@@ -30,8 +33,14 @@ export const Assets = () => {
...
         <ShardSelectorButton />
-
-        <AssetsChainView query={query} activeShards={activeShards} hideZeroBalances={hideZeroBalances} />
+        <div className="flex flex-col gap-y-4 w-full h-full overflow-y-scroll">
+          {activeShards.length > 0 &&
+            (assetsView === AssetsListView.CHAIN_CENTRIC ? (
+              <AssetsChainView query={query} activeShards={activeShards} hideZeroBalances={hideZeroBalances} />
+            ) : (
+              <AssetsPortfolio />
+            ))}
```
Asmadek: Maybe use object here? 

```typescript
{
    [AssetsListView.CHAIN_CENTRIC]: <AssetsChainView query={query} activeShards={activeShards} hideZeroBalances={hideZeroBalances} />,
    [AssetsListView.TOKEN_CENTRIC]: <AssetsPortfolio />,
}[assetsView]
```

It will help faster find place when search, and provide better readability

src/renderer/features/assets/AssetsPortfolio/ui/TokenBalanceList.tsx
```diff
@@ -0,0 +1,114 @@
...
+                </Link>
+              </CheckPermission>
+            </div>
+          </div>
+        </Accordion.Button>
+
+        <Accordion.Content className="mt-1">
+          <ul className="flex flex-col gap-y-1.5 pl-6">
+            {asset.chains.map((chain) => (
+              <NetworkCard key={chain.chainId + chain.assetId} chain={chain} asset={asset} />
```
Asmadek: Maybe use template literals here? Concat string with + may provide some issues in future

Asmadek (approved):

src/renderer/features/assets/AssetsPortfolio/model/portfolio-model.ts
Asmadek: If wallet is not multisig we accept it, otherwise (if wallet is multisig) we check that multisig pallet exists. We don't need to check `isMultisig && hasPallet` directly here.

src/renderer/shared/core/types/balance.ts
```diff
@@ -20,3 +20,11 @@ export type BalanceLock = {
   type: LockTypes;
   amount: string;
 };
+
+export type TokenBalance = {
```
tuul-wq: This looks pretty strange, because we already have `Balance` with same fields ...
Maybe if we had `ChainAccountBalance` instead of `Balance` and `Balance` instead of `TokenBalance` things were a bit clearer but it involves DB model refactoring 😔
The only suggestion from my side - `export type AssetBalance = Pick<Balance, 'verified' | 'free' | 'reserved' | 'frozen' | 'locked'>;`

src/renderer/shared/lib/utils/balance.ts
```diff
@@ -86,7 +86,7 @@ export const formatBalance = (balance = '0', precision = 0): FormattedBalance =>
   };
 };
 
-export const totalAmount = (balance?: Balance): string => {
+export const totalAmount = (balance?: Balance | TokenBalance): string => {
```
tuul-wq: I suppose something like this should work
```suggestion
export const totalAmount = <T extends TokenBalance>(balance?: T): string => {
```

src/renderer/features/assets/AssetsChainView/ui/AssetCard/AssetCard.tsx
```diff
@@ -72,24 +66,7 @@ export const AssetCard = ({ chainId, asset, balance }: Props) => {
             </div>
           )}
         </div>
-        <div className="flex gap-x-2 ml-3">
```
tuul-wq, line 75: 👍

src/renderer/features/assets/AssetsPortfolio/lib/tokensService.ts
```diff
@@ -0,0 +1,81 @@
+import tokensProd from '@shared/config/tokens/tokens.json';
+import tokensDev from '@shared/config/tokens/tokens_dev.json';
+import { sumValues } from '@shared/api/network/service/chainsService';
+import type { Account, AccountId, Balance, ChainId, TokenAsset, TokenBalance } from '@shared/core';
+import { ZERO_BALANCE, totalAmount } from '@shared/lib/utils';
+import { balanceUtils } from '@entities/balance';
+import { accountUtils } from '@entities/wallet';
+import { AssetChain } from './types';
+
+const TOKENS: Record<string, any> = {
```
tuul-wq: We use enum-like naming for such constants
```javascript
const Tokens = { DEV: ..., PROD: ... };
```

src/renderer/features/assets/AssetsPortfolio/lib/tokensService.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  return {
+    verified: firstBalance.verified && secondBalance.verified,
+    free: sumValues(firstBalance.free, secondBalance.free),
+    reserved: sumValues(firstBalance.reserved, secondBalance.reserved),
+    frozen: sumValues(firstBalance.frozen, secondBalance.frozen),
+    locked: (firstBalance.locked || []).concat(secondBalance.locked || []),
+  };
+}
+function getSelectedAccountIds(accounts: Account[], chainId: ChainId): AccountId[] {
```
tuul-wq: Empty line between function declarations 😉

src/renderer/features/assets/AssetsPortfolio/model/portfolio-model.ts
```diff
@@ -0,0 +1,167 @@
...
+const hideZeroBalancesSet = createEvent<boolean>();
+
+const $activeView = createStore<AssetsListView | null>(null);
+const $accounts = createStore<Account[]>([]);
+const $hideZeroBalances = createStore<boolean>(false);
+const $tokens = createStore<TokenAsset[]>([]);
+const $activeTokens = createStore<TokenAsset[]>([]);
+
+const updateTokensFx = createEffect(
+  ({ activeWallet, chains }: { activeWallet?: Wallet; chains: Record<ChainId, Chain> }): TokenAsset[] => {
```
tuul-wq: Nice to have dedicated types for object arguments like - `UpdateTokenParams`

src/renderer/pages/Assets/Assets/Assets.tsx
tuul-wq: Assets page should only include features, features must decide wheather to render or not
`AssetsPortfolio` & `AssetsChainView` should check `assetsView === AssetsListView.CHAIN_CENTRIC`
and return `null` or meaningful content.

tuul-wq (changes requested):

src/renderer/pages/Assets/Assets/Assets.tsx
sokolova-an: I think in general it's more common to have a ternary operator for such cases so it could be easier for outsiders and  newcomers to understand what's going on here

src/renderer/features/assets/AssetsPortfolio/lib/tokensService.ts
sokolova-an: In this case, I suggest changing construction in `chainsService` as well otherwise we'll have a different approach for the same thing
```
const CHAINS: Record<string, any> = {
  chains: chainsProd,
  'chains-dev': chainsDev,
};
```
Sounds good?

src/renderer/shared/core/types/balance.ts
sokolova-an: I agree, would be better this way. I'd like to have there `BalanceDetails`or something like that and make it part of the `Balance` type. But struggle to come up with a meaningful name for it since it will be used separately as well 
maybe something like `BalanceSummary/BalanceDetails/BalanceAmount`?

src/renderer/pages/Assets/Assets/Assets.tsx
sokolova-an: @tuul-wq I tried this way but in this case, on the `Assets` page we have a structure like this: 

```
 <AssetsChainView query={query} activeShards={activeShards} hideZeroBalances={hideZeroBalances} />
 <AssetsPortfolio />
```

And it's not obvious that it is not just part of the page's component but a separate view of this page when one excluded another. If we have it this way, then we can't read the page at first glance, we need to go deeper through files to understand what's going on there and how this works

src/renderer/features/assets/AssetsPortfolio/lib/tokensService.ts
tuul-wq: Yeah, agree

src/renderer/pages/Assets/Assets/Assets.tsx
tuul-wq: That's totally fine, because page includes features/widgets so each UI component leads us directly to dedicated functionality
We could change Assets views names so they more clearly tell us about their purpose
`AssetsChainView` - `AssetsPortfolioView` 
This way anyone who's looking at the Assets page will see that it has X features (filter, settings, and a bunch of different views)
If one day condition for PortfolioView or ChainView changes you won't need to alter Assets page at all which is perfectly matches Single responsibility principle

src/renderer/features/assets/AssetsPortfolio/lib/tokensService.ts
sokolova-an: Update: we use it this way to avoid type issues with `process.env.TOKENS_FILE`

src/renderer/features/assets/AssetsChainView/ui/AssetsChainView.tsx
```diff
@@ -10,14 +10,18 @@ import { balanceModel } from '@entities/balance';
...
 
 type Props = {
   query: string;
   activeShards: Account[];
   hideZeroBalances: boolean;
+  assetsView: AssetsListView;
 };
-export const AssetsChainView = ({ query, activeShards, hideZeroBalances }: Props) => {
+export const AssetsChainView = ({ query, activeShards, hideZeroBalances, assetsView }: Props) => {
+  if (assetsView !== AssetsListView.CHAIN_CENTRIC) return null;
```
tuul-wq: Should come after all the hooks

src/renderer/features/assets/AssetsPortfolio/lib/tokensService.ts
```diff
@@ -0,0 +1,78 @@
...
+  return accounts.reduce<AccountId[]>((acc, account) => {
+    if (accountUtils.isChainIdMatch(account, chainId)) {
+      acc.push(account.accountId);
+    }
+
+    return acc;
+  }, []);
+}
+function getChainWithBalance(
```
tuul-wq: Empty line

src/renderer/features/assets/AssetsPortfolio/lib/tokensService.ts
```diff
@@ -0,0 +1,78 @@
...
+  balances: Balance[],
+  chains: AssetChain[],
+  hideZeroBalances: boolean,
+  accounts: Account[],
+): [AssetChain[], AssetBalance] {
+  let totalBalance = {} as AssetBalance;
+
+  const chainsWithBalance = chains.reduce((acc, chain) => {
```
tuul-wq: instead os `as AssetChain[]`
```suggestion
  const chainsWithBalance = chains.reduce<AssetChain[]>((acc, chain) => {
```

src/renderer/features/assets/AssetsPortfolio/lib/tokensService.ts
```diff
@@ -0,0 +1,78 @@
...
+    const selectedAccountIds = getSelectedAccountIds(accounts, chain.chainId);
+
+    const accountsBalance = balanceUtils.getAssetBalances(
+      balances,
+      selectedAccountIds,
+      chain.chainId,
+      chain.assetId.toString(),
+    );
+
+    const assetBalance = accountsBalance.reduce<AssetBalance>((acc, balance) => {
```
tuul-wq: Why do we have `reduce<AssetBalance>` & `as Balance` at the same time?

src/renderer/features/assets/AssetsPortfolio/index.tsx
tuul-wq, line 1: Don't forget to rename `AssetsPortfolio` to `AssetsPortfolioView`

src/renderer/features/assets/AssetsPortfolio/model/portfolio-model.ts
```diff
@@ -0,0 +1,168 @@
...
+import { Account, Balance, Chain, ChainId, AssetByChains, Wallet } from '@shared/core';
+import { networkModel, networkUtils } from '@entities/network';
+import { accountUtils, walletModel, walletUtils } from '@entities/wallet';
+import { AssetsListView } from '@entities/asset';
+import { balanceModel } from '@entities/balance';
+import { tokensService } from '../lib/tokensService';
+
+const activeViewSet = createEvent<AssetsListView>();
+const accountsSet = createEvent<Account[]>();
+const hideZeroBalancesSet = createEvent<boolean>();
```
tuul-wq: More generic word for events that trigger or change is `somethingChanged`
`activeViewChanged`, `accountsChanged`, `hideZeroBalancesChanged`

src/renderer/features/assets/AssetsPortfolio/model/portfolio-model.ts
```diff
@@ -0,0 +1,168 @@
...
+
+const $activeView = createStore<AssetsListView | null>(null);
+const $accounts = createStore<Account[]>([]);
+const $hideZeroBalances = createStore<boolean>(false);
+const $tokens = createStore<AssetByChains[]>([]);
```
tuul-wq: I'd rather use term `assets` / `activeAssets` and avoid `tokens`
@sokolova-an @Asmadek what do you guys think?

src/renderer/features/assets/AssetsPortfolio/model/portfolio-model.ts
```diff
@@ -0,0 +1,168 @@
...
+        );
+      });
+    });
+
+    if (filteredChains.length > 0) {
+      acc.push({ ...token, chains: filteredChains });
+    }
+
+    return acc;
+  }, [] as AssetByChains[]);
```
tuul-wq: `reduce` accept generic to avoid `as Type` at the end 😉

src/renderer/features/assets/AssetsPortfolio/model/portfolio-model.ts
```diff
@@ -0,0 +1,168 @@
...
+const $hideZeroBalances = createStore<boolean>(false);
+const $tokens = createStore<AssetByChains[]>([]);
+const $activeTokens = createStore<AssetByChains[]>([]);
+
+type UpdateTokenParams = {
+  activeWallet?: Wallet;
+  chains: Record<ChainId, Chain>;
+};
+
+const updateTokensFx = createEffect(({ activeWallet, chains }: UpdateTokenParams): AssetByChains[] => {
```
tuul-wq: What about `getUpdatedTokensFx` to be more clear about our intention to receive data calculated based on `wallet` & `chains`?

src/renderer/features/assets/AssetsPortfolio/model/portfolio-model.ts
```diff
@@ -0,0 +1,168 @@
...
+});
+
+sample({
+  clock: accountsSet,
+  target: $accounts,
+});
+sample({
+  clock: hideZeroBalancesSet,
+  target: $hideZeroBalances,
+});
```
tuul-wq: Take a look at `restore` from [effector](https://effector.dev/en/api/effector/restore/)
It helps to avoid such `samples`, we already use a bunch of them in the project

src/renderer/features/assets/AssetsPortfolio/model/portfolio-model.ts
```diff
@@ -0,0 +1,168 @@
...
+
+sample({
+  clock: [walletModel.$activeWallet, $activeView, once(networkModel.events.networkStarted)],
```
tuul-wq: `networkModel.events.networkStarted` is made for internal usage by the `networkModel`
You can observe `once(networkModel.$chains)`, let me know if that doesn't work or has some drawbacks

src/renderer/features/assets/AssetsPortfolio/model/portfolio-model.ts
```diff
@@ -0,0 +1,168 @@
...
+  fn: ({ connections, chains, tokens, activeWallet }): AssetByChains[] => {
+    const isMultisigWallet = walletUtils.isMultisig(activeWallet);
+
+    return tokens.reduce((acc, token) => {
+      const filteredChains = token.chains.filter((c) => {
+        if (!connections[c.chainId]) return false;
+        const isDisabled = networkUtils.isDisabledConnection(connections[c.chainId]);
+        const hasMultiPallet = networkUtils.isMultisigSupported(chains[c.chainId].options);
+
+        return !isDisabled && (!isMultisigWallet || hasMultiPallet);
```
tuul-wq: @Tbaut made this condition a bit verbose, hope it helps you with readability

src/renderer/features/assets/AssetsPortfolio/ui/AssetsPortfolioView.tsx
```diff
@@ -0,0 +1,67 @@
...
+const getColStyle = (wallet?: Wallet) => {
+  switch (wallet?.type) {
+    case WalletType.WATCH_ONLY:
+      return 'grid-cols-[1fr,100px,105px]';
+    case WalletType.PROXIED:
+      return 'grid-cols-[1fr,100px,105px,30px]';
+    default:
+      return 'grid-cols-[1fr,100px,105px,50px]';
+  }
+};
```
tuul-wq: What about this one?
```suggestion
const getColStyle = (wallet?: Wallet): string => {
  const colStyleMap = {
    [WalletType.WATCH_ONLY]: 'grid-cols-[1fr,100px,105px]',
    [WalletType.PROXIED]: 'grid-cols-[1fr,100px,105px,30px]',
    default: 'grid-cols-[1fr,100px,105px,50px]',
  };

  return colStyleMap[wallet?.type] || colStyleMap.default;
};
```

src/renderer/features/assets/AssetsPortfolio/ui/TokenBalanceList.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+type Props = {
+  asset: AssetByChains;
+};
+
+export const TokenBalanceList = ({ asset }: Props) => {
+  const { t } = useI18n();
+
+  const fiatFlag = useUnit(priceProviderModel.$fiatFlag);
+
```
tuul-wq: Redundant empty line

src/renderer/features/assets/AssetsSettings/model/assets-settings-modal.ts
```diff
@@ -32,6 +33,12 @@ sample({
   target: [getHideZeroBalancesFx, getAssetsViewFx],
 });
 
+sample({
```
tuul-wq: Seems strange that settings feature cares about starting `price rates request` 🤔
Can we move this into the `Assets page` model?

src/renderer/pages/Assets/Assets/Assets.tsx
```diff
@@ -30,8 +32,19 @@ export const Assets = () => {
           </div>
         </Header>
         <ShardSelectorButton />
-
-        <AssetsChainView query={query} activeShards={activeShards} hideZeroBalances={hideZeroBalances} />
+        <div className="flex flex-col gap-y-4 w-full h-full overflow-y-scroll">
+          {activeShards.length > 0 && (
```
tuul-wq: This condition raises same point as previous condition with `AssetView` 😌
We pass `activeShards` to portfolio and can use it to determine if UI should be rendered
For `AssetsChainView` it's passed through props

src/renderer/pages/Assets/Assets/model/assets-model.ts
```diff
@@ -30,6 +43,11 @@ sample({
   target: $activeShards,
 });
 
+sample({
```
tuul-wq, line 58: (continue the last comment) Passing `activeShards` here

tuul-wq (changes requested):
Last part of comments 👍

src/renderer/pages/Assets/Assets/Assets.tsx
sokolova-an: And again I don't like how it looks like, we pass the same prop twice make the view components less readable even though it's a common conditions for the both views 

But if you believe it's will be better overall - we can have it this way

src/renderer/features/assets/AssetsPortfolio/model/portfolio-model.ts
sokolova-an: in ui library we have a new component  -`TokenBalance` thas why I named the same way the component for this and the store we use there

tuul-wq (approved):

Tbaut (approved):

src/renderer/features/assets/AssetsPortfolioView/ui/TokenBalance.tsx
```diff
@@ -0,0 +1,59 @@
...
+        <div className="flex items-center gap-x-2">
+          <AssetIcon src={asset.icon} name={asset.name} />
+          <div>
+            <BodyText>{chain.assetSymbol}</BodyText>
+            <div className="flex items-center gap-x-1.5 mr-3">
+              <FootnoteText className="text-text-tertiary">{chain.name}</FootnoteText>
+              <ChainIcon src={chains[chain.chainId].icon} name={chain.name} size={18} />
+              {chain.balance?.verified && (
+                <div className="flex items-center gap-x-2 text-text-warning">
+                  {/* FIXME: tooltip not visible when first displayed network invalid. For now just render it below icon */}
```
Tbaut, line 37: Can we create an issue and eventually mention it here? Such FIXME, unless they are tackled right away, tend to stay in the code forever.

pgolovkin (approved):

PR #1745, opened 2024-05-06
Feat: Multisig discovery

### LINKED ISSUE #1741: Multisig discovery
As a user, I would like to see all the multisigs that I am a signatory.

follow-up of https://github.com/novasamatech/nova-spektr/pull/1729 and https://github.com/novasamatech/subquery-proxy/pull/6

Tbaut:
closes https://github.com/novasamatech/nova-spektr/issues/1741


src/renderer/entities/multisig/api/MultisigsService.ts
```diff
@@ -0,0 +1,26 @@
...
+export type MultisigResult = {
+  accountId: AccountId;
+  threshold: number;
+  signatories: AccountId[];
+};
+
+async function filterMultisigsAccountIds(client: GraphQLClient, accountIds: AccountId[]): Promise<MultisigResult[]> {
+  const data = await client.request(FILTER_MULTISIG_ACCOUNT_IDS, { accountIds });
+
+  const filteredMultisigIds = (data as any)?.accounts?.nodes?.map(({ id, threshold, signatories }: any) => ({
```
Asmadek: I am not sure about such naming, because it stores not only ids but threshold and signatories

src/renderer/features/multisigs/model/multisigs-model.ts
```diff
@@ -0,0 +1,177 @@
...
+
+const multisigsDiscoveryStarted = createEvent();
+const chainConnected = createEvent<ChainId>();
+
+type GetMultisigsParams = {
+  chain: Chain;
+  accounts: Account[];
+};
+
+type StartChainsProps = {
```
Asmadek: Better to use params instead of props for function params

src/renderer/features/multisigs/model/multisigs-model.ts
```diff
@@ -0,0 +1,177 @@
...
+  chain: Chain;
+  indexedMultisigs: MultisigResult[];
+};
+
+const connectChainsFx = createEffect(({ chains, connections }: StartChainsProps) => {
+  const boundConnected = scopeBind(chainConnected, { safe: true });
+
+  chains.forEach((chain) => {
+    if (networkUtils.isDisabledConnection(connections[chain.chainId])) return;
+    boundConnected(chain.chainId);
```
Asmadek: Better to add empty line here

src/renderer/features/multisigs/model/multisigs-model.ts
```diff
@@ -0,0 +1,177 @@
...
+      const isOwnWalletAlready = sameWallet?.length;
+
+      return !isOwnWalletAlready;
+    });
+
+    const walletsToSave = multisigsToSave.map(
+      ({ threshold, accountId, signatories }) =>
+        ({
+          wallet: {
+            name: `Detected msig ${toAddress(accountId).slice(0, 7)}...`,
```
Asmadek: Maybe use `toAddress` here too as for account name?

src/renderer/features/multisigs/model/multisigs-model.ts
```diff
@@ -0,0 +1,177 @@
...
+  source: {
+    wallets: walletModel.$wallets,
+  },
+  filter: (_, { indexedMultisigs }) => {
+    return indexedMultisigs.length > 0;
+  },
+  fn: ({ wallets }, { indexedMultisigs, chain }) => {
+    const multisigsToSave = indexedMultisigs.filter((multisigrResult) => {
+      // we filter out the multisigs that we already have
+      const sameWallet = walletUtils.getWalletsFilteredAccounts(wallets, {
```
Asmadek: If we need only one wallet, maybe use `walletUtils.getWalletFilteredAccounts`

Asmadek (approved):

src/renderer/app/index.tsx
```diff
@@ -40,6 +43,7 @@ networkModel.events.networkStarted();
 proxyModel.events.proxyStarted();
 assetsSettingsModel.events.assetsStarted();
 notificationModel.events.notificationsStarted();
+setInterval(multisigsModel.events.multisigsDiscoveryStarted, MULTISIG_DISCOVERY_TIMEOUT);
```
tuul-wq: `App` layer is responsible only for initiation, here we not just start the process of multisig discovery but use specific constants and intervals that relates to multisig discovery "feature"
In a while we might decide to change interval time manually through the settings page, with current solution it's not possible.
Overall background processes must happen in `Process` layer. It's ideal for things like `balances, proxy discovery, multisig discovery`.
We don't have that layer right now, it would be right to introduce it during this task.
`Process` has access to all underlying layers - `widgets, features, entities, shared`.
This way we could easily change settings connected with "discovery services" and be sure that responsible process will react as intended.

src/renderer/entities/multisig/api/MultisigsService.ts
tuul-wq, line 1: `multisigService.ts`

src/renderer/entities/wallet/model/wallet-model.ts
```diff
@@ -9,7 +9,7 @@ import { modelUtils } from '../lib/model-utils';
 
 type DbWallet = Omit<Wallet, 'accounts'>;
 
-type CreateParams<T extends Account = Account> = {
+export type CreateParams<T extends Account = Account> = {
```
tuul-wq: Let it stay internal, `wallet-model` should not handle wallet creation, but it is what it is
Every feature that creates wallet should declare it's own `InputParams` required for wallet creation

src/renderer/features/multisigs/model/__tests__/multisigs-model.test.ts
```diff
@@ -0,0 +1,134 @@
...
+  },
+  {
+    accountId: '0x03',
+    address: 'F7XB',
+  },
+];
+
+jest.mock('@entities/multisig', () => ({
+  ...jest.requireActual('@entities/multisig'),
+  multisigService: {
```
tuul-wq: `jest.spyOn(multisigService, 'filterMultisigsAccountIds')` doesn't work?

src/renderer/features/multisigs/model/multisigs-model.ts
```diff
@@ -0,0 +1,177 @@
...
+        read: false,
+        type: NotificationType.MULTISIG_CREATED,
+        dateCreated: Date.now(),
+      },
+    ]);
+  });
+});
+
+sample({
+  clock: [multisigsDiscoveryStarted, once(networkModel.$connections)],
```
tuul-wq: Why not trigger `multisigsDiscoveryStarted` with some specific interval?
Include `networkModel.$connections` in `source` and filter them to be not empty

src/renderer/features/multisigs/model/multisigs-model.ts
```diff
@@ -0,0 +1,177 @@
...
+  source: {
+    connections: networkModel.$connections,
+    chains: networkModel.$chains,
+  },
+  fn: ({ connections, chains }) => ({
+    chains: Object.values(chains).filter((chain) => multisigUtils.isMultisigSupported(chain)),
```
tuul-wq: This can be done a single time with `combine` - `const multisigChains = combine(networkModel$chains)`

tuul-wq (changes requested):

src/renderer/features/multisigs/model/multisigs-model.ts
tuul-wq, line 1: In general this code works, but I want to discuss some approaches that can improve it
@Tbaut let me know when we can have a call

tuul-wq (changes requested):

src/renderer/features/multisigs/model/__tests__/multisigs-model.test.ts
Tbaut: I tried with the following
```ts
jest.spyOn(multisigService, 'filterMultisigsAccounts').mockResolvedValue([
  {
    accountId: '0x00',
    threshold: 2,
    signatories: ['0x01', '0x02', '0x03'],
  },
]);
```

but this doesn't work, as in when I print the result of `multisigService.filterMultisigsAccounts` in the model, it never shows the log. It's called for sure because if I put a log before the call, this is printed.

src/renderer/features/multisigs/model/multisigs-model.ts
```diff
@@ -15,69 +12,71 @@ import {
...
 import { MultisigResult, multisigService } from '@entities/multisig';
-import { multisigUtils } from '../lib/mulitisigs-utils';
 import { isEthereumAccountId, toAddress } from '@shared/lib/utils';
 import { CreateParams } from '@entities/wallet/model/wallet-model';
 import { notificationModel } from '@entities/notification';
+import { multisigUtils } from '../lib/mulitisigs-utils';
+
+const MULTISIG_DISCOVERY_TIMEOUT = 30000;
 
 const multisigsDiscoveryStarted = createEvent();
```
Tbaut: I have kept the `multisigsDiscoveryStarted` event, because the `interval` method needs an event to `start`.

src/renderer/features/multisigs/model/multisigs-model.ts
```diff
@@ -15,69 +12,71 @@ import {
...
-const chainConnected = createEvent<ChainId>();
+const multisigSaved = createEvent<GetMultisigsResult>();
 
-type GetMultisigsParams = {
-  chain: Chain;
-  accounts: Account[];
-};
+const $chainsSupportingMultisigDiscovery = combine(networkModel.$chains, (chains) =>
```
Tbaut: I am doing TDD with this refactor, and I'm facing an issue:
> thrown: "Exceeded timeout of 5000 ms for a test.

Reducing the `MULTISIG_DISCOVERY_TIMEOUT` doesn't help, I checked and the interval is ticking as expected, however, the issue seems to be with this combine, the `chainsSupportingMultisigDiscovery` is always `[]` and this function to filter the chains is never called. Any hint on what is needed to trigger this?

src/renderer/features/multisigs/model/multisigs-model.ts
```diff
@@ -15,69 +12,71 @@ import {
...
+  Object.values(chains).filter(
+    (chain) => multisigUtils.isMultisigSupported(chain) && chain.externalApi?.[ExternalType.MULTISIG]?.[0]?.url,
+  ),
+);
+
+$chainsSupportingMultisigDiscovery.watch((c) => console.log('---> C :', c));
```
Tbaut: this is always `[]`

Tbaut (commented):
@tuul-wq I've started a refactor to:
- filter compatible chains to start with
- filter connected chains
- call the multisig creation effect with all the chains, then call individual event to save multisigs
- remove the interval from the higher level `index.ts`

There's still some work to be done such as moving to a process and moving some large function into a util. However the meat of the refactor should be there, let me know if this is going in the right direction, and if you can spot the issue I'm facing with the chain store being empty.

src/renderer/features/multisigs/model/multisigs-model.ts
tuul-wq: Take a look at `balance-sub-model.test.ts` - `await jest.runAllTimersAsync();`
When you encounter a timer, jest needs to run it so the timeout is not raised

tuul-wq: And don't hesitate to use `{ return ...}` for readability 😉

Tbaut: Thanks, this is not linked to the timer though I think. I added this to the tests, and they're still failing. The timer is ticking however this store `networkModel.$chains` is never getting changed it seems.. and the sample containing `clock: [multisigDiscoveryTriggered, $chainsSupportingMultisigDiscovery],` doesn't get trigger, not because of `multisigDiscoveryTriggered` but because of `$chainsSupportingMultisigDiscovery`. We used to call `$networkModel.$chains` with `once`, but:
- I don't know if that's actually the issue
- we can't use `once` in a combine it seems

Tbaut:
@tuul-wq The tests are still timing out, and I really have no clue why. My logs show that things are resolved, I imagine that the timeout is due to the `interval`, however my research didn't lead me to a solution so far. Also I saw that we never use `interval` in the current code base. do you have any other idea, or is there a chance you can ask in the effector tg group maybe?

tuul-wq:
> @tuul-wq The tests are still timing out, and I really have no clue why. My logs show that things are resolved, I imagine that the timeout is due to the `interval`, however my research didn't lead me to a solution so far. Also I saw that we never use `interval` in the current code base. do you have any other idea, or is there a chance you can ask in the effector tg group maybe?

We had `interval` recently but I managed to remove it from `balance-sub` model
In general we need to use `runAllTimersAsync` for any type of interval/delay/throttle imported from effector
I'll take a closer look today, let you know the result

Tbaut:
I managed to make it pass, and here is what was actually happening I think. The trick was to remove the `await` in front of the `allSettled(multisigsModel.events.multisigsDiscoveryStarted, { scope });`
The reason it was timing out I think is that it was waiting for everything to settle.. but since we have an interval, it actually never settled. For this reason, the `jest.runAllTimersAsync()` was also not working. Now I did try to have it before the `allSettled` but it didn't work. 

Also, when using `runAllTimersAsync` we're supposed to use `jest.useFakeTimers()`. But I think this doesn't play well with infinite intervals, because it also never ends. 

TL;DR: not waiting for the settlement doesn't sound super safe, but it's the only thing that worked for me.

.eslintrc.js
```diff
@@ -48,6 +48,7 @@ module.exports = {
           ['@renderer', './src/renderer/'],
           ['@app', './src/renderer/app/'],
           ['@pages', './src/renderer/pages/'],
+          ['@processes', './src/renderer/processes/'],
```
tuul-wq, line 51: 👍

src/renderer/processes/multisigs/model/multisigs-model.ts
```diff
@@ -0,0 +1,160 @@
...
+};
+
+const MULTISIG_DISCOVERY_TIMEOUT = 30000;
+
+const multisigsDiscoveryStarted = createEvent();
+const multisigSaved = createEvent<GetMultisigsResult>();
+
+const $chainsSupportingMultisigDiscovery = combine(networkModel.$chains, (chains) => {
+  return Object.values(chains).filter((chain) => {
+    return multisigUtils.isMultisigSupported(chain) && Boolean(chain.externalApi?.[ExternalType.MULTISIG]?.[0]?.url);
```
tuul-wq: Seems fine, I would suggest to introduce 2 variables with meaningful names and `return var_1 && var_2`
for readability

src/renderer/processes/multisigs/model/multisigs-model.ts
```diff
@@ -0,0 +1,160 @@
...
+  });
+});
+
+const { tick: multisigDiscoveryTriggered } = interval({
+  start: multisigsDiscoveryStarted,
+  timeout: MULTISIG_DISCOVERY_TIMEOUT,
+});
+
+sample({
+  clock: [multisigDiscoveryTriggered, once(networkModel.$connections)],
```
tuul-wq: Actually we can go this way
```javascript
sample({
  clock: once(networkModel.$connections), 
  target: multisigsDiscoveryStarted,
});

const { tick: multisigDiscoveryTriggered } = interval({
  start: multisigsDiscoveryStarted,
  timeout: MULTISIG_DISCOVERY_TIMEOUT,
});

// sample for multisigDiscoveryTriggered
```
And remove `multisigsDiscoveryStarted` at all
In `app/index.tsx` just import `import '@processes/multisigs';`

src/renderer/processes/multisigs/model/multisigs-model.ts
```diff
@@ -0,0 +1,160 @@
...
+  source: {
+    chains: $chainsSupportingMultisigDiscovery,
+    wallets: walletModel.$wallets,
+    connections: networkModel.$connections,
+  },
+  fn: ({ chains, wallets, connections }) => {
+    return {
```
tuul-wq, line 112: Let's go with `filteredChains` for the filtered chains and `return { wallets, chains: filteredChains }`

src/renderer/processes/multisigs/model/multisigs-model.ts
```diff
@@ -0,0 +1,160 @@
...
+
+const { tick: multisigDiscoveryTriggered } = interval({
+  start: multisigsDiscoveryStarted,
+  timeout: MULTISIG_DISCOVERY_TIMEOUT,
+});
+
+sample({
+  clock: [multisigDiscoveryTriggered, once(networkModel.$connections)],
+  source: {
+    chains: $chainsSupportingMultisigDiscovery,
```
tuul-wq: What about `chainsForMultisigDiscovery`? don't insist, just a minor suggestion 😉
or super short `multisigChains`

src/renderer/processes/multisigs/model/multisigs-model.ts
```diff
@@ -0,0 +1,160 @@
...
+      wallets,
+    };
+  },
+  target: getMultisigsFx,
+});
+
+sample({
+  clock: multisigSaved,
+  fn: ({ indexedMultisigs, chain }) => {
+    const walletsToSave = indexedMultisigs.map(
```
tuul-wq:
```suggestion
    return indexedMultisigs.map(
```

src/renderer/processes/multisigs/model/multisigs-model.ts
```diff
@@ -0,0 +1,160 @@
...
+  chains: Chain[];
+  wallets: Wallet[];
+};
+
+type GetMultisigsResult = {
+  chain: Chain;
+  indexedMultisigs: MultisigResult[];
+};
+
+const getMultisigsFx = createEffect(({ chains, wallets }: GetMultisigsParams): void => {
```
tuul-wq: Across the project we don't use `void`

src/renderer/processes/multisigs/model/multisigs-model.ts
```diff
@@ -0,0 +1,160 @@
...
+  chains.forEach((chain) => {
+    const accounts = walletUtils.getAccountsBy(wallets, (a) => accountUtils.isChainIdMatch(a, chain.chainId));
+    const multisigIndexerUrl = chain.externalApi?.[ExternalType.MULTISIG]?.[0]?.url;
+
+    if (multisigIndexerUrl && accounts.length) {
```
tuul-wq: Good place for a fast return to reduce nesting

src/renderer/processes/multisigs/model/multisigs-model.ts
```diff
@@ -0,0 +1,160 @@
...
+      const client = new GraphQLClient(multisigIndexerUrl);
+
+      multisigService
+        .filterMultisigsAccounts(
+          client,
+          accounts.map((account) => account.accountId),
```
tuul-wq: Better to extract into the variable

src/renderer/processes/multisigs/model/multisigs-model.ts
```diff
@@ -0,0 +1,160 @@
...
+        )
+        .then((indexedMultisigs) => {
+          const multisigsToSave = indexedMultisigs.filter((multisigrResult) => {
+            // we filter out the multisigs that we already have
+            const sameWallet = walletUtils.getWalletFilteredAccounts(wallets, {
+              accountFn: (account) => {
+                return account.accountId === multisigrResult.accountId;
+              },
```
tuul-wq:
```suggestion
              accountFn: (account) => account.accountId === multisigrResult.accountId,
```

src/renderer/processes/multisigs/model/multisigs-model.ts
```diff
@@ -0,0 +1,160 @@
...
+            });
+
+            const walletAllreadyExist = Boolean(sameWallet);
+
+            return !walletAllreadyExist;
```
tuul-wq:
```suggestion
            return !Boolean(sameWallet);
```

src/renderer/processes/multisigs/model/multisigs-model.ts
```diff
@@ -0,0 +1,160 @@
...
+          });
+
+          multisigsToSave.length > 0 && multisigSaved({ indexedMultisigs: multisigsToSave, chain });
```
tuul-wq: Don't forget to use `scopeBind` from `multisigSaved`
Also short condition with function seems clunky we try to avoid that and replace with simple `if`

tuul-wq (changes requested):
Overall code is good and ready to be approved after the remaining comments 👍

src/renderer/processes/multisigs/model/multisigs-model.ts
Tbaut: I thought you commented earlier having return is easier to read. Maybe I'm mixing it with something else.

src/renderer/processes/multisigs/model/multisigs-model.ts
Tbaut: `multisigChains` it is

src/renderer/processes/multisigs/model/multisigs-model.ts
Tbaut: While I see how this is more elegant, I couldn't manage to make the test work with this setup.

tuul-wq (approved):

PR #1751, opened 2024-05-08
Feat: use send instead of submit and watch

### LINKED ISSUE #1585: Try out ".send" instead of ". submitAndWatchExtrinsic"
`send` allows to pass a callback, and will give back all the status, events right away, without the need to subscribe

@Tbaut suggested to try `send` - [link](https://github.com/novasamatech/nova-spektr/pull/1547)
Documentation [here](https://polkadot.js.org/docs/api/start/api.tx.subs#transaction-events) & [here](https://polkadot.js.org/docs/api/cookbook/tx#how-do-i-get-the-decoded-enum-for-an-extrinsicfailed-event)

```javascript
.send(sender, ({ status, events }) => {
    if (status.isInBlock || status.isFinalized) {
      events
        // find/filter for failed events
        .filter(({ event }) =>
          api.events.system.ExtrinsicFailed.is(event)
        )
        // we know that data for system.ExtrinsicFailed is
        // (DispatchError, DispatchInfo)
        .forEach(({ event: { data: [error, info] } }) => {
          if (error.isModule) {
            // for module errors, we have the section indexed, lookup
            const decoded = api.registry.findMetaError(error.asModule);
            const { docs, method, section } = decoded;

            console.log(`${section}.${method}: ${docs.join(' ')}`);
          } else {
            // Other, CannotLookup, BadOrigin, no extra info
            console.log(error.toString());
          }
        });
    }
```

Asmadek:
* use `signAndSend` for sign operation


pgolovkin (approved):

src/renderer/entities/transaction/lib/transactionService.ts
```diff
@@ -35,6 +35,18 @@ import {
 } from './common/types';
 import { walletUtils } from '../../wallet';
 
+class RawSigner implements Signer {
+  signature: HexString;
+
```
sokolova-an: Why do we have to use classes with signer? Is there no way to do it without it?

sokolova-an (approved):

src/renderer/entities/transaction/lib/transactionService.ts
```diff
@@ -59,73 +71,40 @@ async function getTransactionFee(transaction: Transaction, api: ApiPromise): Pro
...
 
-      const signedBlock = await api.rpc.chain.getBlock();
-      const blockHeight = signedBlock.block.header.number.toNumber();
-      const apiAt = await api.at(signedBlock.block.header.hash);
-      const allRecords = await apiAt.query.system.events();
+  const options = { signer: new RawSigner(signature), ...unsigned };
 
-      let actualTxHash = result.inner;
+  extrinsic
+    // @ts-ignore
```
Tbaut: Can we add a comment to explain why it's ok to ignore TS errors here?

src/renderer/entities/transaction/lib/common/types.ts
```diff
@@ -22,13 +22,6 @@ import { DecodedTransaction, Transaction, TransactionType } from '@entities/tran
 // =====================================================
 
 export type ITransactionService = {
-  createPayload: (
```
tuul-wq, line 25: 👍

src/renderer/entities/transaction/lib/transactionService.ts
tuul-wq: The main issue is
```javascript
TS2322: Type `0x${string}` is not assignable to type number | IExtrinsicEra | undefined
```
I also wonder how to correctly convert era, or should it be skipped at all 🤔

tuul-wq (approved):

src/renderer/entities/transaction/lib/transactionService.ts
tuul-wq: Moved to a separate file
```javascript
const options: Partial<SignerOptions> = {
    era: api.registry.createType('ExtrinsicEra', unsigned.era),
    signer: new RawSigner(signature),
    assetId: unsigned.assetId,
    tip: unsigned.tip,
    nonce: unsigned.nonce,
    blockHash: unsigned.blockHash,
  };
```
`signer` expects something that implements `Signer` interface so the easiest way is to create a class
@Asmadek anyway, can discuss another solution

sokolova-an (approved):

PR #1796, opened 2024-05-24
fix: wrong transferable part calculation

### LINKED ISSUE #1794: Wrong transferable part calculation.
## Description

There is cases when we will show Transferable balance part as a zero, but actually that's wrong.

There are two cases, when transferable should be calculate by old and new algorithms:
https://github.com/novasamatech/nova-wallet-android/blob/6fcc7f37cda8f98e540a18c196d0d60c1e69185c/feature-wallet-api/src/main/java/io/novafoundation/nova/feature_wallet_api/domain/model/Asset.kt#L39-L44

New calculation looks like:

```kotlin
 fun holdAndFreezesTransferable(free: Balance, frozen: Balance, reserved: Balance): Balance {
      val freeCannotDropBelow = (frozen - reserved).atLeastZero()

      return (free - freeCannotDropBelow).atLeastZero()
  }
```
https://github.com/novasamatech/nova-wallet-android/blob/6fcc7f37cda8f98e540a18c196d0d60c1e69185c/feature-wallet-api/src/main/java/io/novafoundation/nova/feature_wallet_api/domain/model/Asset.kt#L61-L65

Our calculations:
https://github.com/novasamatech/nova-spektr/blob/605b0a2f635c3f6546c6cf8e556bd80e5391aa20/src/renderer/shared/lib/utils/balance.ts#L105-L112

In case, If bnFree is not greater than bnFrozen, it returns ZERO_BALANCE, so in case when  user has that proportions:
```
{
  nonce: 779
  consumers: 3
  providers: 1
  sufficients: 0
  data: {
    free: 42,813,698,533,061
    reserved: 22,170,750,000,000
    frozen: 47,533,665,515,561
    flags: 170,141,183,460,469,231,731,687,309,040,608,171,718
  }
}
```

### 5EpsavEfPSZC8X2E4zHdJdepzpzfSccqNP5RUVyAU9Hyi3z4

it will return 0, but actually it's not true

*Actual result:*

Screenshot 2024-05-24 at 10 46 41

*Expected result:*

Screenshot 2024-05-24 at 10 46 42

Screenshot 2024-05-24 at 10 46 43

pgolovkin:
Closes #1794

### DIFF
--- a/src/renderer/shared/lib/utils/balance.ts
+++ b/src/renderer/shared/lib/utils/balance.ts
@@ -107,8 +107,10 @@ export const transferableAmount = (balance?: Balance): string => {
 
   const bnFree = new BN(balance.free || ZERO_BALANCE);
   const bnFrozen = new BN(balance.frozen || ZERO_BALANCE);
+  const bnReserved = new BN(balance.reserved || ZERO_BALANCE);
+  const freeCannotDropBelow = bnFrozen.gt(bnReserved) ? bnFrozen.sub(bnReserved) : BN_ZERO;
 
-  return bnFree.gt(bnFrozen) ? bnFree.sub(bnFrozen).toString() : ZERO_BALANCE;
+  return bnFree.gt(freeCannotDropBelow) ? bnFree.sub(freeCannotDropBelow).toString() : ZERO_BALANCE;
 };
 
 const stakedAmount = ({ locked = [] }: Balance): string => {


pgolovkin:
Calculation reference from Nova Wallet:

https://github.com/novasamatech/nova-wallet-android/blob/6fcc7f37cda8f98e540a18c196d0d60c1e69185c/feature-wallet-api/src/main/java/io/novafoundation/nova/feature_wallet_api/domain/model/Asset.kt#L39-L44

```
 fun holdAndFreezesTransferable(free: Balance, frozen: Balance, reserved: Balance): Balance {
      val freeCannotDropBelow = (frozen - reserved).atLeastZero()

      return (free - freeCannotDropBelow).atLeastZero()
  }
```

pgolovkin:
For each asset `REGULAR` and `HOLDS_AND_FREEZES` modes have to be added.

PR #1800, opened 2024-05-27
Feat: Multisig creation with remark

### LINKED ISSUE #1756: Redo the multisig creation to create a system.remark call
With the removal of matrix, and the new way of discovering multisigs, the multisig creation now consists in:
- making a multisig tx, any really, a system.remark is a simple one. It serves as a nice firt tx for users, but it's not necessary to be signed by everyone.
- the indexer picks it up
- the UI shows it

Any chance we can have mockups and a brainstorming sessions @pgolovkin 
Some questions:
- what happens when the Tx is sent, but the indexer still didn't pick it up (I need to check how long it takes exactly)?
- should we 
  - make polling every 5s then instead of 30 for this time
  - or do we actually create the multisig in the DB, show it, and the pending multisig tx will eventually appear? (I prefer this way)

### LINKED ISSUE #1807: Remove multisig Description and Reject reason
Remove multisig description and reject reason.
1. Remove multisig Description on the create multisig operation screen
2. Remove filling Cancel reason when the user rejects multisig operation
3. Remove displaying multisig Description on multisig details screen
4. Remove displaying multisig Cancel reason on multisig details screen

Doc links: 
1. Create and display multisig operation doc https://docs.google.com/document/d/1wWnLwLnu4iddV9O1CJUkZ1ykY56oj18CKaMrI0TytUM/edit#heading=h.vcheoh7621sm
2. Operation confirmation screens doc https://docs.google.com/document/d/1oB7pX_tgBJJD4YI_pSjIwVmSHb7J1la-njYrNPYDmpc/edit#heading=h.vcheoh7621sm

Tbaut:
closes https://github.com/novasamatech/nova-spektr/issues/1756
closes #1807


Tbaut:
There are still quite some things to iron out, such as error management for the form submission, UI glitches, and I'm investigating an issue I found in the multisig indexer, but I'd like to have your opinion on things such as:
- do we need to support proxy signers
- should I organize models differently (right now there's 1 for the flow, 1 for most of the data, 1 for the confirmation). I basically took example on the "add proxy" flow

src/renderer/shared/ui/Modals/BaseModal/BaseModal.tsx
```diff
@@ -40,10 +40,10 @@ export const BaseModal = ({
         <div className="fixed inset-0 overflow-hidden flex min-h-full items-center justify-center p-4">
           <ModalTransition>
             <Dialog.Panel
-              className={cnTw(
-                'transform rounded-lg bg-white text-left align-middle shadow-modal transition-all w-[440px]',
-                panelClass,
-              )}
+              className={
+                'transform rounded-lg bg-white text-left align-middle shadow-modal transition-all w-[440px] ' +
```
Asmadek: String concatenation is tricky, use cnTw is preferable, it correctly work when we try to replace some classes from default class.

src/renderer/widgets/CreateWallet/model/confirm-model.ts
```diff
@@ -0,0 +1,75 @@
...
+  },
+  ({ store, wallets }) => {
+    if (!store) return null;
+
+    return walletUtils.getWalletById(wallets, store.signer.walletId);
+  },
+  { skipVoid: false },
+);
+
+// const $proxiedWallet = combine(
```
Asmadek: Remove commented code or add comment when we should uncomment it

src/renderer/widgets/CreateWallet/model/flow-model.ts
Asmadek, line 1: Now we call main model as widget name, `create-wallet-model` here

src/renderer/widgets/CreateWallet/ui/MultisigWallet/ConfirmationStep.tsx
```diff
@@ -0,0 +1,176 @@
...
+  );
+};
+
+const AccountSelector = ({ chain }: { chain?: Chain }) => {
+  const { t } = useI18n();
+
+  const selectedSigner = useUnit(flowModel.$selectedSigner);
+  const accountSignatories = useUnit(formModel.$accountSignatories);
+  const wallets = useUnit(walletModel.$wallets);
+  // const proxiedAccounts = useUnit(formModel.$proxiedAccounts);
```
Asmadek: commented code

src/renderer/widgets/CreateWallet/ui/MultisigWallet/ConfirmationStep.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+  if (!chain) return null;
+
+  const options = accountSignatories.map(({ accountId, name }) => {
+    //   const isShard = accountUtils.isShardAccount(account);
```
Asmadek: commented code

src/renderer/widgets/CreateWallet/ui/MultisigWallet/FlexibleMultisigWallet.tsx
```diff
@@ -59,33 +59,29 @@ export const MultiChainMultisigWallet = ({ isOpen, onClose, onComplete, onBack }
...
   }, [onComplete]);
 
-  const goToPrevStep = () => {
-    if (activeStep === Step.INIT) {
-      onBack();
-    } else {
-      setActiveStep((prev) => prev - 1);
-    }
-  };
+  // const goToPrevStep = () => {
```
Asmadek: remove it?

src/renderer/widgets/CreateWallet/ui/MultisigWallet/FlexibleMultisigWallet.tsx
```diff
@@ -59,33 +59,29 @@ export const MultiChainMultisigWallet = ({ isOpen, onClose, onComplete, onBack }
...
   const closeMultisigModal = (params: { complete?: boolean; closeAll?: boolean } = { closeAll: true }) => {
     toggleIsModalOpen();
 
     setTimeout(params?.complete ? onComplete : params?.closeAll ? onClose : noop, DEFAULT_TRANSITION);
   };
 
-  const submitHandler = (args: any) => {
-    toggleResultModal();
-    setName(args.name);
+  // const submitHandler = (args: any) => {
```
Asmadek: remove it?

src/renderer/widgets/CreateWallet/ui/MultisigWallet/MultisigWalletWizard.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { DEFAULT_TRANSITION } from '@shared/lib/utils';
+import { MultisigWallet } from './MultisigWallet';
+
+type Props = {
+  isOpen: boolean;
+  onClose: () => void;
+  onComplete: () => void;
+};
+
+// const enum Step {
```
Asmadek: Add comment for commented code or remove it, because now this component looks useless

src/renderer/widgets/CreateWallet/ui/MultisigWallet/NameThresholdStep.tsx
```diff
@@ -0,0 +1,119 @@
...
+import { useI18n } from '@app/providers';
+import { DropdownOption } from '@shared/ui/types';
+import { networkModel, networkUtils } from '@entities/network';
+import { ChainTitle } from '@entities/chain';
+import { Signatory, type Chain } from '@shared/core';
+import { formModel } from '../../model/form-model';
+import { flowModel } from '../../model/flow-model';
+import { Step } from '../../lib/types';
+
+const getThresholdOptions = (optionsAmount: number): DropdownOption<number>[] => {
```
Asmadek: Maybe move this two functions to utils?

src/renderer/widgets/CreateWallet/ui/MultisigWallet/SelectSignatoriesStep.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const signatories = useUnit(formModel.$signatories);
+  const wallets = useUnit(walletModel.$wallets);
+  const contacts = useUnit(contactModel.$contacts);
+
+  const hasEnoughSignatories = signatories.length >= 2;
+  const canContinue = hasOwnSignatory && hasEnoughSignatories && !multisigAlreadyExists;
+
+  console.log('<><> canContinue', canContinue);
+  console.log('<><> hasOwnSignatory', hasOwnSignatory);
+
```
Asmadek: do you need this logs?

Asmadek (changes requested):

src/renderer/shared/ui/Modals/BaseModal/BaseModal.tsx
Tbaut: `cnTw` behaved very weirdly in that case.. and my class (coming from `panelClass`) was actually never appended. Any hint?

src/renderer/widgets/CreateWallet/model/confirm-model.ts
Tbaut: This is a question I asked in the comments, whether the multisig creation should support proxy accounts or not. I may just remove this all, and start simple..

src/renderer/widgets/CreateWallet/ui/MultisigWallet/ConfirmationStep.tsx
Tbaut: same.

src/renderer/widgets/CreateWallet/ui/MultisigWallet/ConfirmationStep.tsx
Tbaut: same, not sure if this is supposed to be supported.

src/renderer/widgets/CreateWallet/ui/MultisigWallet/FlexibleMultisigWallet.tsx
Tbaut: This is all about the Flexible multisig, which is not implemented in this PR. Hence many things commented, before this is done.

src/renderer/widgets/CreateWallet/ui/MultisigWallet/FlexibleMultisigWallet.tsx
Tbaut: same here, this is for the flexible multisig

src/renderer/widgets/CreateWallet/ui/MultisigWallet/MultisigWalletWizard.tsx
Tbaut: same here, this will be used shortly, when I tackled the flexible multisig implementation. I discussed this with Yaroslav and I think it's better to just leave it there.

src/renderer/widgets/CreateWallet/ui/MultisigWallet/NameThresholdStep.tsx
Tbaut: Do we use utils for UI components elsewhere? I've taken these 2 functions from the previous component. I can move them to some utils if we usually do this.

src/renderer/widgets/CreateWallet/ui/MultisigWallet/SelectSignatoriesStep.tsx
Tbaut: those are stray, I'll remove.

src/renderer/widgets/CreateWallet/model/flow-model.ts
Tbaut: This would end up quite verbose though, I copied the behavior of the `AddProxyModal` widget, that also has more or less the same models, and they are named 'form-model', `confirm-model` etc.
Should I rename these `create-wallet-flow-model`, `create-wallet-form-model`...?

Tbaut:
I removed most comments. The multisig creation doesn't take into account proxies for now, and only left the commented logic for the flexible multisig wizard.

src/renderer/widgets/CreateWallet/model/confirm-model.ts
Tbaut: I ended up deleting this :+1:

src/renderer/widgets/CreateWallet/ui/MultisigWallet/ConfirmationStep.tsx
Tbaut: I ended up deleting this :+1:

src/renderer/widgets/CreateWallet/ui/MultisigWallet/ConfirmationStep.tsx
Tbaut: I ended up deleting this :+1:

src/renderer/shared/ui/Modals/BaseModal/BaseModal.tsx
Tbaut: I added it back and it worked!

pgolovkin:
> There are still quite some things to iron out, such as error management for the form submission, UI glitches, and I'm investigating an issue I found in the multisig indexer, but I'd like to have your opinion on things such as:
> 
> * do we need to support proxy signers
> * should I organize models differently (right now there's 1 for the flow, 1 for most of the data, 1 for the confirmation). I basically took example on the "add proxy" flow

Proxy signatories is not a part of this feature.
The UI should match [mockup](https://www.figma.com/proto/HgFuj8QZd0xqEbkG10DIfs/Flexible-multisig-%7C-Milestone-4?page-id=0:1&node-id=518-88080&viewport=73,491,0.03&t=pRppgwa1U82gQVms-1&scaling=min-zoom&starting-point-node-id=518:88080&show-proto-sidebar=1) because:
1. For multishard wallet, WalletConnect, Nova Wallet the user has to select network before selecting acounts
2. There should be an option to add signatory by address only (without contact book)

Tbaut:
May I receive reviews on the content of the PR @tuul-wq please?

Tbaut:
I made the following changes:
- base this PR on top of the matrix removal one https://github.com/novasamatech/nova-spektr/pull/1898 so that we can keep subjects isolated
- I made the change to have first the network selection + name then the signatories selection

Since I did not received any feeback on the effector logic yet, I'd like to start with that. While I'll get the UI to look closer to what the Figma wireframe looks like

src/renderer/widgets/CreateWallet/model/__tests__/signatory-model.test.ts
```diff
@@ -0,0 +1,69 @@
...
+      params: { index: 0, name: 'test', address: toAddress(signerWallet.accounts[0].accountId) },
+    });
+
+    expect(scope.getState(signatoryModel.$signatories).size).toEqual(2);
+  });
+
+  test('should correctly delete signatories', async () => {
+    const scope = fork();
+
+    expect(scope.getState(signatoryModel.$signatories).size).toEqual(0);
```
Tbaut, line 39: Do you guys have any idea how this can actually fail? When run alone, this test succeeds.
When run alone, the test succeeds, when the whole file is ran, it fails, and when observing the `$signatories` store, it's indeed not empty! 
How can I make sure that stores are reset between each test?

@Asmadek @johnthecat @sokolova-an

Asmadek: I think set it directly should work 

```ts
    const scope = fork({
      values: new Map().set(signatoryModel.$signatories, new Map([])),
    });
```

Tbaut: thank you so much, it solved it. I've a hard time understanding why/how it could be non empty though :)

pgolovkin:
@Tbaut I'm getting error when create a multisig wallet.
image

Seems that Signer Wallet is undefined
image

But I'm not able to select wallet on the screen where filling signatoies.
image

src/renderer/entities/transaction/lib/callDataDecoder.ts
```diff
@@ -361,6 +361,9 @@ export const useCallDataDecoder = (): ICallDataDecoder => {
         call: decoded.args[2].toHex(),
       };
     },
+    [TransactionType.REMARK]: (decoded): Record<string, any> => {
```
tuul-wq, line 365: 👍

src/renderer/widgets/CreateWallet/model/__tests__/form-model.test.ts
tuul-wq, line 1: Nice tests!

src/renderer/widgets/CreateWallet/model/flow-model.ts
```diff
@@ -0,0 +1,413 @@
...
+  {
+    chain: formModel.$createMultisigForm.fields.chain.$value,
+    isConnected: $isChainConnected,
+  },
+  ({ isConnected, chain }): Transaction | undefined => {
+    if (!chain || !isConnected) return undefined;
+
+    return {
+      chainId: chain.chainId,
+      address: toAddress(TEST_ACCOUNTS[0], { prefix: 42 }),
```
tuul-wq:
```suggestion
      address: toAddress(TEST_ACCOUNTS[0], { prefix: SS58_DEFAULT_PREFIX }),
```

src/renderer/widgets/CreateWallet/model/flow-model.ts
```diff
@@ -0,0 +1,413 @@
...
+    name: formModel.$createMultisigForm.fields.name.$value,
+    threshold: formModel.$createMultisigForm.fields.threshold.$value,
+    signatories: signatoryModel.$signatories,
+    chain: formModel.$createMultisigForm.fields.chain.$value,
+  },
+  fn: ({ signatories, chain, name, threshold }) => ({
+    name,
+    threshold,
+    chainId: chain.chainId,
+    signatories: sortBy(
```
tuul-wq: Let's add curly braces, `return` and dedicated variable for readability

src/renderer/widgets/CreateWallet/model/form-model.ts
```diff
@@ -0,0 +1,113 @@
...
+    const baseAccounts = filteredAccounts.filter((a) => accountUtils.isBaseAccount(a) && a.name);
+
+    return [...filteredAccounts, ...baseAccounts];
+  },
+);
+
+sample({
+  clock: $createMultisigForm.submit,
+  source: $createMultisigForm.$isValid,
+  filter: (isValid: boolean) => isValid,
```
tuul-wq: I assume `boolean` is not required

src/renderer/widgets/CreateWallet/model/signatory-model.ts
```diff
@@ -0,0 +1,53 @@
...
+      walletFn: (w) => !walletUtils.isWatchOnly(w) && !walletUtils.isMultisig(w),
+      accountFn: (a) => Array.from(signatories.values()).some((s) => toAccountId(s.address) === a.accountId),
+    })?.length,
+);
+
+sample({
+  clock: signatoriesChanged,
+  source: $signatories,
+  fn: (signatories, { index, name, address }) => {
+    signatories.set(index, { name, address });
```
tuul-wq: Modifying original store is not a good idea

src/renderer/widgets/CreateWallet/ui/MultisigWallet/FlexibleMultisigWallet.tsx
```diff
@@ -49,7 +46,7 @@ export const MultiChainMultisigWallet = ({ isOpen, onClose, onComplete, onBack }
   useEffect(() => {
     if (isOpen && !isModalOpen) {
       toggleIsModalOpen();
-      createMultisigWalletModel.events.reset();
+      // flowModel.events.reset();
```
tuul-wq: Serves some purpose?

src/renderer/widgets/CreateWallet/ui/MultisigWallet/NameNetworkSelection.tsx
```diff
@@ -0,0 +1,89 @@
...
+          />
+          <FootnoteText className="mt-2 text-text-tertiary">
+            {t('createMultisigAccount.networkDescription')}
+          </FootnoteText>
+        </div>
+        <div className="flex justify-end items-center mt-auto">
+          <Button
+            key="create"
+            disabled={!canContinue}
+            onClick={() => flowModel.events.stepChanged(Step.SIGNATORIES_THRESHOLD)}
```
tuul-wq, line 82: This is triggered in parallel to submit event?

tuul-wq (commented):
Few suggestions & questions

src/renderer/widgets/CreateWallet/ui/MultisigWallet/components/Signatory.tsx
```diff
@@ -0,0 +1,111 @@
...
+    signatoryModel.events.signatoriesChanged({
+      index,
+      name: newName,
+      address,
+    });
+  };
+
+  const onAddressChange = (newAddress: string) => {
+    try {
+      validateAddress(newAddress);
```
tuul-wq: `validateAddress` can throw an error? 
Why not to use `validateAddress` from `utils/address`?

src/renderer/widgets/CreateWallet/lib/create-multisig-utils.ts
```diff
@@ -0,0 +1,34 @@
...
+export const createMultisigUtils = {
+  isSignStep,
+  isSubmitStep,
+  isConfirmStep,
+  isNameNetworkStep,
+  isSignatoriesThresholdStep,
+  isNotFirstStep,
+};
+
+function isConfirmStep(step: Step) {
```
tuul-wq: @Asmadek I remember we had a talk between me, you and @sokolova-an about some utility function instead of step checker functions. Have anything been done regarding it and @Tbaut should use that function/approach?

src/renderer/widgets/CreateWallet/model/form-model.ts
```diff
@@ -0,0 +1,113 @@
...
+          validator: (name) => name !== '',
+          errorText: 'createMultisigAccount.disabledError.emptyName',
+        },
+      ],
+    },
+  },
+  validateOn: ['submit'],
+});
+
+const formSubmitted = createEvent<void>();
```
tuul-wq: Native event to call after form is submitted - `$form.formValidated`
It doesn't cover your case?

src/renderer/widgets/CreateWallet/ui/MultisigWallet/MultisigWallet.tsx
```diff
@@ -1,66 +1,98 @@
...
 
-  const selectWalletType = (type: MultisigWalletType) => {
-    setWalletType(type);
+  useEffect(() => {
+    flowModel.events.callbacksChanged({ onComplete });
+  }, [onComplete]);
 
-    setStep(Step.CREATE_WALLET);
-  };
+  useEffect(() => {
```
tuul-wq: Maybe this hook can help - `useModalClose`?

src/renderer/widgets/CreateWallet/ui/MultisigWallet/MultisigWallet.tsx
```diff
@@ -1,66 +1,98 @@
...
-    setTimeout(params?.complete ? onComplete : onClose, DEFAULT_TRANSITION);
+    setTimeout(params?.complete ? onComplete : params?.closeAll ? onClose : noop, DEFAULT_TRANSITION);
   };
 
-  const goBack = () => {
-    setWalletType(undefined);
-    setStep(Step.SELECT_WALLET_TYPE);
-  };
+  if (createMultisigUtils.isSubmitStep(activeStep))
+    return <OperationSubmit isOpen={isModalOpen} onClose={closeMultisigModal} />;
```
tuul-wq:
```suggestion
  if (createMultisigUtils.isSubmitStep(activeStep)) {
      return <OperationSubmit isOpen={isModalOpen} onClose={closeMultisigModal} />;
  }
```

src/renderer/widgets/CreateWallet/ui/MultisigWallet/NameNetworkSelection.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const { t } = useI18n();
+
+  const chains = useUnit(networkModel.$chains);
+  const {
+    fields: { name, chain },
+    submit,
+  } = useForm(formModel.$createMultisigForm);
+
+  const chainOptions = getChainOptions(Object.values(chains));
+  const canContinue = name.isValid;
```
tuul-wq: This variable is probably not worth defining for a single use case `disabled={!canContinue}`

tuul-wq (commented):
Part 2

src/renderer/widgets/CreateWallet/lib/create-multisig-utils.ts
Asmadek: [isStep](https://github.com/novasamatech/nova-spektr/blob/a59f223e14813dab5ce2849650b0478c7cb211c0/src/renderer/shared/lib/utils/step.ts#L16) function can be used

src/renderer/widgets/CreateWallet/lib/create-multisig-utils.ts
```diff
@@ -0,0 +1,34 @@
...
+function isSignStep(step: Step) {
+  return step === Step.SIGN;
+}
+
+function isSubmitStep(step: Step) {
+  return step === Step.SUBMIT;
+}
+
+function isNotFirstStep(step: Step) {
+  return step > Step.NAME_NETWORK;
```
Asmadek: Better to use `!==` instead `>` for such cases

src/renderer/widgets/CreateWallet/ui/MultisigWallet/components/Signatory.tsx
```diff
@@ -0,0 +1,111 @@
...
+        name,
+        address: newAddress,
+      });
+    } catch {
+      // do nothing wrong address
+      setAddress('');
+    }
+  };
+
+  // const options = proxyAccounts.map((proxyAccount) => {
```
Asmadek: add todo when it should be uncommented or remove it

src/renderer/widgets/CreateWallet/ui/MultisigWallet/components/SelectSignatories.tsx
```diff
@@ -1,292 +1,105 @@
...
-
-  const isDisabledContact = (contact: ExtendedContact): boolean => {
-    const isThisContact = selectedContactsList.includes(contact);
-    const isSameContactSelected = selectedContactsList.some((c) => c.accountId === contact.accountId);
-    const isSameWalletSelected = selectedWalletsList.some((w) => w.accountId === contact.accountId);
-
-    return !isThisContact && (isSameContactSelected || isSameWalletSelected);
+  }, [contacts.length, chain]);
+
+  //       return !isEthereumAccountId(c.accountId);
```
Asmadek: add todo when it should be uncommented or remove it

src/renderer/widgets/CreateWallet/ui/MultisigWallet/NameNetworkSelection.tsx
```diff
@@ -0,0 +1,89 @@
...
+            selectedId={chain.value.chainId.toString()}
+            options={chainOptions}
+            onChange={({ value }) => chain.onChange(value)}
+          />
+          <FootnoteText className="mt-2 text-text-tertiary">
+            {t('createMultisigAccount.networkDescription')}
+          </FootnoteText>
+        </div>
+        <div className="flex justify-end items-center mt-auto">
+          <Button
```
Asmadek, line 79: looks like `type="submit"` should be here to trigger submit

Asmadek (approved):

src/renderer/widgets/CreateWallet/model/form-model.ts
Tbaut: I get an error with isValid being implicitely of type any, although `$createMultisigForm.$isValid` is of type `StoreWritable<boolean>`. Maybe something specific with effector-forms?

src/renderer/widgets/CreateWallet/model/signatory-model.ts
Tbaut: I originally thought about having a array, instead of a Map. Now the thing is that arrays have no guarantee of keeping in order (although tbh I've never had issues with it). Maps do. However, this adds a lot of complexity everywhere, since looping through Maps require `Array.from`.

Another complexity that it adds, is that Maps don't trigger a re-rendering of the stores. The only way to trigger is to create a new Map from the original one, modify it, and assign the store to the new map (a new object address).

Do you think I should rather go with arrays, and take the risk of having the array order changed?

src/renderer/widgets/CreateWallet/ui/MultisigWallet/FlexibleMultisigWallet.tsx
Tbaut: stray comment, I'll remove. The whole file isn't used at the moment.

src/renderer/widgets/CreateWallet/ui/MultisigWallet/NameNetworkSelection.tsx
Tbaut: Unless I missed something, there's no submit event triggered here. The flow is cut into several steps, and the submission of the form is only happening in the next step, more specifically, here: https://github.com/novasamatech/nova-spektr/blob/bff125e955be3bada5d0a90250661b40ee9cd2f4/src/renderer/widgets/CreateWallet/ui/MultisigWallet/SelectSignatoriesThreshold.tsx#L103

src/renderer/widgets/CreateWallet/ui/MultisigWallet/components/Signatory.tsx
Tbaut: I didn't realize you had one in the utils. Using it now (I'll push after) and it works well, no need to use try/catch any more.

src/renderer/widgets/CreateWallet/ui/MultisigWallet/components/Signatory.tsx
Tbaut: Absolutely, sorry about that, there are still a lot of comments, because for instance I don't know yet how to cater for proxy accounts, or shard accounts etc. I'll remove them.

src/renderer/widgets/CreateWallet/ui/MultisigWallet/NameNetworkSelection.tsx
Tbaut: The submit is triggered in the next step. I have 1 single form, for 2 steps:
1 - Name and network of the multisig
2 - Signatory and threshold selection

only the 2nd is actually submitting.

Asmadek: Should `onSubmit={submitForm}` be in the form component?

src/renderer/widgets/CreateWallet/lib/create-multisig-utils.ts
Tbaut: Thanks, I re-declared one since my steps are different, but used this approach none the less.

Tbaut:
@pgolovkin the issue you encountered is probably because the first account is supposed to be one of your own account. This is not something that is enforced yet, and the selection of your wallet doesn't work yet. You can type in some address though, and if you do so, the first account should be yours, then the error disapears, you can type in other signatories etc. Then the signing should work (it did for me, I just tested). The submission to the blockchain failed bc of "wasm trap" but this is not linked to this PR, it's linked to the v12 api unless I missed something.

pgolovkin:
> @pgolovkin the issue you encountered is probably because the first account is supposed to be one of your own account. This is not something that is enforced yet, and the selection of your wallet doesn't work yet. You can type in some address though, and if you do so, the first account should be yours, then the error disapears, you can type in other signatories etc. Then the signing should work (it did for me, I just tested). The submission to the blockchain failed bc of "wasm trap" but this is not linked to this PR, it's linked to the v12 api unless I missed something.

thanks for explanation!

pgolovkin:
@Tbaut now I'm getting `TooFewSignatories` error. But anyway signing works!
https://westend.subscan.io/extrinsic/21911935-2

Also I'm still getting error when select the account from the Multishard or Polkadot Vault wallet. Seems that there is a problem with multishard and Polkadot Vault wallets (where the wallet may contain may accounts on a single chain).

Tbaut:
I'll take care of this now that I know there doesn't seem to be big changes incoming in the architecture. From your Tx, I can see there are indeed 0 signatories, as `other_signatories` is null.

src/renderer/widgets/CreateWallet/model/form-model.ts
Tbaut: oh, I didn't know about this one. I changed it.

src/renderer/widgets/CreateWallet/ui/MultisigWallet/NameNetworkSelection.tsx
Tbaut: I ended up removing any submission at this stage, since they aren't needed.

src/renderer/widgets/CreateWallet/lib/create-multisig-utils.ts
```diff
@@ -0,0 +1,20 @@
...
+  isNotFirstStep,
+};
+
+/**
+ * Check if current step equal to the target step
+ * @param step  current step
+ * @param targetStep target step
+ * @return {boolean}
+ */
+function isStep(step: Step, targetStep: Step): boolean {
```
Asmadek: We already have general `isStep` function, maybe use it?

src/renderer/widgets/CreateWallet/model/signatory-model.ts
Asmadek: Right, create new map -> modify new map -> return new map

src/renderer/widgets/CreateWallet/ui/MultisigWallet/FlexibleMultisigWallet.tsx
```diff
@@ -60,39 +52,26 @@ export const MultiChainMultisigWallet = ({ isOpen, onClose, onComplete, onBack }
...
   const closeMultisigModal = (params: { complete?: boolean; closeAll?: boolean } = { closeAll: true }) => {
     toggleIsModalOpen();
 
     setTimeout(params?.complete ? onComplete : params?.closeAll ? onClose : noop, DEFAULT_TRANSITION);
   };
 
-  const submitHandler = (args: { name: string; threshold: number; creatorId: HexString }) => {
-    toggleResultModal();
-    setName(args.name);
+  // const submitHandler = (args: { name: string; threshold: number; creatorId: HexString }) => {
```
Asmadek: commented code

src/renderer/widgets/CreateWallet/ui/MultisigWallet/FlexibleMultisigWallet.tsx
```diff
@@ -134,20 +106,9 @@ export const MultiChainMultisigWallet = ({ isOpen, onClose, onComplete, onBack }
...
-            isActive={activeStep === Step.INIT}
-            wallets={wallets}
-            accounts={wallets.map((wallet) => wallet.accounts).flat()}
-            contacts={contacts}
-            onSelect={(wallets, contacts) => {
-              setSignatoryWallets(wallets);
-              setSignatoryContacts(contacts);
-            }}
-          />
+          {/* {activeStep === Step.CONFIRMATION && <ConfirmationStep />} */}
```
Asmadek: commented code. Check comment line above, is is actual if this code commented?

src/renderer/widgets/CreateWallet/ui/MultisigWallet/MultisigWalletWizard.tsx
```diff
@@ -0,0 +1,37 @@
...
+export const MultisigWalletWizard = ({ isOpen, onClose, onComplete }: Props) => {
+  const [isModalOpen, toggleIsModalOpen] = useToggle(isOpen);
+
+  const closeMultisigModal = (params?: { complete: boolean }) => {
+    toggleIsModalOpen();
+
+    setTimeout(params?.complete ? onComplete : onClose, DEFAULT_TRANSITION);
+  };
+
+  // const goBack = () => {
```
Asmadek: commented code

src/renderer/widgets/CreateWallet/model/form-model.ts
tuul-wq: Alright, let's leave it as is

src/renderer/widgets/CreateWallet/ui/MultisigWallet/NameNetworkSelection.tsx
tuul-wq: Got it, thanks

tuul-wq (approved):

src/renderer/widgets/CreateWallet/lib/create-multisig-utils.ts
Tbaut: But the steps supported by `isStep` are different. Do you really want me to add my specific steps to the somewhat generic `Step` type? 

```
export const enum Step {
  NONE,
  INIT,
  CONFIRM,
  SIGN,
  SUBMIT,
  BASKET,
}
```

src/renderer/widgets/CreateWallet/ui/MultisigWallet/FlexibleMultisigWallet.tsx
Tbaut: This whole file isn't used right now. I kept it here because it'll be used soon. But if you prefer I can totally remove it either.

src/renderer/widgets/CreateWallet/model/signatory-model.ts
Tbaut: ah got it, I made the change.

src/renderer/widgets/CreateWallet/ui/MultisigWallet/FlexibleMultisigWallet.tsx
Tbaut: I removed the whole file

src/renderer/widgets/CreateWallet/ui/MultisigWallet/MultisigWalletWizard.tsx
Tbaut: I removed the whole file, it wasn't used.

src/renderer/widgets/CreateWallet/ui/MultisigWallet/FlexibleMultisigWallet.tsx
Tbaut: I removed the whole file

src/renderer/widgets/CreateWallet/lib/create-multisig-utils.ts
Asmadek: Yes, we decided to do it in such way

Tbaut: I made the `isStep` generic with types, so that anyone can check with their own step. https://github.com/novasamatech/nova-spektr/pull/1800/commits/429570a7754570c9f6250966e9417b3ba8070a6c

This should make things better.

pgolovkin:
The list of issues to be fixed:
1. The "multisig" in chains.json must be added only for networks where the multisig indexer is running. So that the list of available networks for the multisig will be limited
image

2. Validation error should be added only when the user proceeds. Now it's opened when the user gets to the 2nd screen of the Create multisig flow 
image

3. There is no additional cost (network fee + multisig deposit) that the user has to pay for the `system_remark` call on the 2nd screen of the Create multisig flow. Figma mockup [link](https://www.figma.com/design/HgFuj8QZd0xqEbkG10DIfs/Flexible-multisig-%7C-Milestone-4?node-id=369-58419&t=h20RjewkkjpFoFmn-4)

4. Confirmation screen doesn't match the [mockup]
(https://www.figma.com/design/HgFuj8QZd0xqEbkG10DIfs/Flexible-multisig-%7C-Milestone-4?node-id=528-42184&t=h20RjewkkjpFoFmn-4).

5. Agree, that operation should not be added into the basket.

6. When the user successfully signs `system_remark` operation then the new wallet has to be selected and perations page has to be opened. As I remember it's a common flow for multisig operations, please check it.

src/renderer/pages/Basket/lib/prepareTransactions.ts
```diff
@@ -654,3 +655,28 @@ async function prepareRevokeDelegationTransaction({
...
+async function prepareRemarkTransaction({ transaction, wallets, chains, apis, feeMap }: DataParams) {
+  const { chain, account, fee } = await getTransactionData(transaction, feeMap, apis, chains, wallets);
+
+  return {
+    id: transaction.id,
+    chain,
+    asset: chain.assets[0],
+    shards: [account],
+    destination: transaction.coreTx.args.dest,
+    remark: '',
```
Asmadek: As I clearly understand from this file `src/renderer/entities/transaction/lib/extrinsicService.ts` remark doesn't have destination param. And why remark param is empty?

src/renderer/widgets/CreateWallet/ui/MultisigWallet/FlexibleMultisigWallet.tsx
Asmadek: Don't forget to remove this file if it's useless, or add comment in code if it can be used in the future

Asmadek (approved):

src/renderer/pages/Basket/lib/prepareTransactions.ts
pgolovkin: @Asmadek I've created the multisig with this branch and got this transaction https://westend.subscan.io/extrinsic/22813617-2 

I don't know how, but the text for remark was saved on-chain.
image

src/renderer/widgets/CreateWallet/ui/MultisigWallet/FlexibleMultisigWallet.tsx
Tbaut: oh thanks, somehow I forgot. 

edit: done in [f0243ca](https://github.com/novasamatech/nova-spektr/pull/1800/commits/f0243ca8b6474d5433a2ae335469b123ce180c7c)

src/renderer/pages/Basket/lib/prepareTransactions.ts
Tbaut: Good point @Asmadek this is not needed since it will never be in the basket. I removed it and excluded to `TransactionData` object type

Tbaut:
@pgolovkin I addressed your comments in the last commit.
1. I now take the proxy url when the network has the `multisig` option
2. The alerts are only displayed after the user made a mistake and clicked on next 
3. I now display all additional costs on all screens
4. My screen is now very close to the mock up. The only difference is that it's larger. The reason I think it's better this way, is so that I can re-utilize the nice "Sign with XYZ" button. And this together with the "additional costs" really doesn't look good on small modal, e.g
image
6. 
> When the user successfully signs system_remark operation then the new wallet has to be selected and perations page has to be opened. As I remember it's a common flow for multisig operations, please check it.

I need help, maybe @Asmadek @sokolova-an or @johnthecat can tell me how to do that.

pgolovkin:
> 4\. https://www.figma.com/design/HgFuj8QZd0xqEbkG10DIfs/Flexible-multisig-%7C-Milestone-4?node-id=528-42184&t=h20RjewkkjpFoFmn-4

let's clarify with designer.

pgolovkin:
> @pgolovkin I addressed your comments in the last commit.
> 
> 1. I now take the proxy url when the network has the `multisig` option
> 2. The alerts are only displayed after the user made a mistake and clicked on next
> 3. I now display all additional costs on all screens
> 4. My screen is now very close to the mock up. The only difference is that it's larger. The reason I think it's better this way, is so that I can re-utilize the nice "Sign with XYZ" button. And this together with the "additional costs" really doesn't look good on small modal, e.g
>    image
> 
> > When the user successfully signs system_remark operation then the new wallet has to be selected and perations page has to be opened. As I remember it's a common flow for multisig operations, please check it.
> 
> I need help, maybe @Asmadek @sokolova-an or @johnthecat can tell me how to do that.

1. ok
2. ok
3. ok
4. let's clarify with designer

Tbaut:
I've adapted the confirmation step to the [new design](https://www.figma.com/design/HgFuj8QZd0xqEbkG10DIfs/Flexible-multisig-%7C-Milestone-4?node-id=952-47351&t=FJzsx86u7iVtwMR0-4).
The only difference I kept on purpose is the conversion in $ value present on my implementation but not on the design, because I think it brings value to users
image

pgolovkin:
@Tbaut looks great!
 Could you please check how the following scenario works:
1. The user selects more then 1 accounts that the user owns (1 is polkadot vault, 1 is nova wallet)
2. The user proceed 
3. Which account will be selected for signing?

Also please check with Sergy which popup has to be used for showing signatories on the confirmation screen. For example for staking it looks like on the screenshot
image

src/renderer/entities/transaction/ui/MultisigCreationFees.tsx/MultisigCreationFees.tsx
sokolova-an, line 1: That looks like a pretty specific file and in entities transaction we keep common things for all transactions. Since it's been used only in CreateWallet maybe move it to CreateWallet/ui/MultisigWallet/components?

sokolova-an (approved):

src/renderer/entities/transaction/ui/MultisigCreationFees.tsx/MultisigCreationFees.tsx
Tbaut: Sure, thanks for the suggestion. You mean widget right? I did it in [915c924](https://github.com/novasamatech/nova-spektr/pull/1800/commits/915c924cbf58525f182f7cf161106cdcd0ef2151)

Tbaut:
I implemented the modal for signatories instead of a tooltip: 
image

Tbaut:
> The user selects more then 1 accounts that the user owns (1 is polkadot vault, 1 is nova wallet)
    The user proceed
    Which account will be selected for signing?

To answer here, what I answered in tg:
For now, the first account selected has to be an account that is owned, and this is the default signer. Let's discuss the design if you want me to add/change something.

Tbaut:
I added the signer selection, with a "Back" button and without the token info for now. I need some more info to manage the multi-shard accounts. I pinged @johnthecat to have a chat.

src/renderer/widgets/CreateWallet/ui/MultisigWallet/components/Signer.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+interface Props {
+  onSubmit: (event: FormEvent, accountId: AccountId) => void;
+  accountId: AccountId;
+  name?: string;
+  chain: Chain;
+}
+
+export const Signer = ({ accountId, name, onSubmit, chain }: Props) => {
+  //   const balance = useBalance({
```
Tbaut: I have an issue here, where only the balance of the selected account is displayed. If I have 2 possible signers, and I want to display the balance of each of them, the balance is `undefined` for the account that isn't selected.

I tried 2 approaches, one using `useBalance`, the other one fetching the balance directly from `balanceModel`. Both ways failed in the same way.

Can you help @johnthecat @sokolova-an ?

Here is an example, I own both accounts, they both have > 0 WND but only the one that is selected (in blue) has a balance.
image

sokolova-an: That's a tricky case, as I remember we don't load balances for all wallets at once, so if you haven't ever opened the wallet - its accounts won't be in the balances storage and you can't get access to the value with balanceModel in this case.
Please check, will it work if before that, you open the second wallet and wait until the balances are loaded?

In any case, I guess, you will have to manually run update or populate balances for the other wallet, because the store data can be outdated anyway. The main logic about balance stores in `balanceSubModel`

Tbaut: Thank you so much for the pointers. Even when I had the other account open before the balance was undefined. I subscribed manually to the balance now, and this fixed it. Thanks again.

Tbaut:
Now showing the Wallet rather than the accounts @pgolovkin in the signer selection. 
image

pgolovkin (approved):

pgolovkin:
@Tbaut looks great! Thanks!

johnthecat (approved):

PR #1836, opened 2024-06-05
Feat: confirm basket transaction

### LINKED ISSUE #1811: Basket. Confirmation screen for single operation
* Show confirm screen for operation when click on it or select only one and sign 
* List of confirm screens 
* Transfer/xcm transfer
* Stacking operations 
* Manage proxy operations

Asmadek:
* Move confirm screens and modals to OperationsConfirm folder
* Create sign operation component with confirm step for basket transactions


pgolovkin:
@Asmadek if the operation is not Valid then the user should not be able to open the confirmation screen.

src/renderer/pages/Basket/lib/sign-operations-utils.ts
```diff
@@ -0,0 +1,24 @@
+import { Step } from '../types';
+
+export const signOperationsUtils = {
+  isNoneStep,
+  isConfirmStep,
+  isSignStep,
+  isSubmitStep,
+};
+
+function isNoneStep(step: Step): boolean {
```
tuul-wq, line 10: What if we ask chatGpt how can we simplify all similar functions regarding `Step`
Not cool to have so much duplication

src/renderer/pages/Basket/model/basket-page-model.ts
```diff
@@ -146,13 +149,20 @@ sample({
   target: $invalidTxs,
```
tuul-wq, line 149: So it works fine even without different `Map` reference?

src/renderer/pages/Basket/model/sign-operations-model.ts
```diff
@@ -0,0 +1,853 @@
...
+  fn: () => Step.CONFIRM,
+  target: stepChanged,
+});
+
+// Restake
+
+sample({
+  clock: restakeDataPreparationStarted,
+  source: {
+    transactions: $transactions,
```
tuul-wq: I guess it can be shortened with `combine` to reduce duplication

src/renderer/pages/Basket/model/sign-operations-model.ts
```diff
@@ -0,0 +1,853 @@
...
+  },
+  target: prepareRestakeTransactionDataFx,
+});
+
+sample({
+  clock: prepareUnstakeTransactionDataFx.doneData,
+  target: unstakeConfirmModel.events.formInitiated,
+});
+
+sample({
```
tuul-wq: Can be combined with other `doneData` operations

tuul-wq (approved):

src/renderer/pages/Basket/model/basket-page-model.ts
Asmadek: Not clearly understand question, I create new `Map` here

sokolova-an (approved):

PR #1903, opened 2024-06-28
feat: select all checkbox

### LINKED ISSUE #1816: Basket. Select all button
Select all available operations
Show amount of selected operations and amount of all operations

Asmadek:
* use checkbox with label for select all
* correct margins

### DIFF
--- a/src/renderer/shared/api/translation/locales/en.json
+++ b/src/renderer/shared/api/translation/locales/en.json
@@ -102,6 +102,8 @@
     },
     "searchPlaceholder": "Search",
     "selectAll": "Select all",
+    "selectedStatus_one": "{ selected } out of { count } operation selected",
+    "selectedStatus_other": "{ selected } out of { count } operations selected",
     "signButton": "Sign",
     "signOperations": {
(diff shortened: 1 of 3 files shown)


pgolovkin (approved):

src/renderer/shared/api/translation/locales/en.json
```diff
@@ -102,6 +102,8 @@
     },
     "searchPlaceholder": "Search",
     "selectAll": "Select all",
+    "selectedStatus_one": "{ selected } out of { count } operation selected",
+    "selectedStatus_other": "{ selected } out of { count } operations selected",
```
sokolova-an, line 106: Where exactly are we using them? I see only `basket.selectedStatus` there

sokolova-an (approved):

src/renderer/shared/api/translation/locales/en.json
Asmadek: it's i18next [plurals feature](https://www.i18next.com/translation-function/plurals), it use `_one` suffix if `count` is 1, and `_other` if other

PR #1924, opened 2024-07-03
fix: don't open transfer modal

Asmadek:
fix: don't open transfer modal
fix: reset selected txs
fix: not open sign modal after new tx added 
fix: validation stake more
fix: fee calculation 
fix: correct order for errors 
fix: remove basket when remove wallet

### DIFF
--- a/src/renderer/pages/Basket/model/basket-page-model.ts
+++ b/src/renderer/pages/Basket/model/basket-page-model.ts
@@ -365,6 +365,13 @@ sample({
   target: stepChanged,
 });
 
+sample({
+  clock: walletModel.events.walletRemovedSuccess,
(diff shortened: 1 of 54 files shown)


src/renderer/pages/Basket/model/basket-page-model.ts
```diff
@@ -341,6 +353,18 @@ sample({
   target: basketModel.events.transactionsRemoved,
 });
 
+sample({
+  clock: selectedTxsReset,
+  fn: () => [],
+  target: $selectedTxs,
```
sokolova-an: you can use`$selectedTxs.reinit() `

sokolova-an (approved):

PR #1931, opened 2024-07-04
chore: Upgrade headlessui

Tbaut:
So as to prevent having to hack around the immediate opening of the options described here: https://github.com/tailwindlabs/headlessui/discussions/1205

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
   },
   "dependencies": {
     "@apollo/client": "^3.7.1",
-    "@headlessui/react": "^1.7.17",
+    "@headlessui/react": "^2.1.1",
     "@matrix-org/olm": "3.2.15",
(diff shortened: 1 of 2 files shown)


Tbaut:
Any chance you can take a quick look @sokolova-an ? the test failing should be familiar, and I don't think it's related to the upgrade.

sokolova-an:
> Any chance you can take a quick look @sokolova-an ? the test failing should be familiar, and I don't think it's related to the upgrade.

This is weird, locally this test is running fine before the new lib's version, and this one we haven't touched for a year + it doesn't rely on headless lib

Tbaut:
I'll forget about this for now, the app is breaking with errors that don't make sense to me. I'll rather use a hack with the version we have right now

PR #1943, opened 2024-07-10
feat: openGov add locks

### LINKED ISSUE #1784: OpenGov. Governance page. Locks
As a user I want to see my locked tokens

To do:

- [ ] Locked value (for multishard account it must be sum of locks)
- [ ] Unlockable state
- [ ] Loading state
- [ ] Button that leads to "next step"

[Figma link](https://www.figma.com/design/ylvl7Ox25UuZj73Rn85DK7/OpenGov?node-id=326-51756&t=RafEPHCQCBylgBPp-4)
[Figma link to Locked popup](https://www.figma.com/design/ylvl7Ox25UuZj73Rn85DK7/OpenGov-voting?node-id=370-15772&t=fKdMbC3b3otNnBfZ-4)

### LINKED ISSUE #1845: Step checker util
To remove lots of duplication regarding step checking, we came up with a generic code:
```javascript
function isStep<T>(step: T, targetStep: T): boolean {
  return step === targetStep;
}
```

sokolova-an:
- Add locks panel;
- Add governance lock modal with lock statuses;
- Add common function  `isStep` to remove duplication;
- Update UI for delegation panel;

The next step for the unlock modal will be added in the next PR


src/renderer/features/governance/lib/unlock.ts
```diff
@@ -0,0 +1,24 @@
+import { Address } from '@shared/core';
+import { Chunks, UnlockChunk } from '@shared/api/governance';
+
+export const unlockService = {
+  getSecondsDuratonToBlock,
+  filterClaims,
+};
+
+function getSecondsDuratonToBlock(timeToBlock: number): number {
```
Asmadek: looks like good util function for `shared/lib/utils/substrate.ts`

src/renderer/features/governance/lib/unlock.ts
```diff
@@ -0,0 +1,24 @@
...
+  const currentTime = new Date().getTime();
```
Asmadek: What do you think to use `Date.now()` here? It's a little bit faster and more meaningful

src/renderer/features/governance/model/locks.ts
```diff
@@ -1,77 +1,68 @@
...
-  filter: (isLocksRequested, api) => !isLocksRequested && Boolean(api),
-  fn: (_, api) => api!,
-  target: getClassLocksFx,
+  clock: [networkSelectorModel.$governanceChainApi, walletModel.$activeWallet],
+  source: {
+    chain: networkSelectorModel.$governanceChain,
+    api: networkSelectorModel.$governanceChainApi,
+    wallet: walletModel.$activeWallet,
+  },
+  filter: ({ chain, api, wallet }) => !!chain && !!api && !!wallet,
```
Asmadek, line 33: I think we decide to use `Boolean` instead of `!!`, but it can be discussed

src/renderer/features/governance/model/locks.ts
```diff
@@ -1,77 +1,68 @@
...
 
 sample({
-  clock: getClassLocksFx.doneData,
+  clock: getTrackLocksFx.doneData,
   fn: (trackLocks) => {
-    const locks = Object.values(trackLocks).reduce<BN>((acc, lock) => (lock.gt(acc) ? lock : acc), BN_ZERO);
+    let maxLockTotal = BN_ZERO;
+    for (const address in trackLocks) {
+      const lock = trackLocks[address];
+      const totalLock = Object.values(lock).reduce<BN>((acc, lock) => (lock.gt(acc) ? lock : acc), BN_ZERO);
```
Asmadek: `BN.max(lock, acc)` should work here. Or we should create utility function to get max value from array of BN

src/renderer/features/governance/model/unlock.ts
```diff
@@ -0,0 +1,144 @@
...
+  clock: [referendumModel.events.requestDone, locksModel.$trackLocks],
+  source: {
+    api: networkSelectorModel.$governanceChainApi,
+    tracks: tracksAggregate.$tracks,
+    trackLocks: locksModel.$trackLocks,
+    voting: votingAggregate.$currentWalletVoting,
+    referendums: referendumModel.$referendums,
+    chain: networkSelectorModel.$governanceChain,
+  },
+  filter: ({ api, voting, trackLocks, chain }) => !!api && !!chain && !!voting && !!trackLocks,
```
Asmadek: Same comment about !!

src/renderer/features/governance/model/unlock.ts
```diff
@@ -0,0 +1,144 @@
...
+sample({
+  clock: flowStarted,
+  fn: () => Step.INIT,
+  target: stepChanged,
+});
+
+sample({
+  clock: getClaimScheduleFx.finally,
+  fn: () => false,
+  target: $isLoading,
```
Asmadek: `$isLoading` is true only for first loading, am I right? Because doesn't reset for it. If it should be reseted maybe`getClaimScheduleFx.pending` will work here

src/renderer/features/governance/model/locks.ts
```diff
@@ -1,77 +1,68 @@
...
-  clock: getClassLocksFx.finally,
+  clock: getTrackLocksFx.finally,
   fn: () => false,
   target: $isLoading,
 });
 
 export const locksModel = {
-  $maxLock,
   $asset,
   $isLoading,
```
Asmadek, line 65: Looks like `$isLoading: getTrackLocksFx.pending` enough here? But I am not sure for 100%

Asmadek (changes requested):

src/renderer/features/governance/model/locks.ts
johnthecat: I've started to use `!!` in governance code because of shorter syntax, but yeah, we can discuss that

src/renderer/features/governance/model/locks.ts
sokolova-an: doesn't look good this way here, so better keep `$isLoading`

Asmadek (approved):

PR #1969, opened 2024-07-17
feat: add unlockable sum

sokolova-an:
Add a total unlockable sum for multiple addresses

### DIFF
--- a/src/renderer/features/governance/model/unlock.ts
+++ b/src/renderer/features/governance/model/unlock.ts
@@ -113,6 +113,7 @@ sample({
   fn: (claimSchedule) => {
     const unlockable = claimSchedule.reduce((acc, claim) => {
       if (claim.type !== UnlockChunkType.CLAIMABLE) return acc;
+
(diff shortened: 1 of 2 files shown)


johnthecat (approved):

src/renderer/features/governance/model/unlock.ts
```diff
@@ -21,6 +21,7 @@ const txSaved = createEvent();
 
 const $step = restore<Step>(stepChanged, Step.NONE);
 const $claimSchedule = createStore<UnlockChunk[]>([]).reset(walletModel.$activeWallet);
+const $totalUnlock = createStore<BN | null>(null);
```
Asmadek: what if use `BN_ZERO` as default? It will helps to skip `return unlockable.isZero() ? null : unlockable;` and use `!unlockable.isZero()` instead of `!!totalUnlock` in `UnlockInfo.tsx`

Just suggestion

Asmadek (approved):

PR #2007, opened 2024-07-31
feat: unlock sign operation

sokolova-an:
- Add completed flow for sign unlock operation with a single wallet/account (nova wallet or vault)
- Lock and unlock sum updated after submitting the operation
- Moved all prepareTransaction functions into a separate file

### DIFF
--- a/src/renderer/features/governance/aggregates/unlock.ts
+++ b/src/renderer/features/governance/aggregates/unlock.ts
@@ -1,11 +1,18 @@
-import { createEvent, restore, sample } from 'effector';
-import { or, spread } from 'patronum';
+import { combine, createEvent, createStore, restore, sample } from 'effector';
+import { delay, or, spread } from 'patronum';
 
+import { type Transaction } from '@/shared/core';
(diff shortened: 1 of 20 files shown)


src/renderer/features/governance/aggregates/unlock.ts
```diff
@@ -60,6 +98,96 @@ sample({
...
+  clock: submitModel.$submitStep,
+  source: {
+    chunks: unlockModel.$claimSchedule,
+    unlockData: $unlockData,
+  },
+  filter: ({ unlockData }, { step }) => !!unlockData && submitUtils.isSuccessStep(step),
+  fn: ({ chunks, unlockData }) => {
+    return chunks.filter(
+      (chunk) =>
+        chunk.type === UnlockChunkType.CLAIMABLE && unlockData!.shards.some((shard) => shard.address !== chunk.address),
```
Asmadek: I think here better to add braces and return (here we have `=>` and return value on two different lines)

Asmadek (approved):

PR #2032, opened 2024-08-06
feat: add unlock multishard & basket

### LINKED ISSUE #1933: OpenGov. Governance page. Unlock tokens
- [x] Unlock confirm screen for all wallet types. [Figma link](https://www.figma.com/design/ylvl7Ox25UuZj73Rn85DK7/OpenGov-voting?node-id=370-15772&t=hZlnKtCMb35nXJZl-4)
- [x] Multishard confirm screen [Figma link](https://www.figma.com/design/ylvl7Ox25UuZj73Rn85DK7/OpenGov-voting?node-id=733-196219&t=jOASZdpaGlLHp8ZJ-0 )
- [x] Create and sign unlock extrinsic
- [x] Save operation into the basket
- [x] Implement function for basket confirmation screen
- [x] Validation for network fee. Call validation on confirm screen and for basket

### LINKED ISSUE #2034: Governance. Unlock. Multisig opertaions list, details
- [x] Show unlock operation list
- [x] Show unlock operation details
- [x] Decode multisig unlock operation
- [x] Approve, reject unlock operation

sokolova-an:
- Multishard support for unlock operation
- Proxy support
- Add to basket and sign operations support
- Multisign support added

### DIFF
--- a/src/renderer/entities/transaction/lib/common/utils.ts
+++ b/src/renderer/entities/transaction/lib/common/utils.ts
@@ -97,7 +97,7 @@ export const isEditDelegationTransaction = (transaction?: Transaction | DecodedT
   return false;
 };
 
-export const isBatchAllWrapped = (type: TransactionType) => {
+export const isWrappedInBatchAll = (type: TransactionType) => {
   const batchAllOperations = new Set([TransactionType.UNSTAKE, TransactionType.BOND, TransactionType.UNLOCK]);
 
   return batchAllOperations.has(type);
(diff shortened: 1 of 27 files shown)


src/renderer/entities/transaction/lib/common/utils.ts
```diff
@@ -97,30 +97,40 @@ export const isEditDelegationTransaction = (transaction?: Transaction | DecodedT
   return false;
 };
 
+export const isBatchAllWrapped = (type: TransactionType) => {
```
Asmadek: I am not sure about name for this function. Is batch all wrapped looks like it just have batch all transaction type. But here we check that we can extract operations from batch all and correctly show it. But I don't know what the best name for this function

Asmadek (approved):

src/renderer/entities/transaction/lib/common/utils.ts
sokolova-an: I agree, I also struggled with the name for this one. Maybe something like `isWrappedInBatchAll` would make more sense here?

PR #2112, opened 2024-08-15
feat: update voting subscribe

sokolova-an:
Update voting subscribe using a common function


src/renderer/entities/governance/lib/governanceSubscribeService.ts
```diff
@@ -126,4 +133,8 @@ async function subscribeVotingFor(
 
     callback(result);
   });
+
+  return () => {
+    unsubscribe.then((x) => x());
```
Asmadek: maybe use `fn` instead of `x`?

Asmadek (approved):

johnthecat (approved):

PR #2116, opened 2024-08-16
fix: initial loading state for referendum list, refactored referendum subscription

### LINKED ISSUE #2002: Governance. Loading state. There is no loading state for referenda list
## Description

When user open referenda list with connection troubles - there is no loading state for referenda list.

## Steps to Reproduce

1. Switch the Internet off
2. Go to Governance page

## Expected Behavior
Loading state for referenda list:
image
https://www.figma.com/design/ylvl7Ox25UuZj73Rn85DK7/OpenGov-voting?node-id=101-11101&t=XU9wnC9WAmdozaBB-4
ongoing number should have loading state too

## Actual Behavior
There is empty page:
Screenshot 2024-07-30 at 10 30 11

## Environment
https://github.com/novasamatech/nova-spektr/commit/a326b27a5828e9884515bb7740a62fb3630fd886

johnthecat:


src/renderer/entities/governance/model/referendum.ts
```diff
@@ -1,74 +1,77 @@
...
+import { createSubscriber } from '../utils/createSubscriber';
+
+const mergeReferendums = (a: Referendum[], b: Referendum[]) => {
+  return merge(
+    a,
+    b,
+    (x) => x.referendumId,
+    (a, b) => parseInt(a.referendumId) - parseInt(b.referendumId),
+  );
 };
```
sokolova-an, line 18: Are you sure we need to merge them? We receive a full new updated list from the network anyway. In the case of merging, it can be a situation when the `referendum.id` is the same but `referendum.status` for example is different. In that case referendum list won't be updated when the `ongoing` referendum becomes `completed`

src/renderer/entities/governance/model/referendum.ts
```diff
@@ -1,74 +1,77 @@
...
+import { merge, nonNullable } from '@shared/lib/utils';
 import { governanceService } from '../lib/governanceService';
-
-type RequestListParams = {
-  chain: Chain;
-  api: ApiPromise;
+import { governanceSubscribeService } from '../lib/governanceSubscribeService';
+import { createSubscriber } from '../utils/createSubscriber';
+
+const mergeReferendums = (a: Referendum[], b: Referendum[]) => {
```
Asmadek, line 11: what do you think to move it so referendum or governance service?

Asmadek (approved):

src/renderer/entities/governance/lib/governanceSubscribeService.ts
```diff
@@ -138,3 +144,24 @@ function subscribeVotingFor(
...
   };
 }
+
+/**
+ * TODO move to chain helpers
+ */
+function subscribeBlockWithInterval(api: ApiPromise, ttl: number, callback: (header: Header) => unknown) {
+  const throttled = throttle(callback, ttl);
+  const unsubscribe = api.rpc.chain.subscribeNewHeads((header) => {
+    throttled(header);
```
sokolova-an, line 154: Trying to understand, why do we need to pass `header` into `throttled`?

src/renderer/entities/governance/lib/governanceSubscribeService.ts
```diff
@@ -138,3 +144,24 @@ function subscribeVotingFor(
...
+  });
+
+  return () => {
+    throttled.cancel();
+    unsubscribe.then((x) => x());
```
sokolova-an: let's use `fn` for these calls

src/renderer/entities/governance/model/referendum.ts
johnthecat: services holds mostly business logic, not transport-level mappings. I don't think that this is a good idea

src/renderer/features/governance/aggregates/list.ts
```diff
@@ -64,24 +65,13 @@ const $referendums = combine(
...
-    chain: networkSelectorModel.$governanceChain,
-  },
-  filter: ({ chain }, api) => !!api && !!chain,
-  fn: ({ chain }, api) => ({ api: api!, chain: chain! }),
-  target: referendumModel.events.updateReferendums,
+  source: networkSelectorModel.$governanceChain,
+  filter: (chain, api) => nonNullable(api) && nonNullable(chain),
+  fn: (chain, api) => ({ chain: chain!, api: api! }),
+  target: referendumModel.events.subscribeReferendums,
 });
```
sokolova-an, line 72: In which place now will we update/resubscribe the referendum list when we change the network?

src/renderer/entities/governance/model/referendum.ts
johnthecat: Merging lists is safer option regardless of current implementation of subscribe method. Merge method overrides record by key (`referendumId` in this case), so it shouldn't be a problem.

src/renderer/entities/governance/lib/governanceSubscribeService.ts
johnthecat: It can be useful, when this helper became shared

src/renderer/features/governance/aggregates/list.ts
johnthecat: in current implementation subscribe method reinit subscription with new parameters

PR #2139, opened 2024-08-19
fix: correct calculation of lock period in forms and confirm screens

### LINKED ISSUE #2128: Governance. Undelegate \ Lock period shouldn't be a constant.
## Description
Undelegate period shouldn't be a constant, that should be a value based on on-chain data.

## Steps to Reproduce

1. Go to Governance page
2. Click "Add delegation"
3. Select Delegate
4. Click "Add new Delegation"
5. Select tracks and click "Continue"

## Expected Behavior
Undelegate period is calculated, based on on-chain data, it's difference for different networks

## Actual Behavior
It takes constant
https://github.com/novasamatech/nova-spektr/blob/b3dfe1e8998b804a7576a620007ea616f19c2b4f/src/renderer/entities/governance/lib/lockService.ts#L8-L16

Screenshot 2024-08-19 at 11 33 44
Screenshot 2024-08-19 at 11 34 03

https://github.com/user-attachments/assets/8d4d7b41-884d-4f91-b8b6-06698ff46d73
https://github.com/user-attachments/assets/e5b004e0-caf8-4eef-a181-08032019fb3f
https://github.com/user-attachments/assets/e2700b72-f051-4afb-9032-5276ab20f4de

## Environment
https://github.com/novasamatech/nova-spektr/commit/b684d4d72c57f1886b1f49125131fafe6db13165

johnthecat:
Снимок экрана 2024-08-19 в 15 58 56

### DIFF
--- a/src/renderer/features/governance/aggregates/locksPeriod.ts
+++ b/src/renderer/features/governance/aggregates/locksPeriod.ts
@@ -7,7 +7,7 @@ import { networkModel } from '@entities/network';
 import { lockPeriodsModel } from '../model/lockPeriods';
 import { networkSelectorModel } from '../model/networkSelector';
 
-const lockPeriodsFetch = createGate<{ chain?: Chain }>();
+const flow = createGate<{ chain?: Chain }>();
 
 const $lockPeriodsInChain = combine(
   networkSelectorModel.$governanceNetwork,
@@ -35,7 +35,7 @@ sample({
 });
 
 sample({
-  clock: lockPeriodsFetch.open,
+  clock: flow.open,
   source: networkModel.$apis,
   filter: (apis, { chain }) => {
     return nonNullable(chain) && chain.chainId in apis;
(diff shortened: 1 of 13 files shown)


src/renderer/features/governance/aggregates/locksPeriod.ts
```diff
@@ -0,0 +1,61 @@
+import { combine, createEvent, sample } from 'effector';
+import { createGate } from 'effector-react';
+
+import { type Chain } from '@/shared/core';
+import { nonNullable } from '@shared/lib/utils';
+import { networkModel } from '@entities/network';
+import { lockPeriodsModel } from '../model/lockPeriods';
+import { networkSelectorModel } from '../model/networkSelector';
+
+const lockPeriodsFetch = createGate<{ chain?: Chain }>();
```
Asmadek: Are you sure about that name? I can't find a naming convention for it, but it seems like it should be a noun, not a verb

Asmadek (approved):

PR #2161, opened 2024-08-21
fix: show reusable lock

### LINKED ISSUE #2135: Governance. Governance lock. Calculation is wrong, when already have locked balance.
## Description

When user already has 10 locked tokens in governance and try to delegate 1 more token the app shouldn't show 11 locked tokens, it will be still 10.
Screenshot 2024-08-19 at 14 01 10
Screenshot 2024-08-19 at 14 01 05

## Steps to Reproduce

1. Vote for referenda with 10 WND
2. Try to add to another track delegation with 1 WND

## Expected Behavior
Shows 10 lock tokens with increased lock time

## Actual Behavior
Shows 10 -> 11

## Environment
https://github.com/novasamatech/nova-spektr/commit/a1e90ef296095ad76870c380166686644a2b55d7

sokolova-an:
- Changed the way we show governance lock differences in voting and delegation if we already have locks 
- Added helper for select tracks


johnthecat (approved):

src/renderer/widgets/DelegateModal/lib/helpers.ts
```diff
@@ -17,3 +17,13 @@ export const getTreasuryTrackDescription = (asset: Asset | null, description: st
 export const getTrackIds = (tracks: Track[], votedTracks: string[]): number[] => {
   return tracks.filter((t) => !votedTracks.includes(t.id)).map((track) => Number(track.id));
 };
+
+export const getTrackPallet = (
```
Asmadek: Better to call it `getGroupPallet`

Asmadek (approved):

PR #2296, opened 2024-09-05
refector: referendums zod migration

### LINKED ISSUE #1994: Governance. Add ongoing filter
## Description
Most users want to see ongoing referenda, but every time we will load the whole history list (around 1000 finished referendums) 
It would be great to add filter to show - ongoing \ completed \ all

That approach may significantly decrease network load. But the search algorithm should be improved, when user want to find referenda not by referenda id, it should show loader and request off-chain data for all referendum

https://www.figma.com/design/ylvl7Ox25UuZj73Rn85DK7?node-id=116-13598#889037788

johnthecat:
* Moved referendum requests to the new `pallet` transport system
* Switched referenda subscription from throttling to system events + extrinsic parsing
* Small cleanup

### DIFF
--- a/src/renderer/entities/governance/lib/governanceSubscribeService.ts
+++ b/src/renderer/entities/governance/lib/governanceSubscribeService.ts
@@ -7,6 +7,7 @@ import {
   type Referendum,
   type ReferendumId,
   type TrackId,
+  TransactionType,
   type Voting,
   type VotingMap,
 } from '@/shared/core';
@@ -52,7 +53,7 @@ function subscribeVotingFor(
   api: ApiPromise,
   tracksIds: TrackId[],
   addresses: Address[],
-  callback: (res?: VotingMap) => void,
+  callback: (voting: VotingMap) => void,
 ) {
   const tuples = addresses.flatMap((address) => tracksIds.map((trackId) => [address, trackId] as const));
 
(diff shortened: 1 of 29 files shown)


src/renderer/shared/polkadotjs-helpers/subscribeExtrinsics.ts
```diff
@@ -0,0 +1,21 @@
+import { type ApiPromise } from '@polkadot/api';
+import { type GenericExtrinsic } from '@polkadot/types';
+
+type Params = {
+  api: ApiPromise;
+  name: string[];
+};
+
+export const subscribeExtrinsics = ({ api, name }: Params, fn: (extrinsic: GenericExtrinsic) => unknown) => {
+  const methodsMap = Object.fromEntries(name.map((x) => [x, true]));
```
Asmadek: Why not to use `Set` for it? It will looks more readable

src/renderer/entities/governance/lib/governanceSubscribeService.ts
```diff
@@ -50,118 +53,131 @@ function subscribeVotingFor(
...
     }
 
     callback(result);
   });
+}
 
-  return () => {
-    unsubscribe.then((fn) => fn());
+function subscribeReferendums(api: ApiPromise, callback: (referendums: IteratorResult<Referendum[], void>) => unknown) {
+  let currectAbortController = new AbortController();
```
Asmadek, line 137: current?

Asmadek (approved):

src/renderer/entities/governance/lib/governanceSubscribeService.ts
johnthecat: On each event `AbortController` breaks current page request, but since we can't reset controller it got recreated every time
So, it's `current`

PR #2314, opened 2024-09-11
feat: revoke delegation

Asmadek:
* add revoke delegation transaction to delegations, basket, multisigs
* Add types for transaction types in basket


src/renderer/entities/transaction/lib/common/utils.ts
```diff
@@ -110,12 +110,25 @@ export const isDelegateTransaction = (transaction?: Transaction | DecodedTransac
   return false;
 };
 
+export const isUndelegateTransaction = (transaction?: Transaction | DecodedTransaction): boolean => {
```
johnthecat, line 103: Can you move this logic into separated function and also make is recursive? Like

```ts
export const hasTransaction = (transaction: Transaction, filter: (transaction: Transaction) => boolean): boolean => {
  if (transaction.type === TransactionType.BATCH_ALL) {
    return transaction.args.transactions?.some((tx: Transaction) => hasTransaction(tx, filter)) ?? false;
  }
  
  return filter(transaction);  
}
```

src/renderer/features/governance/components/Delegations/TotalDelegation.tsx
```diff
@@ -66,14 +66,15 @@ export const TotalDelegation = ({ onClick }: Props) => {
             <FootnoteText>{t('governance.delegations.label')}</FootnoteText>
           </div>
 
-          {isLoading && <Shimmering width={120} height={20} />}
-          {!isLoading &&
-            asset &&
-            (totalDelegation !== '0' ? (
-              <AssetBalance className="text-small-title" value={totalDelegation} asset={asset} />
+          {!isLoading && network?.asset ? (
```
johnthecat: You can check network against null and it will be enough
```suggestion
          {!isLoading && network ? (
```

src/renderer/widgets/RevokeDelegationModal/lib/constants.ts
```diff
@@ -0,0 +1,94 @@
+export const adminTracks = [
```
sokolova-an: don't we have the same constants for delegateModal? Maybe reuse it?

johnthecat (approved):

PR #2398, opened 2024-10-03
fix: multishard voting

### LINKED ISSUE #2221: Governance. Multishard. Can't vote as a multishard wallet
## Description
Can't Vote as a second account in multishard wallet

https://github.com/user-attachments/assets/65a685b0-de30-41d7-bf39-06b44b27e956

## Steps to Reproduce

1. Select multishard wallet with at least two accounts in network
2. Vote as account_A
3. Try to vote as account_B

## Expected Behavior
Can voting for as a multishard wallet with many accounts

## Actual Behavior
Can't vote as account_B

## Environment
https://github.com/novasamatech/nova-spektr/commit/cb012730236d518d6ac8c00c662b47397ad0135d

johnthecat:
Closes #2221

### DIFF
--- a/src/renderer/features/governance/aggregates/voting.ts
+++ b/src/renderer/features/governance/aggregates/voting.ts
@@ -1,9 +1,9 @@
 import { combine, createEvent, sample } from 'effector';
 
 import { type Address, type TrackId, type VotingMap } from '@shared/core';
-import { nonNullable } from '@shared/lib/utils';
+import { nonNullable, nullable } from '@shared/lib/utils';
 import { votingModel } from '@entities/governance';
-import { accountUtils, walletModel } from '@entities/wallet';
+import { accountUtils, walletModel, walletUtils } from '@entities/wallet';
 import { networkSelectorModel } from '../model/networkSelector';
 
 import { tracksAggregate } from './tracks';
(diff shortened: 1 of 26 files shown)


src/renderer/widgets/RemoveVotesModal/ui/RemoveVotesModal.tsx
```diff
@@ -112,17 +97,108 @@ export const RemoveVotesModal = ({ votes, voter, chain, asset, api, onClose }: P
...
+      }
+    }
+  }, [shouldPickAccount, isSelectAccountClosed, isSelectAccountClosed, setIsSelectAccountOpen, onCancel]);
+
+  return (
+    <AccountSelectModal
+      isOpen={isSelectAccountOpen}
+      asset={asset}
+      chain={chain}
+      title="Select account which vote you want to remove"
```
Asmadek: Don't forget to add i18n here

src/renderer/features/governance/aggregates/voting.ts
```diff
@@ -34,6 +34,28 @@ const $activeWalletVotes = combine(
...
+    if (walletUtils.isPolkadotVault(wallet)) {
+      const shards = wallet.accounts.filter((a) => accountUtils.isShardAccount(a) && a.chainId === chain.chainId);
+
+      if (shards.length) {
+        return shards;
+      } else {
+        return wallet.accounts.filter(
+          (a) => accountUtils.isBaseAccount(a) && accountUtils.isChainAndCryptoMatch(a, chain),
+        );
+      }
```
Asmadek: PV wallet can includes not only shard account. Here we may lost chain accounts. And if we don't have shards can use base account, which should be excluded.

src/renderer/features/governance/lib/createTransactionForm.ts
```diff
@@ -108,14 +108,23 @@ export const createTransactionForm = <FormShape extends NonNullable<unknown>>({
...
-        const isBase = accountUtils.isBaseAccount(a);
-        const isPolkadotVault = walletUtils.isPolkadotVault(w);
-
-        return (!isBase || !isPolkadotVault) && accountUtils.isChainAndCryptoMatch(a, chain);
-      });
+      if (nullable(wallet) || nullable(chain) || nullable(asset)) return [];
+
+      let walletAccounts: Account[] = [];
+
+      if (walletUtils.isPolkadotVault(wallet)) {
```
Asmadek, line 115: Same comment as previous

Asmadek (changes requested):

src/renderer/features/governance/aggregates/voting.ts
johnthecat: In case of voting mutishard accounts can vote only with shards, not base account
CC @stepanLav

Asmadek: Right, base account shouldn't be supported. But also we have chain accounts in PV wallets. And we skip it if we filter by shard accounts. 
image

Asmadek (approved):

PR #2455, opened 2024-10-12
feat: new TransactionDetails component, fixed Hash, fixed storybook

johnthecat:


src/renderer/shared/ui-entities/TransactionDetails/TransactionDetails.tsx
```diff
@@ -0,0 +1,116 @@
...
+});
+
+const AccountInfo = ({ account, chain }: { account: Account; chain: Chain }) => {
+  return (
+    <div className="flex w-full min-w-0 gap-2 text-text-secondary">
+      <Address showIcon variant="truncate" address={toAddress(account.accountId, { prefix: chain.addressPrefix })} />
+      <AccountExplorers accountId={account.accountId} chain={chain} />
+    </div>
+  );
+};
```
sokolova-an: We pretty often use it this way, is it worth moving it into a separate component?

sokolova-an (approved):

PR #2479, opened 2024-10-15
chore: import refactoring

johnthecat:
Removed all `@*` aliases in favor of `@/*`.
Also added stricter rules for boundaries - now user can import only from "package" root.
e.g. `@/domains/collective` instead of `@/domains/collective/models/tracks/types`).


binjospookie:
😍

sokolova-an (approved):

PR #2516, opened 2024-10-21
Fix: enforce threshold being set

### LINKED ISSUE #2511: Multisig. Can create without threshold.
## Description

User can create multisig wallet without threshold

Example:

https://github.com/user-attachments/assets/dda9e95d-f557-43aa-8fc7-6eebeac31894

## Steps to Reproduce

2. Add creator account
3. Add signatories
4. Press Continue button

## Expected Behavior
Can't proceed with empty threshold, show error

## Actual Behavior
Can proceed

## Environment
https://github.com/novasamatech/nova-spektr/commit/0a3e32d5edae91d623fe09e5f930055e615e403a

Tbaut:
closes #2511

### DIFF
--- a/src/renderer/shared/i18n/locales/en.json
+++ b/src/renderer/shared/i18n/locales/en.json
@@ -188,7 +188,7 @@
     "successMessage": "Wallet added",
     "thresholdDescription": "Set the threshold of signatories required to execute a transaction",
     "thresholdErrorTitle": "Invalid threshold",
-    "thresholdErrorDescription": "Threshold must be set and >= 2",
+    "thresholdErrorDescription": "Threshold must be set and superior or equal to { minThreshold }",
     "thresholdHint": "Threshold of signatories required to execute a transaction",
     "thresholdName": "Threshold",
     "thresholdOutOf": "{ threshold } out of { signatoriesLength }",
--- a/src/renderer/shared/i18n/locales/en.json
+++ b/src/renderer/shared/i18n/locales/en.json
@@ -187,6 +187,8 @@
     },
     "successMessage": "Wallet added",
     "thresholdDescription": "Set the threshold of signatories required to execute a transaction",
+    "thresholdErrorTitle": "Invalid threshold",
+    "thresholdErrorDescription": "Threshold must be set and >= 2",
     "thresholdHint": "Threshold of signatories required to execute a transaction",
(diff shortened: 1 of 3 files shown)


src/renderer/shared/i18n/locales/en.json
```diff
@@ -187,6 +187,8 @@
     },
     "successMessage": "Wallet added",
     "thresholdDescription": "Set the threshold of signatories required to execute a transaction",
+    "thresholdErrorTitle": "Invalid threshold",
+    "thresholdErrorDescription": "Threshold must be set and >= 2",
```
Asmadek: Is it confirmed translation? Looks like `>=` can be not clear for users.

src/renderer/widgets/CreateWallet/ui/MultisigWallet/SelectSignatoriesThreshold.tsx
```diff
@@ -46,7 +46,9 @@ export const SelectSignatoriesThreshold = () => {
   const hasOwnedSignatory = !!ownedSignatoriesWallets && ownedSignatoriesWallets?.length > 0;
   const hasEnoughSignatories = signatories.length >= 2;
   const hasEmptySignatory = signatories.map(({ address }) => address).includes('');
-  const canSubmit = hasOwnedSignatory && hasEnoughSignatories && !multisigAlreadyExists && !hasEmptySignatory;
+  const isThresholdValid = threshold.value >= 2 && threshold.value <= signatories.length;
```
Asmadek: Maybe add constant for 2 as `MIN_THRESHOLD` and use it in other places? In line 47 for example

Asmadek (approved):

src/renderer/shared/i18n/locales/en.json
Tbaut: It's not confirmed, I can change for "superior than 2"

src/renderer/widgets/CreateWallet/ui/MultisigWallet/SelectSignatoriesThreshold.tsx
Tbaut: Good point. The 2 is set in stone in the pallets, so we can use it in several places. I'll change this.

johnthecat (approved):

PR #2518, opened 2024-10-21
fix: prevent proceeding with same signatory

### LINKED ISSUE #2510: Multisig. Can process with the same signatories.
## Description

When user creates multisig it should not allow to add the same addresses multiple times

Example:

https://github.com/user-attachments/assets/516f08c3-1856-4c03-be0d-2cd2fec54fb3

## Steps to Reproduce

2. On "Create multisig" window add a few signatories with the same addresses

## Expected Behavior
Unique account list in signatory participations

## Actual Behavior
Can add the same address multiple times

## Environment
https://github.com/novasamatech/nova-spektr/commit/0a3e32d5edae91d623fe09e5f930055e615e403a

Tbaut:
closes #2510


src/renderer/widgets/CreateWallet/model/signatory-model.ts
```diff
@@ -31,6 +32,16 @@ sample({
   target: $signatories,
 });
 
+sample({
```
johnthecat: Can you use simple map here?
```ts
const $hasDuplicateSignatories = $signatories.map((signatories) => {
    const signatoriesArray = Array.from(signatories.values()).map(({ address }) => toAccountId(address));

    return new Set(signatoriesArray).size !== signatoriesArray.length;
  })
```

sokolova-an (approved):

src/renderer/widgets/CreateWallet/model/signatory-model.ts
Tbaut: I can't because `$signatories` is a `Map`. so that we keep the order (there's no guarantee in arrays).

Tbaut: or do you mean, I shouldn't bother with sample, and directly derive the store. This makes sense. Let me change

Tbaut: I used a combine now in [34765f6](https://github.com/novasamatech/nova-spektr/pull/2518/commits/34765f6bbc0a66dd7f9c2f79262b8167009698fe)

```ts
const $hasDuplicateSignatories = combine($signatories, (signatories) => {
  const signatoriesArray = Array.from(signatories.values()).map(({ address }) => toAccountId(address));

  return new Set(signatoriesArray).size !== signatoriesArray.length;
});
```

johnthecat (approved):

pgolovkin (approved):

PR #2520, opened 2024-10-22
fix: Remove description from confirmation modals

### LINKED ISSUE #1807: Remove multisig Description and Reject reason

Tbaut:
closes #1807 

I thought these descriptions could be used for some other use-case than multisig, and since we removed the input, they would only be displayed when used, but it sounds like we want to remove them all no matter what, so here it is.


sokolova-an (commented):
Should we, in this case, delete all description props from the confirm-modals and form-modals since we no longer use them?

Tbaut:
Valid question, I am not sure this will never be used in the future, so I went with the safe way to only remove their display, but keep the stores.

tuul-wq (approved):

pgolovkin:
@Tbaut I think it's better to remove code that is not used. And not going to be used in future.

Tbaut:
ok I removed the descriptions form the stores @sokolova-an

johnthecat (approved):

PR #2528, opened 2024-10-23
feat: extracted app shell feature from main layout

johnthecat:


src/renderer/features/wallets-select/index.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { createFeature } from '@/shared/effector';
+import { navigationHeader } from '@/features/app-shell';
+import { SelectWalletPairing, WalletSelect } from '@/features/wallets';
+
+export const walletsSelectFeature = createFeature({
+  name: 'Wallets select',
+  enable: createStore(true),
+});
+
+walletsSelectFeature.inject(navigationHeader, () => {
```
Asmadek: Just a notice. Inject has no difference between inject pipeline and slot. During development we should check what we inject, or validation will says us that we have incorrect second argument. I am not sure that we should do something with it, but it can confuse a little. Use `Slot`suffix (`navigationHeaderSlot`) maybe can fix it

Asmadek (approved):

src/renderer/features/app-shell/components/NavItem.tsx
```diff
@@ -0,0 +1,40 @@
...
+export type Props = {
+  order?: number;
+  title: string;
+  link: string;
+  icon: IconNames;
+  badge?: ReactNode;
+};
+
+export const NavItem = ({ title, link, icon, badge }: Props) => {
+  const { t } = useI18n();
```
tuul-wq, line 17: Would be great if we can remove `useI18n` from this component, so it renders whatever `title` it gets

src/renderer/features/app-shell/components/Navigation.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const lowerItems = usePipeline(navigationBottomLinks, []);
+
+  return (
+    <nav className="h-full overflow-y-auto">
+      <ul className="flex h-full flex-col gap-2">
+        {upperItems.map(({ icon, title, link, badge }) => (
+          <NavItem key={link} icon={icon} title={title} link={link} badge={badge} />
+        ))}
+
+        <div className="mt-auto flex flex-col gap-2">
```
tuul-wq, line 25: Not sure we need `div` inside `ul`
Could place "settings group" below main navigation items

src/renderer/features/assets-navigation/index.ts
```diff
@@ -0,0 +1,16 @@
+import { createStore } from 'effector';
+
+import { $features } from '@/shared/config/features';
+import { createFeature } from '@/shared/effector';
+import { Paths } from '@/shared/routes';
+import { navigationTopLinks } from '@/features/app-shell';
```
tuul-wq: Path should be relative isn't it? `../app-shell`? Or exporting one feature into another is ok?

src/renderer/features/assets-navigation/index.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const assetsNavigationFeature = createFeature({
+  name: 'Assets navigation',
+  input: createStore({}),
+  enable: $features.map(({ assets }) => assets),
+});
+
+assetsNavigationFeature.inject(navigationTopLinks, (items) => {
+  return items.concat({ order: 0, icon: 'asset', title: 'navigation.balancesLabel', link: Paths.ASSETS });
+});
```
tuul-wq: To increase readability (same for the following features).
By the way, now I see why do we have to `useI18n` in `NavItem` 😢
So until we extract i18n into effector these strings would exist
```suggestion
assetsNavigationFeature.inject(navigationTopLinks, (items) => {
  return items.concat({
    order: 0,
    icon: 'asset',
    title: 'navigation.balancesLabel',
    link: Paths.ASSETS,
  });
});
```

src/renderer/features/contacts-navigation/index.ts
```diff
@@ -0,0 +1,18 @@
...
+import { createFeature } from '@/shared/effector';
+import { Paths } from '@/shared/routes';
+import { navigationTopLinks } from '../app-shell';
+
+export const contactsNavigationFeature = createFeature({
+  name: 'Contacts navigation',
+  enable: $features.map(({ contacts }) => contacts),
+});
+
+contactsNavigationFeature.inject(navigationTopLinks, (items) => {
```
tuul-wq: 👍

src/renderer/features/wallets-select/index.tsx
tuul-wq: Agree, just stumbled upon this when came to look `.tsx` file

sokolova-an (approved):

src/renderer/features/wallets-select/index.tsx
johnthecat: The difference between identifiers are backed into types, which inferred into handler (2nd arg). Naming convention maybe a good decision and we should also fix it into our lint rules.

src/renderer/features/app-shell/components/NavItem.tsx
johnthecat: it will be possible when we move i18n outside react

src/renderer/features/assets-navigation/index.ts
johnthecat: each feature is threatened like separated "package", so relative imports are forbidden

PR #2545, opened 2024-10-28
Fix: UI nits

### LINKED ISSUE #2531: Multisig. UI issues. Create ms flow
### 1. Create ms popup
Image

[Layouts reference](https://www.figma.com/design/HgFuj8QZd0xqEbkG10DIfs/Flexible-multisig-%7C-Milestone-4?node-id=369-58126&t=bKjB2kBmhRjZrNaX-1)

1. align to one line and update text of description 
2. Close button state of is hover on initial screen enter
3. update placeholder text: "Choose wallet name" -> "Enter name"
4. update text
5. add Back button
6. skeleton place is shifted close to Continue
7. "Additional costs" -> "Additional cost"

### 2. Create ms on the network popup
Image

[Layouts reference](https://www.figma.com/design/HgFuj8QZd0xqEbkG10DIfs/Flexible-multisig-%7C-Milestone-4?node-id=369-58363&t=bKjB2kBmhRjZrNaX-1)

1. NO NEED TO FIX (update popup title)
2. align and update description text
3. Label: Name -> Signatory name
4. Label: Your account -> Address
5. update placeholder text
6. plus button is wrong color
7. update placeholder text: Select threshold -> Select number
8. maybe disable threshold dropdown when <2 signatories inserted ???
9. update text
10. shifted skeleton place
11. Continue button is active anytime

### 3. Create ms editing fields
Image

[Layouts reference](https://www.figma.com/design/HgFuj8QZd0xqEbkG10DIfs/Flexible-multisig-%7C-Milestone-4?node-id=1105-46243&t=bKjB2kBmhRjZrNaX-1)

1. when input address: inserted name is replaced with placeholder, name field is not editable
2. too big remove icon
3. shifted skeleton place

### 4. Confirm ms popup

Image

[Layouts reference](https://www.figma.com/design/HgFuj8QZd0xqEbkG10DIfs/Flexible-multisig-%7C-Milestone-4?node-id=976-125596&t=bKjB2kBmhRjZrNaX-1)

1. close button is in hover state
2. no address under the name

### 5. Wallet details

- Singlechain multisig looks like this, which is not according to layout: network should be in the title with threshold info, there are two tabs: "Account and signatories" and "Delegated authorities"

Image

[Layouts reference](https://www.figma.com/design/HgFuj8QZd0xqEbkG10DIfs/Flexible-multisig-%7C-Milestone-4?node-id=383-77731&t=GP7uXMP5hmNOy6f7-1)

- Multichain multisig looks like this, also not according the layout: all multisig wallets are separate in each network, so in network 	 only one network is possible

Image
Image

[Layouts reference](https://www.figma.com/design/HgFuj8QZd0xqEbkG10DIfs/Flexible-multisig-%7C-Milestone-4?node-id=907-52209&t=V2B6hpWHF4LJPQL4-1)

Image
Image
Image

1. function button is in hover state
2. Remove treshold info, "Your accounts 1" -> "Account"
3. Contacts -> Signatories
4. Wallet name should be near Signatory address

### 6. Operations

Image

[Layouts reference](https://www.figma.com/design/HgFuj8QZd0xqEbkG10DIfs/Flexible-multisig-%7C-Milestone-4?node-id=445-31031&t=nNUCyVpyhtqd8KiY-1)

1. wrong icon
2. missing dialog icon
3. missing Sender
4. Call Hash -> Call hash
5. Call Data -> Call data
6. missing i icon

Tbaut:
closes #2531
image

image

image
image


src/renderer/widgets/CreateWallet/ui/MultisigWallet/components/MultisigCreationFees.tsx
```diff
@@ -62,7 +62,7 @@ export const MultisigCreationFees = memo(
     }, []);
 
     if (isLoading) {
-      return <FeeLoader fiatFlag={Boolean(fiatFlag)} />;
+      return <FeeLoader className="mr-4" fiatFlag={Boolean(fiatFlag)} />;
```
johnthecat: Maybe just div with `mr-4`? Should be easier, than passing new prop to logical component

Tbaut: I'd say it's pretty standard to let any component get styled according to the use-case isn't it? I could obviously wrap in a div, but it seemed to me that it's not what we were doing for components in general. What do you think?

src/renderer/widgets/CreateWallet/ui/MultisigWallet/components/SelectedSignatoriesModal.tsx
```diff
@@ -26,17 +25,14 @@ export const SelectedSignatoriesModal = ({ isOpen, signatories, onClose, address
...
-            >
-              <AccountAddress
-                size={20}
-                type="short"
-                address={toAddress(address, { prefix: addressPrefix })}
-                name={name}
-                canCopy={true}
-              />
+            <li key={address} className="mb-2 ml-5 mr-2 flex items-center">
+              <Identicon className="inline-block" address={address} size={26} background={false} />
```
Asmadek: What is the reason to use AccountAddress component? Maybe new [Address component](https://github.com/novasamatech/nova-spektr/blob/af89d11aebeaa978e708602808dcdc1b91fe5c34/src/renderer/shared/ui-entities/Address/Address.tsx#L24) from ui-entities can help [storybook](https://spektr-storybook.s3-website.nl-ams.scw.cloud/?path=/story/design-system-entities-account--with-title)

Asmadek (approved):

src/renderer/widgets/CreateWallet/ui/MultisigWallet/components/MultisigCreationFees.tsx
tuul-wq: I assume the idea is to seal ui components because they're designed absolutely perfect (at least we should try to do so), then use additional wrappers to add space if it's required and not bother about ui component implementation and styles. Seems to me that's an Open-closed principle in UI.
@johnthecat what do you think?

tuul-wq (approved):

src/renderer/widgets/CreateWallet/ui/MultisigWallet/components/MultisigCreationFees.tsx
Tbaut: got it, I'm unopinionated about this. I've added a wrapper `div`.

src/renderer/widgets/CreateWallet/ui/MultisigWallet/components/SelectedSignatoriesModal.tsx
Tbaut: Somehow, I thought the `Address` component didn't support names and icons. I changed this, thanks

PR #2573, opened 2024-11-05
fix: refresh wc session

Asmadek:
* add error on incorrect refresh
* remove old session after refresh wc session
* don't remove accounts on refresh


johnthecat (approved):

sokolova-an (approved):

src/renderer/entities/walletConnect/model/wallet-connect-model.ts
```diff
@@ -167,24 +177,20 @@ type InitConnectResult = {
 };
 const initConnectFx = createEffect(
```
tuul-wq: Can be simplified to this:
```javascript
type InitConnectResult = {
  uri?: string;
  approval: () => Promise<SessionTypes.Struct>;
};
const initConnectFx = createEffect(({ client, chains, pairing }: InitConnectParams): Promise<InitConnectResult> => {
  const optionalNamespaces = {
    polkadot: {
      chains,
      methods: [DEFAULT_POLKADOT_METHODS.POLKADOT_SIGN_TRANSACTION],
      events: [DEFAULT_POLKADOT_EVENTS.CHAIN_CHANGED, DEFAULT_POLKADOT_EVENTS.ACCOUNTS_CHANGED],
    },
  };

  return client.connect({ pairingTopic: pairing?.topic, optionalNamespaces });
});
```

And I see that `client.connect` doesn't return `undefined` does it?

tuul-wq (approved):

PR #2601, opened 2024-11-08
fix: integrated with new asset type from chains.json

johnthecat:


Asmadek (approved):

sokolova-an (approved):

src/renderer/shared/core/types/asset.ts
```diff
@@ -19,6 +19,7 @@ export const enum StakingType {
 }
 
 export const enum AssetType {
```
tuul-wq, line 21: Now we can make `type` required
```javascript
export type Asset = {
  // ... other props
  type: AssetType;
};
```

tuul-wq (changes requested):

tuul-wq (approved):

PR #2619, opened 2024-11-11
Feat: Select ui-kit

tuul-wq:
- New `Select` component

image


src/renderer/shared/ui-kit/Select/Select.tsx
```diff
@@ -0,0 +1,149 @@
...
+}>;
+
+type RootProps = PropsWithChildren<ControlledSelectProps & ContextProps>;
+
+const Root = ({
+  theme = 'light',
+  invalid,
+  disabled,
+  testId = 'Select',
+  open,
```
sokolova-an, line 37: let's call boolean values for `open` as `isOpen` it's easier to read in components

sokolova-an (approved):

src/renderer/shared/ui-kit/Select/Select.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+  parameters: {
+    layout: 'centered',
+  },
+  render: (params) => {
+    const [value, onChange] = useState('');
+
+    return (
+      <Box width="200px">
+        <Select {...params} value={value} onChange={onChange}>
+          <Select.Button placeholder="Select a fruit" />
```
johnthecat: I don't see any use case for separated Button component

src/renderer/shared/ui-kit/Select/Select.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+  },
+};
+
+export default meta;
+
+type Story = StoryObj<typeof Select>;
+
+export const Default: Story = {};
+
+export const Controlled: Story = {
```
johnthecat: This stories make no sense

src/renderer/shared/ui-kit/Select/Select.tsx
```diff
@@ -0,0 +1,149 @@
...
+  onChange: (value: string) => void;
+} & XOR<{
+  open: boolean;
+  onToggle: (value: boolean) => void;
+}>;
+
+type RootProps = PropsWithChildren<ControlledSelectProps & ContextProps>;
+
+const Root = ({
+  theme = 'light',
```
johnthecat, line 33: Why?

tuul-wq: We have a dark colored Select when you scan QR from PV 
image

src/renderer/shared/ui-kit/Select/Select.tsx
tuul-wq: Select props were based on Dropdown props that has boolean `open` 
Also discussed this moment with @johnthecat and decided to stay with `open`

PR #2625, opened 2024-11-12
Feat: Field UI component

tuul-wq:
#### Motivation

Every input, dropdown, select etc. inside a form is wrapped into a `<label>` according to our design system.
Instead of component composition we hard wrap inputs, dropdowns, selects which is less maintainable and expandable.
In order to solve this issue I introduce `<Field />` that can wrap any child with a `<label>`
(I'd have betted chosen `Label` name but it's occupied)

- Field UI component that wraps child in <label>

image


src/renderer/shared/ui-kit/Field/Field.tsx
```diff
@@ -0,0 +1,14 @@
+import { type PropsWithChildren } from 'react';
+
+type Props = {
+  text: string;
+};
+
+export const Field = ({ text, children }: PropsWithChildren<Props>) => {
+  return (
+    <label className="flex w-full flex-col gap-y-2">
+      <span className="text-footnote font-medium text-text-tertiary">{text}</span>
```
Asmadek, line 10: What about case when we have addition text on the right side? 
image

tuul-wq: Hm, good question
Is it the only component with such `suffix` element?
Can introduce something like `suffix` prop that accepts `ReactNode`
@sokolova-an @johnthecat what you guys think?

sokolova-an: Sounds good to me

johnthecat: Maybe you can just use flexbox for this particular case. There is no need to add additional prop

johnthecat (approved):

PR #2635, opened 2024-11-12
feat: flexible multisig discovery

johnthecat:


src/renderer/entities/notification/model/notification-model.ts
```diff
@@ -44,5 +51,6 @@ export const notificationModel = {
   events: {
     notificationsStarted,
     notificationsAdded,
+    notificationAdded,
```
sokolova-an: where do we trigger it?

src/renderer/processes/multisigs/lib/mulitisigs-utils.ts
```diff
@@ -1,43 +1,85 @@
...
 };
 
-function isMultisigSupported(chain: Chain): boolean {
-  return Boolean(chain.options?.includes(ChainOptions.MULTISIG));
+function isMultisigSupported(chain: Chain) {
+  return chain.options?.includes(ChainOptions.MULTISIG) ?? false;
+}
+
+function isFlexibleMultisigSupported(chain: Chain) {
+  return isMultisigSupported(chain) && chain.options?.includes(ChainOptions.PURE_PROXY);
```
sokolova-an: As I recall it can be any proxy not only pure for multisig to be flexible (but we create flexible only with pure)

src/renderer/processes/multisigs/model/multisigs-model.ts
```diff
@@ -1,151 +1,258 @@
...
+        indexedMultisigs
+          // filter out multisigs that already exists
+          .filter((multisigResult) => !(multisigResult.accountId in accountsMap))
+          .map(({ threshold, accountId, signatories }): GetMultisigResponse => {
+            const proxiesList = proxies[accountId];
+            const proxy =
+              nonNullable(proxiesList) && proxiesList.length > 0
+                ? proxiesList.find((p) => {
+                    return p.chainId === chain.chainId && p.proxyType === 'Any';
+                  })
```
sokolova-an: again need to check about 'proxyType' can it be some other proxy here?

src/renderer/entities/notification/model/notification-model.ts
johnthecat: nowhere, just for fun

src/renderer/processes/multisigs/lib/mulitisigs-utils.ts
johnthecat: oh, okay, i'll fix it

sokolova-an (approved):

PR #2664, opened 2024-11-15
fix: correctly save new contacts after add new multisig

Asmadek:
* fix: save multiple contacts


src/renderer/entities/contact/model/contact-model.ts
```diff
@@ -37,6 +37,9 @@ $contacts
   .on(createContactFx.doneData, (state, contact) => {
     return contact ? state.concat(contact) : state;
   })
+  .on(createContactsFx.doneData, (state, contacts) => {
+    return contacts ? state.concat(contacts) : state;
+  })
```
sokolova-an: do we need to have it twice?

johnthecat: This

johnthecat (approved):

PR #2668, opened 2024-11-15
Update test implementation

stepanLav:
That PR:
- Change logic to find elements in the app by testID
- Update test implementation in order to decrease repetition


tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "strict": true,
     "target": "es2022",
     "module": "es2022",
-    "lib": ["es2022"],
+    "lib": ["es2022", "dom"],
```
stepanLav, line 13: Added in order to find indexedDB object in tests/system/utils/interactWithDatabase.ts, can't pass eslint rules

src/renderer/pages/Onboarding/Welcome/WelcomeCard.tsx
```diff
@@ -9,13 +9,15 @@ type Props = {
   iconName: IconNames;
   onClick?: () => void;
   disabled?: boolean;
+  buttonTestId?: string;
```
johnthecat: Why not simple `testId`?

src/renderer/shared/ui-entities/AccountExplorer/AccountExplorers.tsx
```diff
@@ -20,7 +21,12 @@ export const AccountExplorers = memo(({ accountId, chain, children }: Props) =>
   return (
     <Popover align="end" dialog testId="AccountExplorers">
       <Popover.Trigger>
-        <IconButton name="details" className="text-icon-default" onClick={(e) => e.stopPropagation()} />
+        <IconButton
+          name="details"
+          className="text-icon-default"
+          testId={TEST_IDS.MAIN.INFO_BUTTON}
```
johnthecat: I think ui-entities and ui-kit should be free from app-centric constants like this. It's better to get testId from component props and have reasonable default, usually it's component name

johnthecat (changes requested):

src/renderer/pages/Onboarding/Welcome/WelcomeCard.tsx
stepanLav: WelcomeCard has many elements, I thought that it may need to have a few ids, but now only one is needed, I will change it, ty!

src/renderer/shared/constants/testIds.ts
```diff
@@ -0,0 +1,26 @@
...
+    NOVA_WALLET_BUTTON: 'onboarding-nova-wallet-button',
+    WALLET_CONNECT_BUTTON: 'onboarding-wallet-connect-button',
+    LEDGER_BUTTON: 'onboarding-ledger-button',
+    WALLET_NAME_INPUT: 'onboarding-wallet-name-input',
+    WALLET_ADDRESS_INPUT: 'onboarding-wallet-address-input',
+  },
+  ASSETS: {
+    SETTINGS_WIDGET: 'assets-settings-widget',
+  },
+  MAIN: {
```
tuul-wq: As I understand here we locate common items? If so let's name is as `common`

src/renderer/shared/constants/testIds.ts
```diff
@@ -0,0 +1,26 @@
...
+    WALLET_BUTTON: 'main-wallet-button',
+    BACK_BUTTON: 'main-back-button',
+    CONTINUE_BUTTON: 'main-continue-button',
+    INFO_BUTTON: 'main-info-button',
+  },
+  TRANSFER: {
```
tuul-wq: Maybe just `operations`?
`Network fee` is a really common thing for any kind of operation
Thus I suggest to not go deeper that 1 level of nesting, or we'll end up with file similar to `en.json` which is hard to parse visually.

src/renderer/shared/ui-entities/AccountExplorer/AccountExplorers.tsx
stepanLav: Is that what you meant?
https://github.com/novasamatech/nova-spektr/pull/2668/commits/d6a12e55d588fc3eb8b699293ed50b70a3df73e8

src/renderer/shared/constants/testIds.ts
stepanLav: https://github.com/novasamatech/nova-spektr/pull/2668/commits/97b462c05a7e92381ef0de87eba4ef253a4342a8

src/renderer/shared/constants/testIds.ts
stepanLav: https://github.com/novasamatech/nova-spektr/pull/2668/commits/d6a12e55d588fc3eb8b699293ed50b70a3df73e8

tuul-wq (approved):

johnthecat (approved):

PR #2692, opened 2024-11-20
fix: update signatory name if edited

### LINKED ISSUE #2663: Multisig. Creation flow. Can't change signatory name if that address has in Address book
## Description

Name field is read only in multisig creation flow, if that address has in Address Book. User should be able to change it, while create a multisig

Example:

https://github.com/user-attachments/assets/0f8cd430-5eaf-4605-955c-678e90958189

## Steps to Reproduce

1. Add any address in Address book
2. Go to Multisig creation flow
3. Add signatory from the address book
4. Try to update the name

## Expected Behavior
User should be able to change signatory name, even if it comes from the Address Book

## Actual Behavior
Can't change name while create multisig

## Environment
https://github.com/novasamatech/nova-spektr/commit/53c448e24a584dfd10867e6805e68d557fa5bda8

Asmadek:
* add update contacts function
* update contact name if it was edited during create multisig


src/renderer/entities/contact/model/contact-model.ts
```diff
@@ -24,6 +24,14 @@ const updateContactFx = createEffect(async ({ id, ...rest }: Contact): Promise<C
   return { id, ...rest };
 });
 
+const updateContactsFx = createEffect(async (contacts: Contact[]): Promise<Contact[]> => {
+  if (contacts.length === 9) return [];
```
sokolova-an: why specifically 9? Maybe add a comment or make a const with this number so it will be clear

src/renderer/entities/contact/model/contact-model.ts
```diff
@@ -44,6 +52,13 @@ $contacts
     const position = state.findIndex((s) => s.id === contact.id);
 
     return splice(state, contact, position);
+  })
+  .on(updateContactsFx.doneData, (state, contacts) => {
```
tuul-wq, line 56: Oh, it's been so long since I've seen `.on` in effector code 😄

src/renderer/entities/contact/model/contact-model.ts
```diff
@@ -44,6 +52,13 @@ $contacts
...
+    return contacts.reduce((acc, contact) => {
```
johnthecat: Use merge helper from shared for such cases

src/renderer/widgets/CreateWallet/model/flow-model.ts
```diff
@@ -411,6 +411,29 @@ sample({
...
 });
 
+sample({
+  clock: signModel.output.formSubmitted,
+  source: {
+    signatories: signatoryModel.$signatories,
+    contacts: contactModel.$contacts,
+  },
+  fn: ({ signatories, contacts }) => {
+    return signatories.slice(1).reduce<Contact[]>((acc, { address, name }) => {
```
tuul-wq: I suggest to rewrite using `for_or`
Also there is a validation in contact create form that forbids duplicated `accountId`, I assume we can't have duplicated contacts with the same `accountId`
Thus we can create a lookup set `contactsSet: Set<AccountId>` and use it instead of `.find`

src/renderer/widgets/CreateWallet/model/flow-model.ts
```diff
@@ -411,6 +411,29 @@ sample({
...
+      const contact = contacts.find((c) => c.accountId === toAccountId(address) && c.name !== name);
+
+      if (contact) {
+        acc.push({
```
tuul-wq: Maybe recently added `immer.js` can play well in here 🤔

src/renderer/entities/contact/model/contact-model.ts
Asmadek: It's typo🙈

tuul-wq (approved):

PR #2697, opened 2024-11-20
fix: multisig wallet select

### LINKED ISSUE #2612: Multisig. Create multisig - won't change the wallet when create multisig.
## Description

When user create multisig by himself - the app should switch on new wallet

## Steps to Reproduce

2. Create a new one

## Expected Behavior
After creation - current wallet should be selected as multisig

## Actual Behavior
Nothing happened

## Environment
https://github.com/novasamatech/nova-spektr/commit/1b29836ffc7ca408a44df236bae6f0d8b98bf087

johnthecat:
Closes #2612

### DIFF
--- a/src/renderer/processes/multisigs/model/multisigs-model.ts
+++ b/src/renderer/processes/multisigs/model/multisigs-model.ts
@@ -51,7 +51,7 @@ type GetMultisigsResult = {
 const getMultisigsFx = createEffect(({ chains, wallets }: GetMultisigsParams) => {
   for (const chain of chains) {
     const filteredWallets = walletUtils.getWalletsFilteredAccounts(wallets, {
-      walletFn: (w) => !walletUtils.isMultisig(w) && !walletUtils.isWatchOnly(w) && !walletUtils.isProxied(w),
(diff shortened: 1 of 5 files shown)


src/renderer/processes/multisigs/model/multisigs-model.ts
```diff
@@ -51,7 +51,7 @@ type GetMultisigsResult = {
 const getMultisigsFx = createEffect(({ chains, wallets }: GetMultisigsParams) => {
   for (const chain of chains) {
     const filteredWallets = walletUtils.getWalletsFilteredAccounts(wallets, {
-      walletFn: (w) => !walletUtils.isMultisig(w) && !walletUtils.isWatchOnly(w) && !walletUtils.isProxied(w),
+      walletFn: (w) => !walletUtils.isMultisig(w) && !walletUtils.isProxied(w),
     });
```
sokolova-an: If we don’t exclude watchOnly, is it possible that we could receive multisig wallets for those as well?

johnthecat: That was for debug purposes(((

sokolova-an (approved):

PR #2718, opened 2024-11-22
Feat: add new operation view for flexible multisig creation

sokolova-an:
It would be good to move this part to feature and use models there as soon as Operation is refactored


src/renderer/pages/Operations/Operations.test.tsx
tuul-wq, line 1: 👍

tuul-wq, line 1: 👍

src/renderer/pages/Operations/components/FlexibleMultisigShell.tsx
```diff
@@ -0,0 +1,180 @@
...
+import RejectTxModal from './modals/RejectTx';
+
+type Props = {
+  tx: FlexibleMultisigTransactionDS;
+  account: FlexibleMultisigAccount;
+};
+
+export const FlexibleMultisigShell = ({ tx, account }: Props) => {
+  const { t } = useI18n();
+  const { connection, extendedChain } = useNetworkData(tx.chainId);
```
tuul-wq: Any chance to avoid `useNetworkData`?

src/renderer/pages/Operations/components/FlexibleMultisigShell.tsx
```diff
@@ -0,0 +1,180 @@
...
+              </Button>
+            </RejectTxModal>
+          )}
+          {connection && (
+            <ApproveTxModal tx={tx} account={account} connection={extendedChain}>
+              <Button className="ml-auto">{t('operation.approveButton')}</Button>
+            </ApproveTxModal>
+          )}
+        </div>
+        {<Signatories signatories={tx.signatories} connection={extendedChain} events={events} />}
```
tuul-wq:
```suggestion
        <Signatories signatories={tx.signatories} connection={extendedChain} events={events} />
```

src/renderer/pages/Operations/components/FlexibleMultisigShell.tsx
```diff
@@ -0,0 +1,180 @@
...
+            <ul className="flex flex-col gap-y-2">
+              {walletSignatories.map((signatory) => (
+                <SignatoryCard
+                  key={signatory.accountId}
+                  accountId={signatory.accountId}
+                  addressPrefix={connection.addressPrefix}
+                  status={signatory.status}
+                  explorers={connection.explorers}
+                >
+                  <div className="flex w-[100px] grow items-center gap-x-2 text-text-secondary">
```
tuul-wq: If this is possible 😉
```suggestion
                  <div className="flex w-25 grow items-center gap-x-2 text-text-secondary">
```

src/renderer/pages/Operations/components/FlexibleMultisigShell.tsx
```diff
@@ -0,0 +1,180 @@
...
+                      title={signatory.wallet.name}
+                      address={signatory.address}
+                      showIcon={false}
+                      variant="truncate"
+                    />
+                  </div>
+                </SignatoryCard>
+              ))}
+
+              {contactSignatories.length > 0 && (
```
tuul-wq: If `length` is 0 then it won't render a thing, so it's safe to just do `contactSignatories.map`

tuul-wq (commented):
Some minor stuff

src/renderer/pages/Operations/components/FlexibleMultisigShell.tsx
johnthecat: I'm trying to fix it right now, it leads to some refactoring on Operations page, but we wanted it anyway

sokolova-an: Will be fixed in a separate PR as a part of Operation refactoring

tuul-wq (approved):

PR #2727, opened 2024-11-26
fix: dev build for web

stepanLav:


.github/workflows/docker_images.yaml
```diff
@@ -16,6 +16,7 @@ concurrency:
 env:
   REGISTRY: ghcr.io
   IMAGE_NAME: ${{ github.repository }}
+  BUILD_TYPE: prod
```
johnthecat: it's better to use "development"/"production" to match our toolchain wording

johnthecat (approved):

Issue #2589, opened 2024-11-07
Update drop-down component through the whole app


stepanLav:
## Description

Need to update drop-down component on all screens, based on that bug:
https://github.com/novasamatech/nova-spektr/issues/2588

## Environment
https://github.com/novasamatech/nova-spektr/commit/6e5b60e8edb3ab459034c64cf277a3dbc8487fd3

pgolovkin:
Should be done in `flexible-multisig` branch.

stepanLav:
Stale issue

PR #2736, opened 2024-11-26
Feat: New Select & nullish signatory

tuul-wq:
- New Select for operations
- Signatory is now honest `Account | null`
- Improved readability
- More usage of `nonNullable`

part of #2589


src/renderer/shared/ui-entities/Address/Address.tsx
```diff
@@ -14,32 +14,43 @@ type IconProps = XOR<{
...
-  ({ title, variant = 'full', canCopy = true, showIcon, iconSize = 16, address, testId = 'Address' }: Props) => {
+  ({
+    title,
+    variant = 'full',
+    canCopy = true,
+    showIcon,
+    iconSize = 16,
+    address,
+    testId = 'Address',
+    hideAddress,
```
tuul-wq, line 31: Renamed `replaceAddressWithTitle` to `hideAddress`
If `hideAddress = true` and title is provided as well then hide address
Otherwise `address` becomes `text-text-primary`
Looks like we've planned to do so but forgot to implement because `replaceAddressWithTitle` was never used

src/renderer/shared/ui-kit/Select/Select.tsx
```diff
@@ -85,8 +85,8 @@ const Button = ({ placeholder }: TriggerProps) => {
         },
       )}
     >
-      <div className="overflow-hidden">
-        <RadixSelect.Value placeholder={placeholder} />
+      <div className="flex-1 overflow-hidden text-left">
```
tuul-wq: For some reason `Select` centers content if it occupies all width, but in our components we never intent to do so

Asmadek (approved):

src/renderer/shared/ui-entities/Address/Address.tsx
```diff
@@ -14,32 +14,43 @@ type IconProps = XOR<{
...
+          hidden: title && hideAddress,
         })}
       >
         <Hash value={address} variant={variant} />
       </span>
     );
 
     return (
-      <span className="flex w-full min-w-0 items-center gap-2 overflow-hidden" data-testid={testId}>
+      <div className="flex w-full min-w-0 items-center gap-x-2 overflow-hidden" data-testid={testId}>
```
johnthecat: I've used `span` because Address can be used inside text elements (`p`, etc.).

src/renderer/shared/ui-kit/Select/Select.tsx
johnthecat: It's better to use `text-start`

PR #2737, opened 2024-11-26
Feat: Select camera

tuul-wq:
- New Select for Dynamic derivations reader
- Operations QR scan camera select

part of #2589


src/renderer/entities/transaction/ui/QrCode/QrReader/QrReaderWrapper.tsx
```diff
@@ -54,18 +54,15 @@ export const QrReaderWrapper = ({ className, onResult, countdown, validationErro
...
       value: camera.id,
-      id: camera.id,
     }));
 
     setAvailableCameras(formattedCameras);
 
-    if (formattedCameras.length > 1) {
-      // if multiple cameras are available we set first one as active
-      setActiveCamera(formattedCameras[0]);
+    if (formattedCameras.length > 0) {
```
tuul-wq: Actually we need to set default camera if there is at least one camera

Asmadek (approved):

src/renderer/entities/transaction/ui/QrCode/QrReader/QrReaderWrapper.tsx
```diff
@@ -54,18 +54,15 @@ export const QrReaderWrapper = ({ className, onResult, countdown, validationErro
...
+      setActiveCamera(formattedCameras[0].value);
```
johnthecat: Once again - get first camera as variable and then check if it's nullish
```
const defaultCamera = formattedCameras.at(0);
if (defaultCamera) {...}
```

johnthecat (approved):

PR #2741, opened 2024-11-27
Feat: Gov/Fellowship selectors

tuul-wq:
- Selector for Gov
- Selector for Fellowship
- Simplified filter models for Gov and Fellowship
- Removed unnecessary constants
- Switched from `networkSelectorModel.$governanceChain` to `networkSelectorModel.$governanceChainId` where possible
- Turned `api?: ApiPromise` into `api: ApiPromise | null` for Fee components

part of #2589


Asmadek (approved):

src/renderer/entities/operations/ui/SignatorySelector.tsx
```diff
@@ -30,42 +30,45 @@ export const SignatorySelector = ({
...
+                    canCopy={false}
+                  />
+                  <AssetBalance value={balance.toString()} asset={asset} />
+                </div>
+              </Select.Item>
+            );
+          })}
+        </Select>
+      </Field>
       <InputHint variant="error" active={hasError}>
```
johnthecat, line 71: Why input hint isn't inside field? It feels more natural imo

src/renderer/features/governance/aggregates/delegatedVotes.ts
```diff
@@ -10,12 +10,14 @@ import { networkSelectorModel } from '../model/networkSelector';
 import { proposerIdentityAggregate } from './proposerIdentity';
 
 const $delegatedVotesInChain = combine(
-  delegatedVotesModel.$delegatedVotes,
```
johnthecat, line 13: Why you've changed so much unnecessary code?

johnthecat (changes requested):

src/renderer/features/governance/aggregates/delegatedVotes.ts
tuul-wq: In lots of cases `chainId` was enough, so I've switched from `chain` to the least meaningful chunk of data.
Also I've updated `combines` so they are more readable (top to bottom) and expandable if additional data would be required. Almost from the start of effector integration me and @Asmadek decided to stick [to this rule](https://arc.net/l/quote/azxszgrz) for `combine`
Hope this explanation helps 😉

johnthecat (approved):

PR #2754, opened 2024-11-28
fix: tests for chain name converter

### LINKED ISSUE #2743: OpenGov. Improve referendum URL UX
Update referenda URL:
1. Default chain must be Polkadot. If chain is not specified in URL then Polkadot must be used by default
2. For chain the chain name from chains.json may be used. The example is `/polkadot/<referenda_id>`

Asmadek:
* add tests for chain name converter

### DIFF
--- a/src/renderer/entities/network/lib/__tests__/network-utils.test.ts
+++ b/src/renderer/entities/network/lib/__tests__/network-utils.test.ts
@@ -1,10 +1,6 @@
 import { networkUtils } from '../network-utils';
 
 describe('entities/network/lib/network-utils', () => {
-  beforeEach(() => {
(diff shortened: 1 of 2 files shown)


src/renderer/entities/network/lib/__tests__/network-utils.test.ts
```diff
@@ -0,0 +1,20 @@
+import { networkUtils } from '../network-utils';
+
+describe('entities/network/lib/network-utils', () => {
+  beforeEach(() => {
```
johnthecat: do we need to restore mocks here?

johnthecat (approved):

tuul-wq (approved):

PR #2829, opened 2024-12-11
fix: support xcm v6 config & delivery fee

Asmadek:
* support xcm delivery fee
* support v3/v4 junctions
* support network name for junctions


src/renderer/shared/api/xcm/lib/xcm-utils.ts
```diff
@@ -152,48 +155,50 @@ function getConcreteAssetLocation(assetLocation?: LocalMultiLocation) {
...
   if (destinationParaId) {
-    return getChildLocation(destinationParaId, accountId);
+    return getChildLocation(destinationParaId, accountId, originChain.specName as Network);
   }
 
   return undefined;
 }
 
-function getAccountLocation(accountId?: AccountId) {
+function getAccountLocation(accountId?: AccountId, network?: string) {
```
tuul-wq: `network` seems too general coupled with `string` type. Have you tried something like this?
```javascript
function getAccountLocation<T extends Network>(accountId?: AccountId, network?: T) {
```

src/renderer/shared/lib/utils/substrate.ts
```diff
@@ -27,8 +27,7 @@ import { DEFAULT_TIME, ONE_DAY, THRESHOLD } from './constants';
 
 export type TxMetadata = { registry: TypeRegistry; options: OptionsWithMeta; info: BaseTxInfo };
 
-// TODO: Add V3, V4 support
-const SUPPORTED_VERSIONS = ['V2'];
+const SUPPORTED_VERSIONS = ['V3', 'V4'];
```
tuul-wq, line 30: 👍

src/renderer/widgets/Transfer/model/xcm-transfer-model.ts
```diff
@@ -165,13 +165,12 @@ const $txBeneficiary = combine(
     api: $api,
     destination: $destination,
     transferDirection: $transferDirection,
+    network: $networkStore,
   },
-  (params) => {
-    const { api, destination, transferDirection } = params;
-
+  ({ api, destination, transferDirection, network }) => {
     if (!api || !destination || !transferDirection) return undefined;
```
tuul-wq, line 170: If not time consuming, let's return `null` and remove `{ skipVoid: false }`

src/renderer/shared/api/xcm/lib/xcm-utils.ts
```diff
@@ -152,48 +155,50 @@ function getConcreteAssetLocation(assetLocation?: LocalMultiLocation) {
...
     parents: 0,
     interior: {
-      X1: {
-        [isEthereum ? 'accountKey20' : 'accountId32']: {
-          network: 'Any',
-          [isEthereum ? 'key' : 'id']: accountId,
+      X1: [
+        {
+          [isEthereum ? 'accountKey20' : 'accountId32']: {
+            network: network || 'Any',
```
stepanLav: Why do we choose specName to specify network here? looks like that parameter is used for another purpose in XcmV3JunctionNetworkId, most of XCMs sets here Null

johnthecat (approved):

src/renderer/entities/transaction/lib/transactionService.ts
```diff
@@ -347,6 +348,12 @@ async function getTxWeight(
   return getExtrinsicWeight(extrinsic, options);
 }
 
+function getTxBytesLength(transaction: Transaction, api: ApiPromise): number {
```
johnthecat: I think it's time to stop using `getExtrinsic` map. It always add more coupling and make it harder to decompose work with transactions later on

src/renderer/shared/api/xcm/lib/types.ts
```diff
@@ -43,6 +44,21 @@ export type NetworkBaseWeight = {
...
+export type DeliveryFeeConfig = {
+  type: 'exponential';
+  factorPallet: 'ParachainSystem' | 'XcmpQueue' | 'Dmp';
+  sizeBase: string;
+  sizeFactor: string;
+  alwaysHoldingPays: boolean;
+};
+
+export type NetworkDeliveryFee = {
+  [chainId: string]: {
```
johnthecat, line 56: why not `ChainId`?

src/renderer/shared/api/xcm/service/xcmService.ts
```diff
@@ -326,3 +329,50 @@ function decodeXcm(chainId: ChainId, data: XcmPalletPayload | XTokensPayload): D
...
+
+  const deliveryFeeConfig = config.networkDeliveryFee[originChain]?.[direction];
+
+  if (!deliveryFeeConfig) return new BN(0);
+
+  let deliveryFactor: string;
+
+  if (direction === 'toParent') {
+    deliveryFactor = (
+      await parentApi.query[toFirstCharLowercase(deliveryFeeConfig.factorPallet)].upwardDeliveryFeeFactor()
```
johnthecat: You can safely replace `toFirstCharLowercase ` with `camelCase` from `lodash`

src/renderer/shared/api/xcm/service/xcmService.ts
```diff
@@ -326,3 +329,50 @@ function decodeXcm(chainId: ChainId, data: XcmPalletPayload | XTokensPayload): D
...
+    ).toString();
+  } else {
+    deliveryFactor = (
+      await originApi.query[toFirstCharLowercase(deliveryFeeConfig.factorPallet)].deliveryFeeFactor(destinationChainId)
```
johnthecat: Same here

src/renderer/widgets/Transfer/model/form-model.ts
```diff
@@ -87,17 +91,26 @@ const $fee = restore(feeChanged, ZERO_BALANCE);
 const $multisigDeposit = restore(multisigDepositChanged, ZERO_BALANCE);
 const $isFeeLoading = restore(isFeeLoadingChanged, true);
 const $isXcm = createStore<boolean>(false);
+const $deliveryFee = createStore<string>(ZERO_BALANCE);
```
johnthecat: It's better to make it BN from the start

src/renderer/shared/api/xcm/lib/types.ts
Asmadek: Because in config chainid is in string format, not hex string

PR #2861, opened 2024-12-13
Fix: Connections number

### LINKED ISSUE #1456: The amount of connections isn't equal to the amount of success + connecting + error
## Description
image
An image sums it up. It is weird that the amount of the left isn't equal to the sum of the amounts on the right.
## Steps to Reproduce

1. go to settings>networks
2. see the issue

## Expected Behavior
Have the sum of the numbers on the right equal to the number on the left

## Actual Behavior
They are not equal

## Screenshots
See above

## Possible Solution
I looked into it and realized that we do not take into account the connections with the status `isDisconnectedStatus`. Also in the list we mark them as "Connecting..." which is misleading.

tuul-wq:
- Consistent network status

closes #1456

### DIFF
--- a/src/renderer/features/network/NetworkList/lib/networks-list-utils.ts
+++ b/src/renderer/features/network/NetworkList/lib/networks-list-utils.ts
@@ -21,17 +21,18 @@ function getExtendedChain(
 }
 
 type Metrics = Record<'success' | 'connecting' | 'error', number>;
+
 function getStatusMetrics(networkList: ExtendedChain[]): Metrics {
-  return networkList.reduce(
-    (acc, network) => {
-      if (networkUtils.isDisabledConnection(network.connection)) return acc;
+  const metrics = { connecting: 0, success: 0, error: 0 };
 
-      if (networkUtils.isConnectedStatus(network.connectionStatus)) acc.success += 1;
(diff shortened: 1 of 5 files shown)


src/renderer/features/network/NetworkList/lib/networks-list-utils.ts
```diff
@@ -21,17 +21,18 @@ function getExtendedChain(
...
+
+  for (const network of networkList) {
+    if (networkUtils.isDisabledConnection(network.connection)) continue;
 
-      if (networkUtils.isConnectedStatus(network.connectionStatus)) acc.success += 1;
-      if (networkUtils.isConnectingStatus(network.connectionStatus)) acc.connecting += 1;
-      if (networkUtils.isErrorStatus(network.connectionStatus)) acc.error += 1;
+    if (networkUtils.isConnectedStatus(network.connectionStatus)) metrics.success += 1;
+    if (networkUtils.isConnectingStatus(network.connectionStatus)) metrics.connecting += 1;
+    if (networkUtils.isDisconnectedStatus(network.connectionStatus)) metrics.connecting += 1;
```
tuul-wq, line 33: It's a state when connection is running but node tries to reconnect, which is similar to `connecting`
but is being invoked on `provider.disconnect` event

johnthecat (approved):

sokolova-an (approved):

PR #2864, opened 2024-12-16
fix: support caching of new Metadata_metadata_at_version method, fix of cache miss

johnthecat:


src/renderer/shared/api/network/provider/CachedProvider.ts
```diff
@@ -1,23 +1,32 @@
...
 
-      return hasMetadata && isMetadataMethod && !hasParams ? this.metadata : super.send(method, params, ...args);
+      if (method === 'state_getMetadata' && !hasParams && this.metadata) {
+        return Promise.resolve(this.metadata?.metadata);
+      }
+
+      if (method === 'state_call' && hasParams && this.metadata) {
+        const call = params[0];
+
+        if (call === 'Metadata_metadata_at_version') {
```
valentunn, line 59: I think we should also check the version of the cached metadata. Returning metadata in v14 format for the request that expects v15 might be a problem

src/renderer/entities/network/model/network-model.ts
```diff
@@ -367,12 +375,17 @@ sample({
 // =====================================================
 
 sample({
-  clock: createApiFx.doneData,
-  target: subscribeMetadataFx,
+  clock: createApiFx.done,
+  source: $metadata,
+  fn: (metadata, { params, result }) => ({
+    api: result,
+    cachedVersion: metadata.find((m) => m.chainId === params.chainId)?.runtimeVersion ?? null,
```
tuul-wq: Let's create a variable for `metadata.find` for readability

tuul-wq (approved):

PR #2875, opened 2024-12-18
fix: add pallet name for transfer fee

sokolova-an:


src/renderer/widgets/Transfer/model/form-model.ts
```diff
@@ -237,11 +238,20 @@ const $fakeTx = combine(
 
     const transactionType = network.asset.type ? TransferType[network.asset.type] : TransactionType.TRANSFER;
 
+    const palletName =
```
johnthecat, line 241: I think it's time to move all duplicates into separate method somewhere. Maybe 'chainsService' for now

johnthecat (approved):

tuul-wq (approved):

Asmadek (approved):

PR #2885, opened 2024-12-23
refactor: accounts and wallets separation

johnthecat:
This PR is the first step towards simplifying account management. The changes focus on three main areas:
* Introduced a `network` domain with an abstract accounts interface and logic.
* Replaced account management logic in the wallet model with the new implementation. I've removed instances where accounts were directly mutated, and instead now use proper model events.
* Removed `AccountId` type from `@/shared/core` and using the new one from `@/shared/polkadotjs-schemas`.

The new system introduces two generic interfaces: `ChainAccount` and `UniversalAccount`. A `UniversalAccount` is chain-agnostic and can be used with any chain that shares the same crypto type. A `ChainAccount`, as the name suggests, is chain-specific.

Key differences from the previous implementation:
* `type` field has been renamed to `accountType`. The `type` field is now used by the generic interfaces `ChainAccount` and `UniversalAccount`
* Removed the `chainType` field since `cryptoType` effectively handles chain type differentiation
* Added a `signingType` field, which mirrors the functionality from the wallet. 
* `signingExtras` fields have been moved from the generic type to the `WcAccount` type

The new account management logic now lives in a separate model. As a result, the `Wallet['accounts']` field is now derived and deprecated. Any code that needs to access accounts from the current wallet should use `walletModel.$activeAccounts` instead.


src/renderer/entities/wallet/lib/account-utils.ts
```diff
@@ -59,127 +63,158 @@ export const accountUtils = {
...
+  );
 }
 
-function isFlexibleMultisigAccount(account: Partial<Account>): account is FlexibleMultisigAccount {
-  return account.type === AccountType.FLEXIBLE_MULTISIG;
+function isRegularMultisigAccount(account: Partial<AnyAccount>): account is MultisigAccount {
+  return (
+    // @ts-expect-error Partial type breaks required type field usage
+    accountsService.isChainAccount(account) && 'accountType' in account && account.accountType === AccountType.MULTISIG
+  );
```
sokolova-an: we still have multi chain multisigs related to several chains at once will they work correctly with this type as well?

sokolova-an (approved):

PR #2922, opened 2024-12-31
feat: change flexible icon

sokolova-an:
Updated proxy fee label
Added inactive flexible multisig icon


src/renderer/entities/wallet/ui/Cards/WalletCardLg.tsx
```diff
@@ -19,10 +19,15 @@ export const WalletCardLg = ({ wallet, description, full, className }: Props) =>
 
   const isWalletConnect = walletUtils.isWalletConnectGroup(wallet);
 
+  const type =
+    walletUtils.isFlexibleMultisig(wallet) && !wallet.accounts.at(0)?.proxyAccountId
```
johnthecat: let's check `wallet.activated` flag instead

src/renderer/shared/core/types/wallet.ts
```diff
@@ -113,3 +113,7 @@ export const enum SigningType {
   POLKADOT_VAULT = 'signing_pv',
   WALLET_CONNECT = 'signing_wc',
 }
+
+export enum WalletIconType {
```
johnthecat, line 130: Why we need new enum here?

sokolova-an: To match type\structure of `WalletType` but use it separately from that

johnthecat (approved):

PR #3115, opened 2025-02-05
feat: basket operations aggregate

Asmadek:
* basket operations aggregate
* use this aggregate on basket page and save tx to basket


tuul-wq (approved):

src/renderer/aggregates/basket-operations/model.ts
```diff
@@ -0,0 +1,52 @@
+import { createEffect, createStore, sample } from 'effector';
+import { readonly } from 'patronum';
+
+import { storageService } from '@/shared/api/storage';
+import { type BasketTransaction } from '@/shared/core';
+
+const $basketTransactions = createStore<BasketTransaction[]>([]);
+
+const populateFx = createEffect((): Promise<BasketTransaction[]> => storageService.basketTransactions.readAll());
```
johnthecat: You don't need to specify return type

src/renderer/aggregates/basket-operations/model.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const addTransactionsFx = createEffect(
+  async (transactions: BasketTransaction[]): Promise<BasketTransaction[] | undefined> => {
```
johnthecat: Always return array, this undefined just add complexity to result handling

src/renderer/aggregates/basket-operations/model.ts
```diff
@@ -0,0 +1,52 @@
...
+    return storageService.basketTransactions.createAll(transactions);
+  },
+);
+
+const updateTransactionsFx = createEffect((transactions: BasketTransaction[]): Promise<number[] | undefined> => {
```
johnthecat: same here - don't use undefined as result

src/renderer/aggregates/basket-operations/model.ts
```diff
@@ -0,0 +1,52 @@
...
+  return storageService.basketTransactions.updateAll(transactions);
+});
+
+const removeTransactionsFx = createEffect((transactions: BasketTransaction[]): Promise<number[] | undefined> => {
```
johnthecat, line 19: same here

PR #3150, opened 2025-02-20
Update NOTICE.md

taozui472:
update the copyright year to 2025


stepanLav:
Thanks for the PR, but we do not need that

PR #3202, opened 2025-02-20
Release/1.14.4 2

stepanLav:


src/renderer/shared/lib/utils/assets.ts
```diff
@@ -32,6 +34,31 @@ export const getAssetById = (id: string, assets?: Asset[]): Asset | undefined =>
...
+  return (
+    assets.find((asset) => {
+      if (!asset.typeExtras) return;
+
+      if ('assetId' in asset.typeExtras) {
+        return asset.typeExtras.assetId === assetId;
+      }
+
+      const id = api.createType(asset.typeExtras.currencyIdType, asset.typeExtras.currencyIdScale).toJSON();
+      const currencyId = api.createType(asset.typeExtras.currencyIdType, assetId).toJSON();
```
valentunn, line 55: This can be done only once and not for each iteration of the find loop

PR #3229, opened 2025-02-25
refactor: changed address to account id in transactions

johnthecat:
Transaction type now have accountId instead of address field. It simplifies transaction setup and unlocks complex account structure feature


src/renderer/features/multisig-wallet/index.tsx
```diff
@@ -37,6 +37,25 @@ accountSDK(multisigWalletFeature, {
...
+    //     type: TransactionType.MULTISIG_AS_MULTI,
+    //     args: {
+    //       threshold: account.threshold,
+    //       otherSignatories,
+    //       maybeTimepoint: null,
+    //       callData,
+    //       callHash,
+    //     },
+    //   };
+    // }
```
sokolova-an, line 56: Why is this commented on?

src/renderer/entities/transaction/lib/transactionBuilder.ts
```diff
@@ -301,7 +297,7 @@ type DelegateParams = {
 function buildDelegate({ chain, accountId, tracks, target, conviction, balance }: DelegateParams): Transaction {
   const delegateTxs = tracks.map((track) => ({
     chainId: chain.chainId,
-    address: toAddress(accountId, { prefix: chain.addressPrefix }),
+    accountId: accountId,
```
sokolova-an:
```suggestion
    accountId,
```

here and below

sokolova-an (approved):
cool!

src/renderer/features/multisig-wallet/index.tsx
johnthecat: This is hypothetical implementation of the wrapper, but now we don't have callData in transaction. So it's basically a large TODO

tuul-wq (approved):

PR #3266, opened 2025-03-03
feat: use indexer for multisig operations

### LINKED ISSUE #2900: Multisig. Can't decode multisig batch (batchAll) operation with unknown extrinsics
Multisig. Can't decode multisig batch (batchAll) operation with unknown extrinsics.

### LINKED ISSUE #3049: Multisig. Improve subquery discovery
## Description

multisig.asmulti() inside the utility.batch() call won't be parsed by subquery proxject, due to handlers realisation.

Would be better to have event handler on multisig.NewMultisig and then use call visitor in order to parse block and all extrinsics https://github.com/novasamatech/subquery-call-visitor

Example of that block:
https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fwestend.dotters.network#/explorer/query/0x323a3c2ac7398acc6f05c766fae1d3843f81d8bc53d9f95096d7d554ef977294

### LINKED ISSUE #3112: Multisig operations. Use indexer for getting multisig operations
### The problem
There are 3 problems with current approach for getting and decoding multisig operations:
1. Multisig operations that are executed or rejected won't be added into the operations list because they are removed from multisig pallet storage.
2. Multisig operations call data may be retrieved only from the archive node if it was created more then 30 blocks ago (~180 seconds on Polkadot)
3. If runtime upgrade happened before the user gets the multisig operation it may not be decoded.

### Proposed solution
Use indexer for getting operation list.
Pros:
1. The user will get all multisig operations including executed and rejected
2. The user will get correct block number when the operation was executed, rejected, signed
3. The call data and decoded call data will be always provided for exectued transaction
4. The call data will be always provided for ongoing and rejected operations if call data was sent to blockchain
5. The call data decoding will be decoupled from the runtime update
6. If multisig initiated with batch, proxy call it will be decoded
Cons:
1. Indexer may not be available
2. Indexer may be attacked and the user will get wrong decoded operation information

### How to avoid security risks
1. Compare call hash and hash(call data)
2. For ongoing operations check if operation with call hash exists on the blockchain in multisig pallet storage

### DOD
- [ ] In progress operations get from the chain in order to show as a counter for Operations menu item
- [ ] Operation parsing
  - [ ] Simple operations parsed in detailed operation, transfer, vote, stake, etc.
  - [ ] Complex operations shows by first call with additional row for parsed js object
  - [ ] Link to subscan operation
- [ ] Call data parser should parse operation if didn't get subquery data or it is a different operation for multisig creation

### LINKED ISSUE #3139: Multisig. Unknown operations parsed wrong
## Description

Image

## Test data

[spektr-database.json](https://github.com/user-attachments/files/18824295/spektr-database.json)

## Environment
https://github.com/novasamatech/nova-spektr/commit/aeb8b7e10ed215d580dc577596b797088b15354c

Asmadek:
* Remove call data decoder
* Remove multisig context
* Use indexer to get multisig if exists
* Use `OperationData` for multisig instead of `Transaction`


src/renderer/aggregates/basket-operations/service.ts
```diff
@@ -9,7 +9,9 @@ const getCoreTx = (tx: BasketTransaction): Transaction => {
     return tx.coreTx;
   }
 
-  return tx.coreTx.type === TransactionType.BATCH_ALL ? findCoreBatchAll(tx.coreTx) : tx.coreTx;
+  return tx.coreTx.type === TransactionType.BATCH_ALL
+    ? (findCoreBatchAll(tx.coreTx) as unknown as Transaction)
```
aaa123eee: is it necessary to cast to unknown here?

src/renderer/entities/chain/ui/OperationTitle/lib/operation-title.ts
```diff
@@ -6,5 +6,7 @@ export const getCoreTx = (tx: BasketTransaction): Transaction => {
     return tx.coreTx;
   }
 
-  return tx.coreTx.type === TransactionType.BATCH_ALL ? findCoreBatchAll(tx.coreTx) : tx.coreTx;
+  return tx.coreTx.type === TransactionType.BATCH_ALL
+    ? (findCoreBatchAll(tx.coreTx) as unknown as Transaction)
```
aaa123eee: same question, should be fine without such a casting

at least, without unknown it will fire us an error once it becomes a nonsense

src/renderer/app/App.tsx
```diff
@@ -38,12 +38,10 @@ export const App = () => {
...
 
   return (
-    <MultisigChainProvider>
-      <ConfirmDialogProvider>
-        <StatusModalProvider>
-          <GraphqlProvider>{appRoutes}</GraphqlProvider>
-        </StatusModalProvider>
-      </ConfirmDialogProvider>
-    </MultisigChainProvider>
+    <ConfirmDialogProvider>
```
tuul-wq, line 41: ❤️

src/renderer/entities/multisig/lib/multisigTx/common/utils.ts
```diff
@@ -42,167 +25,50 @@ export const getPendingMultisigTxs = async (
...
+    id: eventId,
+    accountId: multisigTx.accountId,
+    extrinsicHash: params.extrinsicHash,
+    blockCreated: pjsSchema.helpers.toBlockHeight(params.timepoint.height),
+    indexCreated: params.timepoint.index,
+    timestamp: Date.now(),
+    status: 'approve',
+  };
+
+  const transaction: MultisigOperation = {
```
tuul-wq: `return {...}` without a variable

src/renderer/features/governance/model/lockPeriods.ts
```diff
@@ -30,11 +25,27 @@ sample({
...
+  },
+  fn({ apis, chains }, { transaction }) {
+    const chain = chains[transaction.chainId];
+    const api = apis[transaction.chainId];
+
+    return {
+      api,
+      chain,
+    };
+  },
```
tuul-wq: Can do this, if you don't mind
```suggestion
  fn: ({ apis, chains }, { transaction }) => ({
    api: apis[transaction.chainId],
    chain: chains[transaction.chainId],
  }),
```

src/renderer/features/multisig-operations/components/Operation.tsx
```diff
@@ -1,39 +1,55 @@
...
-          <OperationTitleStatus operation={tx} />
-        </div>
-      </Accordion.Button>
-      <Accordion.Content className="border-t border-divider">
-        <OperationFullInfo tx={tx} account={account} />
-      </Accordion.Content>
-    </Accordion>
-  );
-});
+const Operation = memo(
```
tuul-wq: 👍

tuul-wq (approved):
Great job! Left minor suggestions

PR #3296, opened 2025-03-06
fix: qr reader stuck

stepanLav:
1. Made the function async to properly handle asynchronous operations
2. Added proper error handling with try/catch
3. Used Promise.all to ensure all stop operations complete before proceeding
4. Added .filter(Boolean) to handle potential undefined/null controls
5. Clear the stream reference after stopping
6. Handle any pending ready promises on tracks


src/renderer/entities/transaction/ui/QrCode/QrReader/QrSignatureReader.tsx
```diff
@@ -111,57 +111,41 @@ export const QrSignatureReader = ({
...
-
-    (async () => {
-      try {
-        controlsRef.current?.stop();
-        bgControlsRef.current?.stop();
-        await startScanning();
-      } catch {
-        onError?.(QR_READER_ERRORS[QrError.BAD_NEW_CAMERA]);
-      }
-    })();
```
stepanLav, line 164: the main issue was that startScanning() called twice which run 2 usage objects. when successfully scan - closed only one and the second one remain opened. It become worse, when user make operation multiple times

src/renderer/entities/transaction/ui/QrCode/QrReader/QrReader.tsx
```diff
@@ -337,7 +331,7 @@ export const QrReader = ({
     })();
 
     return () => {
-      stopScanning();
+      handleStopScanning();
```
tuul-wq: We can simply call `stopScanning({ streamRef, controlsRef, bgControlsRef });`
No need for a separate function
Same for other places

src/renderer/entities/transaction/ui/QrCode/QrReader/QrSignatureReader.tsx
```diff
@@ -111,57 +111,41 @@ export const QrSignatureReader = ({
...
+            delayBetweenScanSuccess: 50,
+          });
+
+          if (!cameraId && (!camerasAmount || camerasAmount === 1)) {
+            await startScanning();
+          }
         }
       } catch (error) {
-        onError?.(error as ErrorObject);
+        {
```
tuul-wq: No reason for double parenthesis
```javascript
catch (error) {
  onError?.(error as ErrorObject);
}
```

tuul-wq (commented):
Small comments
Looking into `stopScanning `...

aaa123eee (changes requested):
image

src/renderer/shared/ui-kit/QrReader/helpers.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export const requestAccess = async () => {
+  return navigator.mediaDevices.getUserMedia({ video: true }).catch(() => {
+    throw QR_READER_ERRORS[QrReaderErrorCode.USER_DENY];
+  });
+};
+
+export const stopMediaStream = async (stream: MediaStream): Promise<void> => {
+  try {
+    if (stream) {
```
tuul-wq: Can do fast return before `try` and reduce nesting
```javascript
if (!stream) return;
```

tuul-wq (approved):

aaa123eee (approved):

PR #3349, opened 2025-03-18
feat: transfer dynamic state & reflections

### LINKED ISSUE #3338: Transfer. Available balance won't update if that data changed on blockchain.
## Description

When user opens Transfer window, available balance will snapshotted, if it changes on blockchain user won't get updates.

Additional items:

- Continue button should be connected with network fee and available balance.
- Add shimmer for available balacne

## Example:

https://github.com/user-attachments/assets/331289c5-4dde-4bbb-80ae-b7f1d477caa3

## Steps to Reproduce

1. Open transfer
2. Change balance for the address in 3rd app

## Expected Behavior
Available balance updated based on subscription

## Actual Behavior
Balance data - snapshotted

## Environment
https://github.com/novasamatech/nova-spektr/commit/0b00a30edb09f51d38e5d4be86be080b98acb3a9

aaa123eee:
- [x] nullable balances
- [x] skeleton for balance
- [x] disable submit until balance available

closes #3338

### DIFF
--- a/src/renderer/features/assets-balances/components/AmountInput.tsx
+++ b/src/renderer/features/assets-balances/components/AmountInput.tsx
@@ -28,7 +28,7 @@ type Props = {
   disabled?: boolean;
   asset: Asset;
   balancePlaceholder?: string;
-  balance: string | string[] | null;
+  balance?: string | string[] | null;
   invalid?: boolean;
   showCurrency?: boolean;
   onChange?: (value: string) => void;
--- a/src/renderer/features/assets-balances/components/AmountInput.tsx
+++ b/src/renderer/features/assets-balances/components/AmountInput.tsx
@@ -28,7 +28,7 @@ type Props = {
   disabled?: boolean;
   asset: Asset;
   balancePlaceholder?: string;
-  balance: string | string[] | ReactNode | null;
+  balance: string | string[] | null;
   invalid?: boolean;
   showCurrency?: boolean;
   onChange?: (value: string) => void;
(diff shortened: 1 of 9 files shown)


tuul-wq (approved):

src/renderer/features/assets-balances/components/AmountInput.tsx
```diff
@@ -27,7 +28,7 @@ type Props = {
   disabled?: boolean;
   asset: Asset;
   balancePlaceholder?: string;
-  balance?: string | string[] | ReactNode;
+  balance: string | string[] | ReactNode | null;
```
aaa123eee: I wasn't sure about this one

We might also want to pass skeleton when null, please tell your opinion

tuul-wq: Good question. I see that in some places we do use Skeleton as a `balance` prop when something is loading.
Would be great to remove `ReactNode` type for balance and allow `AmountInput` to display `Skeleton` by itself. Seems that `null` is a valid signal to show the `Skeleton`.
At the moment we set `ZERO_BALANCE` thus form displays `0 DOT`, and we'd rather have `Skeleton` until everything (balance) has been loaded.
```javascript
balance={isStakingLoading ? <Shimmering width={50} height={10} /> : withdrawBalance}
```

src/renderer/features/operations/OperationsValidation/lib/transfer-rules.ts
```diff
@@ -91,10 +91,12 @@ export const TransferRules = {
       validator: (
         amount: string,
         _: any,
-        { network, balance }: { network: NetworkStore | null; balance: BalanceMap },
+        { network, balance }: { network: NetworkStore | null; balance?: BalanceMap },
       ) => {
         if (!network) return false;
 
+        if (!balance) return true;
```
johnthecat, line 98: As i understand it, you skip validation if there is no balance. IMO it's better to fail like it's done with `networks` couple lines before

aaa123eee: the rule is named `notEnoughBalance`

I consider empty balance as not enough, that's why it returns true

PR #3394, opened 2025-03-20
feat: fellowship sorting algorithm

### LINKED ISSUE #3358: Fellowship. Sorting tags and Sorting rules
We need to create a new tags for activities and apply new sorting rules.

## Rules description:

Based on first implementation was built that conception:
https://gist.github.com/valentunn/e4cd82ffb5a64692e963c6689ff0001e

But it was simplified to tags rules, shown on figma screen. Lets take some calculations from first implementation and use it in tags approach.

### 1. Kingmaker

- Measures individual voter influence

**Maximum (1.0)**: User holds all or nearly all voting power
**Minimum (0.0)**: User has negligible voting power

**_Question - how to get all available votes for referenda?_**

### 2. Flashpoint 
- Measures how contested the vote is based on two factors:
  - Participation rate (fraction of total possible votes cast)
  - How close the vote is to 50-50 split

**Maximum (1.0)**: High participation with votes very close to 50-50
**Minimum (0.0)**: No votes cast or votes heavily skewed to one side

### 3. Last Call
 - Applied when 20% or less of total voting time remains
 - Is only added if there are no other tags

### 4. Tiebreaker
won't implement on that stage

### Expire soon
On top of that tags should be applied Expire soon, except Voted.
Expire soon - should be added if remaining time less than 3 days and score should be higher than less time remain.
1.0 - (time_left / three_days_seconds)
That tag should be add on top and sort by remaining time inside each group

Mockups link:

https://www.figma.com/design/uBSPUDFpAkKjYNpA9XTXzh/OpenGov-fellowship?node-id=3214-63979&t=yYPqZCFYuu6YfeOG-4

## DOD:
- [ ] Answer all questions, add in description how it works
- [ ] Comunicate with Sabit about implementation

johnthecat:
## Description
First draft implementation of sorting algo in fellowship referendums

Closes #3358

### DIFF
--- a/src/renderer/domains/collectives/referendum/resource.ts
+++ b/src/renderer/domains/collectives/referendum/resource.ts
@@ -166,12 +166,12 @@ export async function mapReferendums(
 ) {
   const undecidingTimeout = referendaPallet.consts.undecidingTimeout(palletType, api);
   const tracks = referendaPallet.consts.tracks(palletType, api);
-  const value: Referendum[] = [];
+  const referendums: Referendum[] = [];
   for (const { id, info } of list) {
     if (!info) continue;
-    const record = await mapReferendum({ id, info, tracks, undecidingTimeout, api });
-    value.push(record);
+    const referendum = await mapReferendum({ id, info, tracks, undecidingTimeout, api });
+    referendums.push(referendum);
   }
 
-  return value;
+  return referendums;
 }
--- a/src/renderer/domains/collectives/referendum/resource.ts
+++ b/src/renderer/domains/collectives/referendum/resource.ts
@@ -5,10 +5,13 @@ import { type HexString } from '@/shared/core';
 import { nullable } from '@/shared/lib/utils';
 import {
   type FrameSupportPreimagesBounded,
+  type ReferendaDecidingStatus,
   type ReferendaReferendumInfoConvictionVotingTally,
   type ReferendumId,
+  referendaPallet,
 } from '@/shared/pallet/referenda';
-import { pjsSchema } from '@/shared/polkadotjs-schemas';
+import { type BlockHeight, pjsSchema } from '@/shared/polkadotjs-schemas';
+import { type CollectivePalletsType } from '../_lib/types';
 
 import { type Proposal, type Referendum } from './types';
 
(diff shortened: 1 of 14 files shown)


src/renderer/domains/collectives/referendum/resource.ts
```diff
@@ -76,23 +79,50 @@ async function parseProposal(proposal: FrameSupportPreimagesBounded, api: ApiPro
...
-  info: ReferendaReferendumInfoConvictionVotingTally,
-  api: ApiPromise,
-): Promise<Referendum> {
+function getEndBlock(
+  deciding: ReferendaDecidingStatus | null,
+  submitted: BlockHeight,
+  decisionPeriod: BlockHeight,
+  undecidingTimeout: BlockHeight,
+) {
+  if (deciding && deciding.confirming) return deciding.confirming;
```
tuul-wq:
```suggestion
  if (deciding?.confirming) return deciding.confirming;
```

src/renderer/domains/collectives/referendum/resource.ts
```diff
@@ -103,35 +133,43 @@ async function mapReferendum(
...
 }
 
 export async function mapReferendums(
   list: { id: ReferendumId; info: ReferendaReferendumInfoConvictionVotingTally | null }[],
   api: ApiPromise,
+  palletType: CollectivePalletsType,
 ) {
+  const undecidingTimeout = referendaPallet.consts.undecidingTimeout(palletType, api);
+  const tracks = referendaPallet.consts.tracks(palletType, api);
   const value: Referendum[] = [];
```
tuul-wq: English is poor in words but we still have some 😉
```suggestion
  const referendums: Referendum[] = [];
```

src/renderer/features/fellowship-tasks/model/tasks.ts
```diff
@@ -172,34 +178,65 @@ const $evidenceTasks = combine(
...
         : 'other';
     });
 
+    const getWeight = (referendum: OngoingReferendum) => {
+      const maximumAvailableVotingWeight = tasksService.getMaximumAvailableVotingWeight(
+        members,
+        maxRank,
+        referendum.track,
+      );
+      const memberVotingWeight = trackService.getVoteWeight({
```
tuul-wq, line 205: Spacing between expressions would help a bit

src/renderer/features/fellowship-tasks/service.ts
```diff
@@ -0,0 +1,153 @@
...
+    tags.push('controversial');
+  }
+  if (isImportantVote) {
+    tags.push('importantVote');
+  }
+
+  return {
+    sortingScore,
+    tags,
+  };
```
tuul-wq, line 130: Personally I like inline objet with a couple properties, can skip it if you don't like
```suggestion
  return { sortingScore, tags };
```

src/renderer/features/fellowship-tasks/service.ts
```diff
@@ -0,0 +1,153 @@
...
+}
+
+function getMaximumAvailableVotingWeight(members: Member[], maxRank: number, track: TrackId): number {
+  let max = 0;
+
+  for (const member of members) {
+    if (trackService.rankSatisfiesVotingThreshold(member.rank, maxRank, track)) {
```
tuul-wq, line 137: To reduce nesting
```suggestion
    if (!trackService.rankSatisfiesVotingThreshold(member.rank, maxRank, track)) {
      continue;
    }
```

johnthecat: I think such long lines with negotiation at the start are harder to read. Early exit is not always a good idea.

PR #3449, opened 2025-03-27
fix: votes impact and small UI fixes

### LINKED ISSUE #3440: Fellowship. Vote via likes - shouldn't shows impact if threshold one side
## Description

When user try to vote on one sided referenda - there is shouldn't be changes on threshold line

Image
Image

## Environment
https://github.com/novasamatech/nova-spektr/commit/1fc39d8371e7af52bf37caed02ed4967063301eb

sokolova-an:
## Description
Rework progress bar for voting
Small UI fixes along the way
## Related Issues
#3440
## Screenshots/Recordings
image
image

### DIFF
--- a/src/renderer/features/fellowship-tasks/components/tasks/RequestRetention.tsx
+++ b/src/renderer/features/fellowship-tasks/components/tasks/RequestRetention.tsx
@@ -1,10 +1,12 @@
 import { useUnit } from 'effector-react';
+import { useEffect, useState } from 'react';
 
 import { Slot, createSlot } from '@/shared/di';
 import { useI18n } from '@/shared/i18n';
-import { toRomanNumeral } from '@/shared/lib/utils';
+import { getCreatedDateFromApi, toRomanNumeral } from '@/shared/lib/utils';
 import { FootnoteText, SmallTitleText } from '@/shared/ui';
 import { Box } from '@/shared/ui-kit';
+import { fellowshipTasksFeature } from '../../model/feature';
 import { periods } from '../../model/periods';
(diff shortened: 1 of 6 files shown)


src/renderer/features/fellowship-tasks/components/tasks/RequestRetention.tsx
```diff
@@ -1,31 +1,46 @@
...
   return (
     <Box direction="row" padding={4} gap={5} verticalAlign="flex-end">
       <Box gap={3} grow={1}>
         <SmallTitleText>{t('fellowship.tasks.task.retention.title')}</SmallTitleText>
         <FootnoteText>
           {t('fellowship.tasks.task.retention.description', { rank: toRomanNumeral(track?.id ?? 0) })}
         </FootnoteText>
+        <FootnoteText className="text-text-secondary">
+          {t('fellowship.tasks.task.retention.until', {
+            date: formatDate(periodEnd, 'dd.MM.yyyy'),
```
aaa123eee, line 39: can it improperly render current date when periodEnd has initial value `0`? maybe null as a pending state can help us to avoid

aaa123eee (approved):

johnthecat (approved):

src/renderer/features/fellowship-tasks/components/tasks/RequestRetention.tsx
sokolova-an: it shouldn't since we take time from API but if there's no response - we can confirm with design what should we show in this case

PR #3547, opened 2025-04-10
fix: balances for pjs chain account

### LINKED ISSUE #3538: Portfolio view. Balances are not loading for PJS extension wallet
## Description

When user select pjs extension - there is endless loading state for balances on Portfolio view

## Test data

[spektr-database (1).json](https://github.com/user-attachments/files/19363276/spektr-database.1.json)

## Steps to Reproduce

1. Import test database
2. Open Portfolio view

## Expected Behavior

Shows balances

## Actual Behavior

Endless loading state

## Screenshots

image

aaa123eee:
## Description

Request balances for single chain PJS account

## Related Issues

closes #3538

## Changes Made

- support for all crypto types


src/renderer/domains/network/account/service.ts
```diff
@@ -34,9 +34,12 @@ function uniqId(account: AnyAccountDraft) {
 }
 
 function isCryptoMatch(account: Pick<AnyAccount, 'cryptoType'>, chain: Chain): boolean {
-  const cryptoType = networkUtils.isEthereumBased(chain.options) ? CryptoType.ETHEREUM : CryptoType.SR25519;
+  const isEthereumBased = networkUtils.isEthereumBased(chain.options);
+  if (isEthereumBased) {
+    return [CryptoType.ED25519, CryptoType.ECDSA, CryptoType.ETHEREUM].includes(account.cryptoType);
```
valentunn: Can you please explain why do check for ED25519 and ECDSA?

* There is no way ED25519 is valid for Ethereum crypto
* For ECDSA I imagine some signers mistekently return ECDSA instead of  ETHERUM for Evm-addressable Substrate chains?

src/renderer/domains/network/account/service.ts
```diff
@@ -34,9 +34,12 @@ function uniqId(account: AnyAccountDraft) {
...
+  }
 
-  return account.cryptoType === cryptoType;
+  return account.cryptoType === CryptoType.SR25519;
```
valentunn: Why do we hard-code SR25519 here? Does it match with how we create accounts?
Overall, for a substrate chain, any of SR25519, ED25519 and ECDSA is a valid crypto type

What is the main usage of `isCryptoMatch`?

aaa123eee: because it is the only left after previous check

src/renderer/domains/network/account/service.ts
aaa123eee: Exactly, I faced ECDSA acc that couldn't pass this check

src/renderer/domains/network/account/service.ts
valentunn: Previuos check only happens when `isEthereum` is true. So when `isEthreum` is false we cannot guarantee that it is not ECDSA or ED22519

src/renderer/domains/network/account/service.ts
```diff
@@ -34,9 +34,12 @@ function uniqId(account: AnyAccountDraft) {
 }
 
 function isCryptoMatch(account: Pick<AnyAccount, 'cryptoType'>, chain: Chain): boolean {
-  const cryptoType = networkUtils.isEthereumBased(chain.options) ? CryptoType.ETHEREUM : CryptoType.SR25519;
+  const isEthereumBased = networkUtils.isEthereumBased(chain.options);
+  if (isEthereumBased) {
```
tuul-wq: `isEthereumBased` is used only once, can omit this variable
```suggestion
  if (networkUtils.isEthereumBased(chain.options)) {
```

tuul-wq (approved):

johnthecat (approved):

PR #3559, opened 2025-04-13
feat: handle chain account connect from extension

### LINKED ISSUE #3539: Extension. Handle single chain accounts
## Description

When connect a single chain account via PJS, universal account is connected instead

## Steps to Reproduce

1. Setup PJS account for 1 chain only
2. Add this account to spektr
3. Observe connection screen
4. Check account type in DB

## Expected Behaviour

- connection screen shows only related chains
- added account has type=chain 

## Actual Behaviour

- on connection screen the list of all available chains is showed instead of 1 chain
- added account has type=universal 

## Test data

[spektr-database (1).json](https://github.com/user-attachments/files/19363276/spektr-database.1.json)

aaa123eee:
## Description

Handle single chain accounts from PJS

## Related Issues

closes #3539

## Changes Made

- adjust pairing form to show only account chain
- filter data related for the account chain
- update feature injection using accountSDK

## Screenshots/Recordings

image

image

### DIFF
--- a/src/renderer/entities/wallet/lib/account-utils.ts
+++ b/src/renderer/entities/wallet/lib/account-utils.ts
@@ -121,19 +121,14 @@ function isAccountWithShards(accounts: AnyAccount | VaultShardAccount[]): accoun
   return Array.isArray(accounts) && isVaultShardAccount(accounts[0]);
 }
 
-function isChainDependant(account: AnyAccountDraft): boolean {
+function isChainDependant(account: AnyAccountDraft) {
   return accountService.isChainAccount(account);
 }
 
 function isChainIdMatch(account: AnyAccount, chainId: ChainId): boolean {
   if (!isChainDependant(account)) return true;
 
-  const chainAccountMatch = isVaultChainAccount(account) && account.chainId === chainId;
-  const shardAccountMatch = isVaultShardAccount(account) && account.chainId === chainId;
-  const wcAccountMatch = isWcAccount(account) && account.chainId === chainId;
-  const proxiedAccountMatch = isProxiedAccount(account) && account.chainId === chainId;
-
-  return chainAccountMatch || wcAccountMatch || shardAccountMatch || proxiedAccountMatch;
+  return account.chainId === chainId;
 }
 
 function isChainAndCryptoMatch(account: AnyAccount, chain: Chain): boolean {
--- a/src/renderer/entities/wallet/lib/account-utils.ts
+++ b/src/renderer/entities/wallet/lib/account-utils.ts
@@ -22,7 +22,7 @@ import { toAddress } from '@/shared/lib/utils';
 import { type AccountId } from '@/shared/polkadotjs-schemas';
 // TODO all this type checks should be defined in features with own context
 // eslint-disable-next-line boundaries/element-types
-import { type AnyAccount, type AnyAccountDraft, accountService } from '@/domains/network';
+import { type AnyAccount, accountService } from '@/domains/network';
 import { networkUtils } from '@/entities/network';
 
 import { walletUtils } from './wallet-utils';
@@ -37,7 +37,6 @@ export const accountUtils = {
   isProxiedAccount,
   isPureProxiedAccount,
 
-  isChainDependant,
   isChainIdMatch,
   isChainAndCryptoMatch,
   isAccountWithShards,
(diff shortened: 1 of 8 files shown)


src/renderer/features/extension-wallet/model/pairingForm.ts
```diff
@@ -10,14 +10,20 @@ import { type AnyAccountDraft } from '@/domains/network';
...
+  type ExtensionUniversalAccount,
+} from '../types';
 
 import { wallets } from './wallets';
 
 type Step = 'idle' | 'pairing' | 'select' | 'rejected' | 'success';
 
 type ConnectedAccount = WalletAccount & {
   type: 'sr25519' | 'ed25519' | 'ecdsa' | 'ethereum';
+  genesisHash?: `0x${string}` | '';
```
johnthecat: `HexString` type instead of ``0x${string}``

johnthecat (approved):

src/renderer/features/extension-wallet/model/pairingForm.ts
aaa123eee: we can get rid of this override once this PR in talisman connect is merged https://github.com/TalismanSociety/talisman-connect/pull/63

src/renderer/entities/wallet/lib/account-utils.ts
```diff
@@ -121,19 +121,14 @@ function isAccountWithShards(accounts: AnyAccount | VaultShardAccount[]): accoun
   return Array.isArray(accounts) && isVaultShardAccount(accounts[0]);
 }
 
-function isChainDependant(account: AnyAccountDraft): boolean {
+function isChainDependant(account: AnyAccountDraft) {
```
tuul-wq: Maybe completely remove `isChainDependant` and use `accountService.isChainAccount` where needed?

src/renderer/features/assets/AssetsPortfolioView/model/portfolio-model.ts
```diff
@@ -75,9 +76,17 @@ const $activeTokens = combine(
         if (nullable(connection)) return false;
         if (networkUtils.isDisabledConnection(connection)) return false;
         if (nullable(chains[c.chainId])) return false;
-        if (!isMultisigWallet) return true;
+        if (isMultisigWallet) {
+          return networkUtils.isMultisigSupported(chains[c.chainId].options);
+        }
+
+        return selectedAccounts.some((acc) => {
+          if (accountService.isUniversalAccount(acc)) {
```
tuul-wq:
```suggestion
          return accountService.isUniversalAccount(acc) || acc.chainId === c.chainId;
```

src/renderer/features/extension-wallet/model/pairingForm.ts
```diff
@@ -57,17 +63,32 @@ const WalletTypeFromExtension = (extension: ExtensionType) => {
 const $accounts = combine($rawAccounts, $extensionType, (accounts, extensionType) => {
   if (nullable(extensionType)) return [];
 
-  return accounts.map<AccountDraft>(({ address, type, name }) => {
-    return {
+  return accounts.map<AccountDraft>(({ address, type, name, genesisHash }) => {
+    const isChainAccount = !!genesisHash;
```
tuul-wq, line 67:
```suggestion
    const isChainAccount = nonNullable(genesisHash);
```

src/renderer/features/extension-wallet/model/pairingForm.ts
```diff
@@ -57,17 +63,32 @@ const WalletTypeFromExtension = (extension: ExtensionType) => {
...
+      type: isChainAccount ? 'chain' : 'universal',
       signingType: SigningType.EXTENSION,
     };
+
+    if (isChainAccount) {
+      return {
+        ...baseAccount,
+        type: 'chain',
+        chainId: genesisHash,
+      } satisfies AnyAccountDraft<ExtensionChainAccount>;
```
tuul-wq, line 85: Why not `as`? I assume it should work as well

src/renderer/features/extension-wallet/types.ts
```diff
@@ -18,7 +18,14 @@ export type SubWalletExtensionWallet = Wallet & {
   extension: ExtensionType;
 };
 
-export interface ExtensionAccount extends UniversalAccount {
+export type ExtensionAccount = ExtensionUniversalAccount | ExtensionChainAccount;
```
tuul-wq: Could we move this line after `export interface ExtensionChainAccount extends ChainAccount {`?
So it reads top down like a book 😄

tuul-wq (approved):
Little suggestions

src/renderer/entities/wallet/lib/account-utils.ts
aaa123eee: I had the same question in my head, however the one from utils is used across the app. Do you think we can replace it everywhere?

src/renderer/features/extension-wallet/model/pairingForm.ts
aaa123eee: unfortunately the extension sends us `''` as an empty value which we need to handle

maybe I can map it during the extraction, otherwise I have to keep it as it is

src/renderer/features/extension-wallet/model/pairingForm.ts
aaa123eee: `as` is better for overriding types when `satisfies` is primary for validation and ensurance

src/renderer/entities/wallet/lib/account-utils.ts
tuul-wq: Yeah, I would have definitely done that

PR #3576, opened 2025-04-15
feat: new resources transport

johnthecat:


src/renderer/domains/collectives/_lib/helpers.ts
```diff
@@ -32,3 +33,33 @@ export const combineStores = <const T extends Record<string, CollectivesStruct<u
...
+    for (const [chainId, items] of entries(chainItems)) {
+      const prev = pickNestedValue(store, pallet, chainId);
+      const merged = merge({
+        a: prev ?? [],
+        b: items ?? [],
+        mergeBy,
+        filter: (a, b) => !isEqual(a, b),
+      });
+
+      // @ts-expect-error weird type error
```
aaa123eee, line 59: 😂

aaa123eee (approved):
Maybe we should move entities more careful to keep git history. Now I can see that many existing ones are considered as removed/created again

PR #3637, opened 2025-04-25
refactor: async history rendering

aaa123eee:


johnthecat (approved):

src/renderer/shared/ui-kit/AsyncItem/AsyncItem.tsx
```diff
@@ -4,6 +4,7 @@ import { type PropsWithChildren, useEffect, useState } from 'react';
 import { Box } from '../Box/Box';
 
 type AsyncItemProps = PropsWithChildren<{
+  sync?: boolean;
   delay?: number;
   onRender?: () => void;
   spaceToReserve?: {
```
johnthecat: plz use `fallback` prop instead with react node inside, it will be more reliable

PR #3646, opened 2025-04-28
fix: show tasks for watch-only wallets in fellowship profile

### LINKED ISSUE #3570: Fellowship. Add watch-only wallets into fellowship profile
Now there is empty state in fellowship tasks if fellowship member is added as watch-only account. We should show tasks, but remove any actions in cards.
Test account: `16SDAKg9N6kKAbhgDyxBXdHEwpwHUHs2CNEiLNGeZV55qHna`

savvyer:
## Description
Show tasks for watch-only wallets on fellowship page. Disable buttons.

## Related Issues
Closes #3570


tuul-wq (approved):

aaa123eee (approved):
Would be useful to add a title on disabled buttons with explanation why the action is inaccessible

PR #3648, opened 2025-04-29
Feat: PAPI fellowship

### LINKED ISSUE #3465: PAPI. Enable for Fellowship

tuul-wq:
## Description
Fellowship can read blockchain data using `polkadot-api` (Supported chains: Polkadot, Collectives, People)

## Related Issues
closes #3465 

## Changes Made
- Updated ChainsRegistry
- Fellowship pallets moved to `papi` regarding storage queries
- Updated all affected models


aaa123eee (dismissed):

src/renderer/shared/pallet/collective/storage.ts
```diff
@@ -1,141 +1,129 @@
...
+  //     }),
+  //   );
+  //
+  //   for await (const result of polkadotjsHelpers.createPagedRequest({
+  //     query: getQuery(type, papi).Members,
+  //     pageSize,
+  //   })) {
+  //     yield schema.parse(result);
+  //   }
+  // },
```
sokolova-an, line 110: do we need it at all?

sokolova-an (dismissed):

src/renderer/shared/pallet/collective/storage.ts
tuul-wq: Hard to say, @johnthecat implemented all pallet queries "just in case" even if it only contains placeholder implementation. This one cannot be moved to `papi` maintaining current approach with generators because `papi` has nothing to offer instead of `pagedEntries`.
If we don't care about some sort of "example" for future implementation that might never happen, then it could be removed 😌.

PR #3737, opened 2025-05-15
feat: add unified address subquery checks

### LINKED ISSUE #3675: Update chains config
- Update chains config and add a new field `legacyAddressPrefix`
If a chain doesn't have a new address format - `legacyAddressPrefix` and `addressPrefix` will be the same
- Align new changes with SubQuery

sokolova-an:
## Description
Make sure that the addresses we receive from subquery will have a correct `addressPrefix`
## Related Issues
Related #3720

### DIFF
--- a/src/renderer/shared/api/governance/off-chain/service/delegationService.ts
+++ b/src/renderer/shared/api/governance/off-chain/service/delegationService.ts
@@ -2,7 +2,7 @@ import { BN } from '@polkadot/util';
 import { GraphQLClient } from 'graphql-request';
 
 import { type Address, type Chain, ExternalType, type ReferendumId } from '@/shared/core';
-import { dictionary, nullable, toPrecision } from '@/shared/lib/utils';
+import { dictionary, nullable, toAccountId, toAddress, toPrecision } from '@/shared/lib/utils';
 import {
   type DelegateAccount,
   type DelegateDetails,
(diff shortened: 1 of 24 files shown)


src/renderer/shared/api/governance/off-chain/service/delegationService.ts
```diff
@@ -54,7 +54,7 @@ async function getDelegatesFromExternalSource(chain: Chain, blockNumber: number)
       return (
         (data as any)?.delegates?.nodes?.map(
           ({ accountId, delegators, delegatorVotes, delegateVotes, delegateVotesMonth }: any) => ({
-            accountId,
+            accountId: toAddress(toAccountId(accountId), { prefix: chain.addressPrefix }), // Address
```
johnthecat: why it's address now? it was account id and it's correct

src/renderer/shared/api/governance/off-chain/service/delegationService.ts
```diff
@@ -122,7 +130,7 @@ async function getDelegatesForAccount(chain: Chain, accountId: string): Promise<
       const result = (data as any)?.delegates?.nodes?.[0];
 
       return {
-        accountId: result.accountId,
+        accountId: toAddress(toAccountId(result.accountId), { prefix: chain.addressPrefix }),
```
johnthecat: same, it should be account id

johnthecat (changes requested):

src/renderer/shared/api/governance/off-chain/service/delegationService.ts
sokolova-an: it was not, for some reason, it's actually an address 
I don't know what was the reason 
image

src/renderer/shared/api/governance/off-chain/service/delegationService.ts
sokolova-an: same

johnthecat (approved):

PR #3742, opened 2025-05-15
fix: clean up multisigs after deleting signatories

### LINKED ISSUE #3727: Remove complex wallets, when connected wallet was removed
We need to remove multisig and proxy wallets, when user remove connected wallets.

Remove multisig wallets:
- When any wallet was removed from the app (including proxied wallets) multisigs should be checked and removed if there is no signatories available in the app any more
- When proxy was removed by proxy-worker (for example remove access to the wallet) connected multisig should be removed too

Remove proxied wallets:
- When user delete wallet, connected to the proxy wallet, proxy should be removed too

sokolova-an:
## Description
Delete multisigs and proxies wallets related to the deleted wallet
## Related Issues
Closes #3727

### DIFF
--- a/src/renderer/features/wallets/ForgetWallet/model/forget-wallet-model.ts
+++ b/src/renderer/features/wallets/ForgetWallet/model/forget-wallet-model.ts
@@ -173,7 +173,7 @@ sample({
 
 // TODO this connection is dirty, we should decouple wallet delete logic and proxy manipulation.
 sample({
-  clock: [walletModel.events.walletRemovedSuccess, walletModel.events.walletHiddenSuccess],
+  clock: [walletModel.events.walletRemovedSuccess, walletModel.events.walletHiddenSuccess], //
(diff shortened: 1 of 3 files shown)


src/renderer/features/wallets/ForgetWallet/model/forget-wallet-model.ts
```diff
@@ -148,7 +173,7 @@ sample({
 
 // TODO this connection is dirty, we should decouple wallet delete logic and proxy manipulation.
 sample({
-  clock: [walletModel.events.walletRemovedSuccess, walletModel.events.walletHiddenSuccess],
+  clock: [walletModel.events.walletRemovedSuccess, walletModel.events.walletHiddenSuccess], //
```
savvyer: typo?

savvyer (approved):
looks good

johnthecat (approved):

PR #3803, opened 2025-05-22
fix: order events by duration

### LINKED ISSUE #3786: Fellowship. All events. Oldest sorting does not work
## Description

When user try to sort all events on Fellowship by "Oldest" nothing happened

## Actual Behavior

https://github.com/user-attachments/assets/c93be33b-f67c-4bea-b4fa-05fff0d8c318

## Environment
DEV

aaa123eee:
## Description

Fix sort by duration

## Related Issues

closes #3786

## Changes Made

- keep fields extraction on higher level to reduce recalc & for sorting purposes

## Screenshots/Recordings

image

### DIFF
--- a/src/renderer/features/fellowship-activity-feed/components/ActivityList.tsx
+++ b/src/renderer/features/fellowship-activity-feed/components/ActivityList.tsx
@@ -9,13 +9,12 @@ import { activityFeed } from '../model/list';
 import { ActivityListView } from './ActivityListView';
 import { getDescription } from './utils';
 
-const now = Date.now();
-
 export const ActivityList = memo(() => {
   const { t } = useI18n();
 
   const feed = useUnit(activityFeed.$activityFeed);
   const identities = useUnit(identityModel.$list);
+  const now = Date.now();
 
   const records = useMemo(
     () =>
--- a/src/renderer/features/fellowship-activity-feed/components/ActivityList.tsx
+++ b/src/renderer/features/fellowship-activity-feed/components/ActivityList.tsx
@@ -1,16 +1,39 @@
 import { useUnit } from 'effector-react';
-import { memo } from 'react';
+import { memo, useMemo } from 'react';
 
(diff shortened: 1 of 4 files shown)


src/renderer/features/fellowship-activity-feed/components/ActivityList.tsx
```diff
@@ -1,16 +1,39 @@
...
 
+import { useI18n } from '@/shared/i18n';
+import { identityService } from '@/domains/network';
+import { identityModel } from '../model/identity';
 import { activityFeed } from '../model/list';
 
 import { ActivityListView } from './ActivityListView';
+import { getDescription } from './utils';
+
+const now = Date.now();
```
savvyer: Are you sure about this?

Date will be evaluated once at module load time. It won’t reflect the current time on re-renders or user interactions.

`now` could have Date from yesterday or week ago when the user first time opened the app

src/renderer/features/fellowship-activity-feed/components/ActivityModal.tsx
```diff
@@ -23,6 +23,8 @@ const orderVariants: Record<OrderKey, { field: string; direction: 'asc' | 'desc'
   'name-desc': { field: 'name', direction: 'desc' },
 };
 
+const now = Date.now();
```
savvyer: Same question here

src/renderer/features/fellowship-activity-feed/components/ActivityList.tsx
aaa123eee: copy pasted from another component

I put it to the component scope

aaa123eee: I got it now

We usually use the Countdown component so I relied on that, then it would be not necessary because of countdown logic, but here is just Duration. TY!

johnthecat (approved):

sokolova-an (approved):

PR #3865, opened 2025-06-02
feat: CAS remove votes modal

### LINKED ISSUE #3701: Migrate RemoveVotesModal
`src/renderer/widgets/RemoveVotesModal`

sokolova-an:
## Description
Update the modal for remove votes and for confirm screen
## Related Issues
Closes #3701


src/renderer/shared/transactions/createTransactionConfirmStore.ts
```diff
@@ -16,7 +16,7 @@ import { walletUtils } from '@/entities/wallet';
 export type TxConfirmInfo = {
   id?: number;
   initiator: AnyAccount;
-  signatory: AnyAccount;
+  signatory: AnyAccount | null;
```
cuteWarmFrog: Why there could not be a signatory?

johnthecat: That's wrong behavior, signatory should always exist, but most of the time it's overlapping initiator

sokolova-an: In case if we pass data from shards, there will be no signatory for sure. So it's cleaner to pass null in this case

cuteWarmFrog (approved):
seems gucci

src/renderer/widgets/RemoveVotesModal/ui/RemoveVotesModal.tsx
```diff
@@ -38,21 +39,26 @@ type Props = {
...
+  const votesList = useUnit(removeVotesModel.$votesList);
 
   const [isModalOpen, closeModal] = useModalClose(!isStep(step, Step.NONE), onClose);
   const [isBasketModalOpen, closeBasketModal] = useModalClose(isStep(step, Step.BASKET), onClose);
 
+  // if there are multiple voters then show shards modal
+  if (votesList.length > 1 && !single) {
+    return <RemoveVotesShardsModal votes={votes} chain={chain} asset={asset} api={api} onClose={onClose} />;
+  }
+
```
savvyer: we should probably do lazy loading for the react components otherwise we end up loading two versions: default and shards everywhere.

maybe we can use some sort of wrapper?
```
const RemoveVotesModal = React.lazy(() => import('./RemoveVotesModal.tsx'));
const RemoveVotesShardsModal = React.lazy(() => import('./RemoveVotesShardsModal.tsx'));

export const RemoveVotesWrapper = memo(function ({ votes, ..rest }: Props) {
  return (
    <div>
      {votes > 1 ? (
        <Suspense fallback={null}>
          <RemoveVotesShardsModal />
        </Suspense>
      ) : (
        <Suspense fallback={null}>
          <RemoveVotesModal />
        </Suspense>
      )}
    </div>
  );
});
```

What do you think?

src/renderer/widgets/RemoveVotesModal/ui/RemoveVotesModal.tsx
```diff
@@ -38,198 +35,26 @@ type Props = {
...
-      chain={chain}
-      title={t('governance.voting.selectAccountTitle')}
-      options={accounts.map((account) => ({ account }))}
-      onSelect={(a) => {
-        onSelect(a);
-        setIsSelectAccountOpen(false);
-      }}
-      onToggle={handleSelectAccountClose}
-    />
+    <div>
```
savvyer: maybe use react fragment <> instead?

src/renderer/widgets/RemoveVotesModal/index.ts
```diff
@@ -1,2 +1,3 @@
 export { RemoveVotesModal } from './ui/RemoveVotesModal';
-export { removeVotesModalAggregate } from './aggregates/removeVotesModal';
+export { RemoveVotesShardsModal } from './shards/components/RemoveVotesShardsModal';
```
johnthecat: move RemoveVotesModal to `default` directory please

johnthecat (approved):

Issue #3840, opened 2025-05-29
Unending loading state for balances when update version


stepanLav:
## Description

After update 1.16.0 balances run into unending loading state, fixing by page reload. The same happened with 1.16.1

![Image](https://github.com/user-attachments/assets/31975cbc-6a87-4bda-af8b-4d72a1db7b3b)

stepanLav:
Can't reproduced, looks like that fixed

PR #3868, opened 2025-06-02
fix: balances loading

savvyer:
## Description
Possibly fixes race condition when loading balances on portfolio page

## Related Issues
https://github.com/novasamatech/nova-spektr/issues/3840

## Changes Made
- Rewrite a sample without `once` operator
- Fix prettier warning


src/renderer/features/assets-balances/subscription/model/balance-sub-model.ts
```diff
@@ -134,17 +134,13 @@ sample({
...
-  chains: networkModel.$chains,
-});
-
 sample({
-  clock: once($walletAndChains.map(({ wallet, chains }) =>
-    nonNullable(wallet) && Object.keys(chains).length > 0
-  )),
-  source: $walletAndChains,
-  fn: ({wallet, chains}) => ({
+  source: {
```
johnthecat, line 138: should it be `clock`?

savvyer: Do you think 2 combines will be better instead of 1 clock?

PR #3871, opened 2025-06-03
fix: handle no activity and loading separately

### LINKED ISSUE #3818: Fellowship. Activity \ Agreement state for no data
## Description

There is no state for no data on Activity \ Agrteement

Image

aaa123eee:
## Description

Explicitly handle loading state and N/A

## Related Issues

closes #3818

## Changes Made

- Add dedicated loading state
- Do not fallback to empty arrays/records if data is loading/null/etc

### DIFF
--- a/src/renderer/domains/collectives/referendumMeta/service.ts
+++ b/src/renderer/domains/collectives/referendumMeta/service.ts
@@ -23,7 +23,7 @@ function getReferendumsSinceLastProof(referendums: ReferendumMeta[], member: Cor
 }
 
 function getActivityInfo(referendums: ReferendumMeta[], member: Member, maxRank: number, votes: Vote[]) {
-  if (referendums.length === 0) return null;
+  if (referendums.length === 0) return { activity: null, agreement: null };
 
   const possibleReferendums = referendums.filter(r =>
     trackService.rankSatisfiesVotingThreshold(member.rank, maxRank, r.track),
--- a/src/renderer/domains/collectives/referendumMeta/service.ts
+++ b/src/renderer/domains/collectives/referendumMeta/service.ts
@@ -22,7 +22,14 @@ function getReferendumsSinceLastProof(referendums: ReferendumMeta[], member: Cor
   return referendums.filter(r => r.created >= member.lastProof);
 }
 
-function getActivityInfo(referendums: ReferendumMeta[], member: Member, maxRank: number, votes: Vote[]) {
+function getActivityInfo(
+  referendums: ReferendumMeta[] | undefined,
+  member: Member,
+  maxRank: number,
+  votes: Vote[] | undefined,
+) {
+  if (!referendums) return { activity: null, agreement: null, isLoading: true };
+
   if (referendums.length === 0) return { activity: null, agreement: null, isLoading: false };
 
   const possibleReferendums = referendums.filter(r =>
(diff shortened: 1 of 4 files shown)


src/renderer/domains/collectives/referendumMeta/service.ts
```diff
@@ -22,15 +22,22 @@ function getReferendumsSinceLastProof(referendums: ReferendumMeta[], member: Cor
...
+
+  if (referendums.length === 0) return { activity: null, agreement: null, isLoading: false };
 
   const possibleReferendums = referendums.filter(r =>
     trackService.rankSatisfiesVotingThreshold(member.rank, maxRank, r.track),
   );
 
   if (referendums.length > 0 && possibleReferendums.length === 0) {
-    return { activity: 100, agreement: 100 };
+    return { activity: 100, agreement: 100, isLoading: false };
```
johnthecat: It will stuck in loading state if member has rank lower than threshold on any given referendum. I think loading state should be derived from actual `referendumMeta.pending` + `voting.pending` states

src/renderer/features/fellowship-profile/model/profile.ts
```diff
@@ -14,8 +14,8 @@ import { fellowship } from './fellowship';
 
 const requestIdentityFx = attach({ effect: identity.request });
 
-const $tracks = fellowship.$store.map(store => store?.tracks ?? []);
-const $referendumMeta = fellowship.$store.map(store => store?.referendumMeta ?? {});
+const $tracks = fellowship.$store.map(store => store?.tracks);
+const $referendumMeta = fellowship.$store.map(store => store?.referendumMeta);
 const $votes = fellowship.$store.map(store => store?.voting ?? []);
```
aaa123eee, line 19: this is the only annoying item that causes `0%` during an intermediate state

but we can't remove this `?? []` because sometimes we get only partial data

the only case is to refactor state reducers so they record an excessive `[]` to the proper store[palletType][chainId]

cuteWarmFrog (approved):
noice!

src/renderer/domains/collectives/referendumMeta/service.ts
aaa123eee: updated

no loading can stuck now but unfinished data can sneak now

src/renderer/domains/collectives/referendumMeta/service.ts
```diff
@@ -22,8 +22,15 @@ function getReferendumsSinceLastProof(referendums: ReferendumMeta[], member: Cor
   return referendums.filter(r => r.created >= member.lastProof);
 }
 
-function getActivityInfo(referendums: ReferendumMeta[], member: Member, maxRank: number, votes: Vote[]) {
-  if (referendums.length === 0) return null;
+function getActivityInfo(
+  referendums: ReferendumMeta[] | undefined,
```
johnthecat: I think it's better to use null everywhere instead of undefined

src/renderer/features/fellowship-profile/model/profile.ts
```diff
@@ -14,8 +14,8 @@ import { fellowship } from './fellowship';
 
 const requestIdentityFx = attach({ effect: identity.request });
 
-const $tracks = fellowship.$store.map(store => store?.tracks ?? []);
-const $referendumMeta = fellowship.$store.map(store => store?.referendumMeta ?? {});
+const $tracks = fellowship.$store.map(store => store?.tracks);
```
johnthecat:
```suggestion
const $tracks = fellowship.$store.map(store => store?.tracks ?? null);
```
and so on

aaa123eee: kk!

PR #3901, opened 2025-06-10
feat: multisig indexer integration

### LINKED ISSUE #3112: Multisig operations. Use indexer for getting multisig operations

johnthecat:
## Description
Integrated subquery indexer source to multisig operations.
Moved multisig operations model into network domain.

Closes #3112
Снимок экрана 2025-06-10 в 13 03 38


src/renderer/domains/network/multisig-operation/model.ts
```diff
@@ -0,0 +1,174 @@
...
+  }),
+});
+
+export const multisigOperation = {
+  $list,
+
+  populate: populateFx,
+  addOperations: addOperationsFx,
+  updateOperations: updateOperationsFx,
+  removeOperationsForAccount,
```
savvyer: This function stands out.
Should it also follow the naming convention with `Fx` in the end?

src/renderer/domains/network/multisig-operation/resource.ts
```diff
@@ -0,0 +1,302 @@
...
+      fn();
+      const interval = setInterval(fn, 60 * 1000);
+      unsubscribeFns.push(() => clearInterval(interval));
+    }
+
+    return Promise.all(unsubscribeFns).then(fns => () => {
+      for (const fn of fns) {
+        fn();
+      }
+    });
```
savvyer: Is there a reason why you use Promise.all here?

Looks like `unsubscribeFns` contains only synchronous functions only to clear interval.

src/renderer/domains/network/multisig-operation/service.ts
```diff
@@ -0,0 +1,101 @@
...
+       * example, consider the scenario with three addresses, A, B, and C,
+       * starting with 5FUGT, 5HMfS, and 5GhKJ. If we build the ABC multisig
+       * with the accounts in that specific order (i.e. first A, then B, and C),
+       * the real order of the accounts in the multisig will be ACB. If, in the
+       * Extrinsic tab, we initiate a multisig call with C, the order of the
+       * other signatories will be first A, then B. If we put first B, then A,
+       * the transaction will fail.
+       */
+      .sort((a, b) => a.localeCompare(b))
+      .filter(account => account !== signer)
```
savvyer: Maybe better to swap the order here?
Filter first and sort on smaller filtered array

savvyer (approved):
awesome 🤌

just sharing couple of things that caught my attention

Issue #4026, opened 2025-06-25
CAS. Governance. Delegate transaction is built wrong


stepanLav:
## Description

The same problem as in #4021 

https://github.com/user-attachments/assets/743cab40-5152-4c22-9a60-6820c8c0a056

stepanLav:
Status verification: ✅

PR #4035, opened 2025-06-26
fix: cas staking wrong operation

### LINKED ISSUE #4023: CAS. Staking. 2 signers with different sources won't handeled on confirm screen
## Description

When user has extension and Nova Wallet signatory - it always shows extension on confirm

https://github.com/user-attachments/assets/38785e94-3e1a-4502-ad7e-bdbee6c29b3d

## Test data

[spektr-database (28).json](https://github.com/user-attachments/files/20909631/spektr-database.28.json)

## Steps to Reproduce

1. Import database
2. Go to staking try to do bond more

## Expected Behavior

On confirm shows the same wallet type as selected in signatory

## Actual Behavior

Always shows kinda "first"

## Environment

https://github.com/novasamatech/nova-spektr/pull/3815/commits/16e305759a198292a24fed59f80a729ad31f257e

cuteWarmFrog:
## Description
1. fixing callData field name (wrapping worked wrong)
2. Hiding signatories if < 2
3. Finding shortest route from all possible

## Related Issues
Closes #4023 #4026


src/renderer/widgets/Staking/Unstake/default/ui/UnstakeForm.tsx
```diff
@@ -63,7 +64,7 @@ const ProxyFeeAlert = () => {
       fee={formattedFee}
       balance={formattedBalance}
       symbol={network.asset.symbol}
-      onClose={() => {}}
+      onClose={noop}
```
savvyer, line 67: Since you are cleaning up the mess here.
Can you please go through all <ProxyWalletAlert components and make sure we use `noop` everywhere. 
i saw cases where we do `onClose={initiator.reset}`, for example in DelegateFor, PayeeForm

src/renderer/domains/network/account/service.ts
```diff
@@ -200,27 +200,45 @@ function findRoute(source: AnyAccount, destination: AnyAccount, accounts: AnyAcc
     return [destination];
   }
 
-  const stack: AnyAccount[] = [];
   const graphs = createAccountGraphs(accounts, chain);
   const entryNode = graphs.get(source);
 
   if (nullable(entryNode)) {
     return [];
   }
```
savvyer, line 208: When this would ever happen?

if we can find a route between the initiator and signatory, wont it make sense to throw an error?
@johnthecat

src/renderer/domains/network/account/service.ts
```diff
@@ -200,27 +200,45 @@ function findRoute(source: AnyAccount, destination: AnyAccount, accounts: AnyAcc
...
+    [entryNode.account.id]: true, // because includes in array works for 0(N)
+  };
+  let i = 0; // because unshift works for 0(N)
+
+  while (queue.length > 0) {
+    const item = queue[i];
+
+    if (!item) {
+      return [];
+    }
```
savvyer, line 221: Check looks redundant. We enter the loop only if queue has some contents

src/renderer/widgets/Staking/Unstake/default/ui/UnstakeForm.tsx
cuteWarmFrog: will clear in other pr

PR #4113, opened 2025-07-07
fix: correct multisig operation creation after a successful multisig transaction

### LINKED ISSUE #3988: Multisig. Create an operation won't add any operation to the page
## Description

When user creates multisig operation it should add that operation locally

## Steps to Reproduce

1. Select Multisig wallet
2. Create any operation
3. Go to Operations page

## Expected Behavior

Created opeartion should be saved locally

## Actual Behavior

IT will get from the subquery

## Environment
DEV

johnthecat:
Closes #3988


cuteWarmFrog (approved):
Class

PR #4121, opened 2025-07-08
feat: no signatory alert + refactor

### LINKED ISSUE #3940: Multisig. Validation on existance address into signatory wallet for particular chain
Since multisig wallet now is multi-chain we run into a problem, when multisig can have tokens\operations on network for which was not created chain account in signatory wallet. To resolve that case we decided to introduce new validation:

https://www.figma.com/design/ZTrGE0r8cD1X7oBZcUzMRu/Flexible-multisig-2.0?node-id=230-53960&m=dev

Image

Signatory picker should be read-only and can't be interacted (if there is no other option) - https://www.figma.com/design/ZTrGE0r8cD1X7oBZcUzMRu?node-id=230-53960&m=dev#1326614050

cuteWarmFrog:
## Description
Added no signatory alert + cleaned a lot of little junk

## Related Issues
Closes #3940


src/renderer/widgets/UnlockModal/shards/ui/UnlockForm.tsx
```diff
@@ -138,25 +139,47 @@ const Signatories = () => {
...
+    if (!network) {
+      return [];
+    }
+
+    return signatories[0].map((signatory) => {
+      const balance = balanceUtils.getBalance(
+        balances,
+        signatory.accountId,
+        network.chain.chainId,
+        network.chain.assets[0].assetId.toString(),
```
savvyer: Let's use `getNativeAsset(chain.assets)` function to get the asset instead of accessing it by the index?!
Usually the native asset is the first one in the chains config, but better be prepared if the order changes.

Lets fix it everywhere in cas?

savvyer (changes requested):

src/renderer/shared/ui-entities/SignatorySelect/SignatorySelect.tsx
```diff
@@ -0,0 +1,147 @@
...
+import { Alert, FootnoteText, InputHint } from '@/shared/ui';
+import { Field, Select } from '@/shared/ui-kit';
+import { type AnyAccount, accountService } from '@/domains/network';
+import { WalletIcon, accountUtils } from '@/entities/wallet';
+import { Address } from '../Address/Address';
+import { AssetBalance } from '../AssetBalance/AssetBalance';
+
+type Props = {
+  signatory: AnyAccount | null;
+  signatories: { signer: AnyAccount; balance: string }[];
```
johnthecat: I don't like this structure, maybe balances should be passed as independent prop and mapped inside component?

src/renderer/features/proxy-add-pure/ui/AddPureProxiedForm.tsx
```diff
@@ -141,24 +142,41 @@ const Signatories = () => {
...
+    return signatories.map((signatory) => {
+      const balance = balanceUtils.getBalance(
+        balances,
+        signatory.accountId,
+        chain.value.chainId,
+        chain.value.assets[0].assetId.toString(),
+      );
+      return { account: signatory, balance: withdrawableAmount(balance) };
+    });
+  }, [signatories, balances]);
```
sokolova-an, line 164: we can also leave it in model combining signatories and balances - that makes ui component simpler, here and in the other form
what do you think?

src/renderer/widgets/DelegateModal/shards/ui/DelegateForm.tsx
```diff
@@ -97,25 +98,46 @@ const Signatories = () => {
...
+      signatories={signatoriesWithBalance}
+      allAccounts={allAccounts}
+      initiator={shards.value[0]}
+      allWallets={allWallets}
       hasError={signatory.hasError()}
       errorText={t(signatory.errorText())}
+      network={network}
       onChange={signatory.onChange}
     />
   );
```
sokolova-an: is there a point to have `SignatorySelect` here? if inside `SignatorySelect` we will return `null` since we map `signatories[0]`?
same about the other shards form

src/renderer/widgets/EditDelegationModal/default/ui/DelegateForm.tsx
```diff
@@ -68,42 +69,45 @@ const Signatories = () => {
...
       return [];
     }
-
     return signatories.map((signatory) => {
       const balance = balanceUtils.getBalance(
         balances,
         signatory.accountId,
         network.chain.chainId,
-        network.asset.assetId.toString(),
+        network.chain.assets[0].assetId.toString(),
```
sokolova-an: why don't use `network.asset` here and in there other forms?

src/renderer/features/proxy-add-pure/ui/AddPureProxiedForm.tsx
cuteWarmFrog: idk why but @johnthecat wanted to map them here

johnthecat: `signatoriesWithBalance` array is a struct made only for rendering, I believe it's better to keep it inside component. Model should be as clear as possible

johnthecat (approved):

src/renderer/widgets/DelegateModal/shards/ui/DelegateForm.tsx
cuteWarmFrog: will delete for shards in future pr

src/renderer/features/proxy-remove/ui/RemoveProxyForm.tsx
```diff
@@ -41,23 +45,46 @@ const Signatories = () => {
...
+    if (!signatories || !chain) {
+      return [];
+    }
+
+    return signatories.map((signatory) => {
+      const balance = balanceUtils.getBalance(
+        balances,
+        signatory.accountId,
+        chain.chainId,
+        chain.assets[0].assetId.toString(),
```
savvyer: There are still places with chain.assets[0] that can be replaced with getNativeAsset function

savvyer (approved):

PR #4175, opened 2025-07-16
fix: use local config even if fetch request fails

### LINKED ISSUE #4116: Update config management approach.
We want to get chains and tokens configs from the GitHub:
https://github.com/novasamatech/nova-spektr-utils/tree/main/chains/v1
https://github.com/novasamatech/nova-spektr-utils/tree/main/tokens/v1

savvyer:
## Description
If request to fetch remote chains/tokens configs fails we should not crash the app and load the available config from the local storage.

## Related Issues
Closes https://github.com/novasamatech/nova-spektr/issues/4116

### DIFF
--- a/src/renderer/entities/network/model/network-model.ts
+++ b/src/renderer/entities/network/model/network-model.ts
@@ -35,12 +35,6 @@ const failed = createEvent<ChainId>();
 
 const $chains = createStore<Record<ChainId, Chain>>({});
 
-persist({
-  key: 'chains_map',
-  store: $chains,
-  sync: true,
-});
-
 const $providers = createStore<Record<ChainId, ProviderWithMetadata>>({});
 const $apis = createStore<Record<ChainId, ApiPromise>>({});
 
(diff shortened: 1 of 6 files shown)


aaa123eee (approved):

src/renderer/entities/network/model/network-model.ts
```diff
@@ -69,6 +64,11 @@ const getDefaultStatusesFx = createEffect((chains: Record<ChainId, Chain>): Reco
   return dictionary(Object.values(chains), 'chainId', () => ConnectionStatus.DISCONNECTED);
 });
 
+persist({
+  key: 'chains_map',
+  store: $chains,
```
johnthecat, line 69: why did you removed sync flag? it indicates that lib should load local storage value on startup, i didn't found any related logic in new code

savvyer: `sync: true` is the default state of the flag in the library.

But I pushed the update that sets this flag explicitly as it was before.

johnthecat (approved):

PR #4180, opened 2025-07-17
feat(transfer): add fee tx

aaa123eee:
## Description

Use fake transaction for a fee estimation until a real transaction estimation possible

## Changes Made

- add optional feeTx for complexTxStore
- create feeTx for default transfer


src/renderer/shared/transactions/createComplexTxStore.ts
```diff
@@ -65,6 +71,19 @@ export const createComplexTxStore = <T extends Transaction>({
     target: wrapTransactionFx,
   });
 
+  if (feeTransaction) {
+    const wrapFeeTransaction = sample({
+      clock: [feeTransaction, api, $route],
+      source: { transaction: feeTransaction, api, route: $route },
```
savvyer, line 77: I think clock is redundant in this case.
If you remove clock, sample will rerun on every source change.

src/renderer/shared/transactions/createComplexTxStore.ts
```diff
@@ -65,6 +71,19 @@ export const createComplexTxStore = <T extends Transaction>({
     target: wrapTransactionFx,
   });
 
+  if (feeTransaction) {
```
savvyer, line 74: I'm not sure if it will work when fakeTx is null initially (for example not connected to the network), and once re-connected fakeTx has store value, but this code wont be executed

src/renderer/shared/transactions/createFeeCalculator.ts
```diff
@@ -32,7 +32,7 @@ export const createFeeCalculator = ({ $active = createStore(true), $transaction,
     return transactionService.getTransactionFee(transaction, api, signerOptions).then((x) => new BN(x));
   });
 
-  const $pending = restore(fetchFeeFx.pending.updates, true);
+  const $pending = restore(fetchFeeFx.pending.updates, false);
```
savvyer: Why do change pending state here?
It was a fix for a specific issue https://github.com/novasamatech/nova-spektr/pull/4037

src/renderer/shared/transactions/createComplexTxStore.ts
savvyer: I think clock is redundant in this case.
If you remove clock, sample will rerun on every source change.

savvyer (approved):

src/renderer/shared/transactions/createComplexTxStore.ts
```diff
@@ -22,6 +23,7 @@ export const createComplexTxStore = <T extends Transaction>({
   api,
   chain,
   transaction,
+  feeTx: feeTransaction,
```
johnthecat: 2 options: rename feeTx to feeTransaction or rename transaction to tx, but it leads to some refactoring 😇

src/renderer/shared/transactions/createComplexTxStore.ts
```diff
@@ -65,6 +71,20 @@ export const createComplexTxStore = <T extends Transaction>({
     target: wrapTransactionFx,
   });
 
+  if (feeTransaction) {
```
johnthecat, line 74: you got 2 `if (feeTransaction )`, let's collapse them

src/renderer/shared/transactions/createComplexTxStore.ts
```diff
@@ -65,6 +71,20 @@ export const createComplexTxStore = <T extends Transaction>({
...
+    const wrapFeeTransaction = sample({
+      source: { transaction: feeTransaction, api, route: $route },
+      filter: ({ transaction, api, route }) => nonNullable(transaction) && nonNullable(api) && nonNullable(route),
+      fn: ({ transaction, api, route }) => ({ transaction: transaction!, api: api!, route: route! }),
+    });
```
johnthecat, line 79: this should work
```suggestion
    const wrapFeeTransaction = sample({
      source: { transaction: feeTransaction, api, route: $route },
      filter: nonNullableMap,
    });
```

src/renderer/shared/transactions/createComplexTxStore.ts
```diff
@@ -84,15 +104,41 @@ export const createComplexTxStore = <T extends Transaction>({
...
+
+    sample({
+      clock: wrapFeeTransactionFx.doneData,
+      target: $feeTx,
+    });
+  }
+
+  const $mergedTx = combine({ tx: $tx, feeTx: $feeTx }, ({ tx, feeTx }) => {
+    return tx || feeTx;
+  });
```
johnthecat:
```suggestion
  const $mergedTx = combine($tx, $feeTx, (tx, feeTx) => tx || feeTx);
```

johnthecat (approved):

src/renderer/shared/transactions/createComplexTxStore.ts
aaa123eee: effector doesn't make a proper type inference in this case, I tried :-(

PR #4342, opened 2025-08-10
fix: vote history identity

### LINKED ISSUE #3837: Governance. Vote history. There is no identity data for voters
## Description

When user opens referenda details there is no identities for voters.

https://github.com/user-attachments/assets/87918e9e-f432-4ae1-ab8a-210326bd031c

savvyer:
## Description
Show identity sub name if present. Fix identity rendering in governance voting history.

## Related Issues
Closes https://github.com/novasamatech/nova-spektr/issues/3837

### DIFF
--- a/src/renderer/features/governance/aggregates/voteHistory.ts
+++ b/src/renderer/features/governance/aggregates/voteHistory.ts
@@ -4,7 +4,7 @@ import { createGate } from 'effector-react';
 import { type ChainId, type Referendum, type ReferendumId } from '@/shared/core';
 import { nonNullable } from '@/shared/lib/utils';
 import { type AccountId } from '@/shared/polkadotjs-schemas';
-import { identity } from '@/domains/network';
+import { identity, identityService } from '@/domains/network';
 import { voteHistoryModel } from '@/entities/governance';
 import { votingListService } from '../lib/votingListService';
 import { networkSelectorModel } from '../model/networkSelector';
@@ -43,9 +43,11 @@ const $voteHistory = combine(
       const aggregatedHistory = historyList.flatMap((vote) => {
         const splitVotes = votingListService.getDecoupledVotesFromVotingHistory(vote);
 
+        const identityName = identityService.getFullName(identities[chainId]?.[vote.voter as AccountId] ?? {});
+
         return splitVotes.map((vote) => ({
           ...vote,
-          name: identities[chainId]?.[vote.voter as AccountId]?.name ?? null,
+          name: identityName,
         }));
       });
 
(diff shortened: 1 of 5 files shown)


src/renderer/features/governance/aggregates/voteHistory.ts
```diff
@@ -45,7 +45,7 @@ const $voteHistory = combine(
 
         return splitVotes.map((vote) => ({
           ...vote,
-          name: identities[chainId]?.[vote.voter as AccountId]?.name ?? null,
+          name: identityService.getFullName(identities[chainId]?.[vote.voter as AccountId] ?? {}),
```
johnthecat: Should be a type error. Move `identities[chainId]?.[vote.voter]` to separated value and check it before passing to `identityService.getFullName`

johnthecat: Also, you can move `identities[chainId]` to separated variable and declare it near result

savvyer: @johnthecat pushed an update

src/renderer/features/governance/aggregates/voteHistory.ts
```diff
@@ -38,15 +38,21 @@ const $voteHistory = combine(
...
     for (const [referendumId, historyList] of Object.entries(history)) {
       const aggregatedHistory = historyList.flatMap((vote) => {
         const splitVotes = votingListService.getDecoupledVotesFromVotingHistory(vote);
 
-        return splitVotes.map((vote) => ({
-          ...vote,
-          name: identities[chainId]?.[vote.voter as AccountId]?.name ?? null,
-        }));
+        return splitVotes.map((vote) => {
+          const identity = chainIdentities[vote.voter as AccountId];
```
johnthecat: why do you need to type cast in the first place? In `DecoupledVote` this field has `Address` type, maybe we got error exactly here? IMO base fix here will be to change vote.voter to to AccountId and fix structure mapping inside transport

savvyer: The initial problem was with react memoization, freshly fetched identities were not applied in the UI.
I fixed it already.

savvyer: Typecasting reveals the hidden problem with `DecoupledVote` in which voter property is inconsistent, its either Address or AccountId.

It doesn't affect identities fetching, its seems the fetch works both with AccountId and an Address.

savvyer: I have removed typecasting is the code segment above by updating `DecoupledVote` type and the related code.

johnthecat (approved):

PR #4348, opened 2025-08-11
fix: portfolio fiat precision

### LINKED ISSUE #4216: Portfolio. Fiat precision
## Description

Use that formatting rules:
- https://github.com/novasamatech/research-docs/blob/main/amount-formatter.md

Additionally, we would like to implement truncate if amount can't be fit on UI, in case like that:

Image

address to test:
1HGnvAkk9nbfZ58CzUJhjcrLdEDMkr5qNkqqYkyD5BF5v6Y

savvyer:
## Description
Truncate fiat balances below 1 with many zeros.

## Related Issues
Closes https://github.com/novasamatech/nova-spektr/issues/4216

## Changes Made
- Fix fiat amount formatter based on https://github.com/novasamatech/research-docs/blob/main/amount-formatter.md
- Truncate tiny balances with css ellipsis
- Replaced Shimmering with Skeleton component


johnthecat:
We skipped this part in task description, but it will be nice to have ui improvement as well.
In `Price` component you should truncate a long value with ellipsis (16 symbols should be enough) and add tooltip in this case.

johnthecat (approved):

PR #4397, opened 2025-08-15
fix: combobox interaction

### LINKED ISSUE #4128: Transfer. Receive field. Problem with data deletion
## Description

Double click into field select not field data, but whole field with title

Example:

https://github.com/user-attachments/assets/80b233aa-d701-450b-908f-6e287a4fab15

## Environment
latest dev

savvyer:
## Description
Fix Input double-click text selection in combobox

## Related Issues
Closes https://github.com/novasamatech/nova-spektr/issues/4128

## Changes Made
- Allow interactions outside the combobox dropdown (radix-ui/react-popover) to restore expected double-click behavior inside the input field.


src/renderer/shared/ui-kit/Input/Input.tsx
```diff
@@ -53,7 +53,7 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(
     const id = useId();
 
     return (
-      <label
+      <div
```
johnthecat: You reduced click area, now paddings on left and right are not interactable

savvyer: reverted this change

sokolova-an (approved):

PR #4401, opened 2025-08-17
refactor: address branded type

johnthecat:
Update `Address` type to branded to remove bunch of error when `AccountId` been set to fields which should be addresses.


johnthecat:
Sunday night, what possibly could went wrong? Merging

PR #4464, opened 2025-08-22
test: autotests for validations

MaliDo:
## Description
- Adds new ids for validation alerts.
- Adds validation checks for msig transfer cases.
- Fixes migration info modal issue.


tests/system/pages/_elements/TransferModalElements.ts
```diff
@@ -7,10 +7,15 @@ export class TransferModalElements {
...
   static recipientInputLocator = TEST_IDS.OPERATIONS.RECIPIENT_INPUT;
+  static myselfButton = TEST_IDS.OPERATIONS.MYSELF_BUTTON;
   static signatoryLocator = TEST_IDS.OPERATIONS.SIGNATORY_SELECTOR;
   static signatoryOptionLocator = TEST_IDS.OPERATIONS.SIGNATORY_SELECTOR_OPTION;
   static xcmSelectorLocator = TEST_IDS.OPERATIONS.XCM_SELECTOR;
   static networkOption = TEST_IDS.MULTISIG.NETWORK_OPTION;
+  static sendingAmountError = `${TEST_IDS.VALIDATIONS.BALANCE}-sending-amount`;
+  static networkFeeAmountError = `${TEST_IDS.VALIDATIONS.BALANCE}-fee`;
+  static crossChainFeeAmountError = `${TEST_IDS.VALIDATIONS.BALANCE}-cross-chain-fee`;
+  static deliveryFeeAmountError = `${TEST_IDS.VALIDATIONS.BALANCE}-delivery-fee`;
```
stepanLav: there is a reason to use composition instead of creating separate key?

MaliDo: The base tag `TEST_IDS.VALIDATIONS.BALANCE` is defined once and reused to locate the main validation area for the total top-up amount.
Custom suffixes like `-delivery-fee` identify specific checks within that section.
Using composition keeps the main tag centralized and easier to maintain.

tests/system/pages/loginPage/BaseLoginPage.ts
```diff
@@ -31,6 +31,12 @@ export class BaseLoginPage extends BasePage<LoginPageElements> {
     await step('Go to the onboarding page', async () => {
       await this.goto(this.pageElements.url);
       await this.page.getByText(this.pageElements.onboardingLabel).waitFor();
+      await step('Mark "read" flag for migration info modal', async () => {
+        await this.page.evaluate(() => {
+          localStorage.setItem('assethub_migration_modal_seen_kusama', 'true');
+          localStorage.setItem('assethub_migration_modal_seen_polkadot', 'true');
+        });
+      });
```
stepanLav: Lets do it via playwright, we can use [addInitScript](https://playwright.dev/docs/api/class-browsercontext#browser-context-add-init-script-option-script)

Image

In that case we will make sure that local storage will be filled, even if test won't use gotoOnboarding

tests/system/pages/modals/TransferModalWindow.ts
```diff
@@ -99,6 +116,86 @@ export class TransferModalWindow extends BaseModal<TransferModalElements> {
     });
   }
 
+  public async isSendingAmountValidationOnPage(): Promise<void> {
+    await step('Check sending amount validation on transfer modal', async () => {
+      const balanceError = this.page.getByTestId(TEST_IDS.VALIDATIONS.BALANCE);
+      const amountError = this.page.getByTestId(TransferModalElements.sendingAmountError);
+
+      await Promise.race([
+        balanceError.waitFor({ state: 'visible', timeout: 5000 }),
```
stepanLav: why do we use custom timeout everywhere?
basically, playwright uses timeouts globally

tests/system/pages/modals/TransferModalWindow.ts
```diff
@@ -99,6 +116,86 @@ export class TransferModalWindow extends BaseModal<TransferModalElements> {
...
+  public async isNetworkFeeValidationOnPage(): Promise<void> {
+    await step('Check network fee validation on transfer modal', async () => {
+      const balanceError = this.page.getByTestId(TEST_IDS.VALIDATIONS.BALANCE);
+      const feeError = this.page.getByTestId(TransferModalElements.networkFeeAmountError);
+
+      await Promise.race([
+        balanceError.waitFor({ state: 'visible', timeout: 5000 }),
+        feeError.waitFor({ state: 'visible', timeout: 5000 }),
+      ]).catch(async () => {
+        throw new Error('Network fee validation did not appear within 5s.');
```
stepanLav: we do the same many times, lets try to adhere [DRY](https://ru.wikipedia.org/wiki/Don%E2%80%99t_repeat_yourself)

tests/system/tests/transfers/nova.transfers.system.test.ts
```diff
@@ -8,8 +8,8 @@ import { transferTestCases } from '../../utils/transferTestCases';
 const feature = 'Wallets. Nova. Single wallet';
 const story = 'Transfers';
 
-test.describe('Regular nova transfers', { tag: ['@regular-nova-transfers', '@regress'] }, () => {
-  for (const { chainName, assetId, amount, recipient } of transferTestCases) {
+test.describe('Regular nova transfers', { tag: ['@regular-nova-transfers', '@regress', '@validations'] }, () => {
+  for (const { chainName, assetId, amount, validationAmount, recipient } of transferTestCases) {
```
stepanLav: in that case we will check validations multiple times on all transfer test cases. Lets move it to separate test, in order to split that type of tests. 
@regular-nova-transfers - should check that basically user can create transfer
@validations - should check basically validations

stepanLav: this is relevant for all tests in this PR

src/renderer/shared/ui-entities/TransactionValidationError/TransactionValidationError.tsx
```diff
@@ -137,6 +150,12 @@ const TransactionBalanceError = ({
 
   const imbalances: { asset: Asset; imbalance: BN }[] = [];
 
+  const toTestId = (s: string) =>
+    s
+      .toLowerCase()
+      .replace(/[^a-z0-9]+/g, '-')
+      .replace(/(^-|-$)/g, '');
+
```
stepanLav: we shouldn't implement it in ui-entities.
also, action may change in the future it will broke the assumption in concatenated test_ids. Lets move it on test logic, instead of creating it in the app.
@johnthecat do you have any suggestion?

tests/system/pages/_elements/TransferModalElements.ts
stepanLav: we can't do that, action may change, also it is not good solution due to https://github.com/novasamatech/nova-spektr/pull/4464/files#r2432431026

MaliDo: Implemented the approach without amending actions for the test-ids. As discussed, actions won't change, therefore can be used as custom test-id add-ons.

tests/system/pages/loginPage/BaseLoginPage.ts
MaliDo: Done

tests/system/pages/modals/TransferModalWindow.ts
MaliDo: Removed timeouts + updated waitForContinueButtonToBeEnabled() commands which were implemented using waitForTimeout(), therefore didn't cause the test to break until the whole test timeout is expired.

tests/system/pages/modals/TransferModalWindow.ts
MaliDo: Implemented another approach to be able to pass a specific validation as an argument to check for its presence.

tests/system/tests/transfers/nova.transfers.system.test.ts
MaliDo: Restructured tests. Made a separate transfer amount validation test without checking for validations in usual transfer tests.

src/renderer/shared/ui-entities/TransactionValidationError/TransactionValidationError.tsx
MaliDo: As explained above, redesigned the test-id approach without amending actions in the ui code base.

stepanLav:
1. We should not wait for validation messages.

This is the wrong approach because a user can click through the flow without triggering the validation. The tests should be designed to find these issues, not to wait for the validation messages to appear.

https://github.com/user-attachments/assets/202c68aa-718d-4cc9-9bc5-f74bcbeafa55

Expected result:

- An alert about missing permissions appears
- Network fees are calculated
- The "Continue" does not available

Test validation:
- If the "Continue" button is available, click it and fail the test
- If network fees are not calculated or show 0, fail the test
- If the alert does not appear after the default implicit wait period (once available balance and network fees are displayed), fail the test

Current tests shouldn't pass, the app works wrong

tests/system/pages/BasePage.ts
```diff
@@ -20,12 +20,20 @@ export abstract class BasePage<T extends BasePageElements = BasePageElements> {
...
+  // public async gotoMain(): Promise<this> {
+  //   await step(`Navigate to the page: ${this.pageElements.url}`, async () => {
+  //     await this.page.goto(this.pageElements.url);
+  //     await this.page.waitForLoadState('load');
+  //     await this.page.waitForLoadState('domcontentloaded');
+  //     await this.page.waitForLoadState('networkidle');
+  //   });
+
+  //   return this;
+  // }
```
stepanLav:
```suggestion

```

tests/system/pages/modals/TransferModalWindow.ts
```diff
@@ -130,4 +175,70 @@ export class TransferModalWindow extends BaseModal<TransferModalElements> {
...
     return this.previousPage;
   }
+
+  private validationLocators(v: Validation) {
+    const fatal = this.page.getByTestId(TEST_IDS.VALIDATIONS.FATAL);
+    const missingAccount = this.page.getByTestId(TEST_IDS.VALIDATIONS.MISSING_ACCOUNT);
+    const permission = this.page.getByTestId(TEST_IDS.VALIDATIONS.PERMISSION);
+    const balance = this.page.getByTestId(TEST_IDS.VALIDATIONS.BALANCE);
+
+    switch (v) {
```
stepanLav: the same switch case in DelegateAuthorityModal, lets move it to utils?

tests/system/pages/BasePage.ts
MaliDo: Done

tests/system/pages/modals/TransferModalWindow.ts
MaliDo: Moved the switch function to a new validationHelpers in utils

stepanLav (approved):

PR #4509, opened 2025-09-02
fix: handle decimal values without leading zeros in formatAmount util

### LINKED ISSUE #3951: Decimal number without leading zero is shown as whole number on transaction confirmation screen
## Description

When a user enters a decimal number without a leading zero (e.g., .1) into the transaction modal form and proceeds to the confirmation screen, the value is incorrectly displayed as a whole number (1). If the user clicks "Back" to return to the form and then proceeds forward again, the value is correctly shown as 0.1 on the confirmation screen.

## Steps to Reproduce

1. Open the transaction modal.
2. In the form, enter .1 as the amount (or any decimal without a leading zero).
3. Click Continue to proceed to the confirmation screen.
4. Observe the displayed value on the confirmation screen - it shows 1 instead of 0.1.
5. Click Back to return to the form.
6. Without changing anything, click Continue again.
7. Observe that the value is now correctly shown as 0.1.

## Expected Behavior

The confirmation screen should consistently display the value 0.1 when the user enters .1 in the form, without requiring navigation back and forth.

## Actual Behavior

The value .1 is incorrectly displayed as 1 on the first visit to the confirmation screen. Only after returning to the form and proceeding again the value is displayed correctly as 0.1.

https://github.com/user-attachments/assets/e040976d-2c44-4aee-ada7-70ee4fd12dae

kalininilya:
Resolves [#3951](https://github.com/novasamatech/nova-spektr/issues/3951)

## Description
Fixes a bug in the transaction confirmation modal where decimal numbers without leading zeros (eg .1 ) were incorrectly displayed as whole numbers on the transaction confirmation screen. The issue occurred because the formatAmount function's regex pattern only matched decimal values with at least one digit before the decimal point, causing values like .1 to fall through to fallback logic that stripped the decimal point.
## Changes Made

1. Updated regex pattern in formatAmount function from `/^(\d+)\.(\d+)$/` to `/^(\d*)\.(\d+)$/` to allow optional digits before decimal point
2. Added fallback `|| '0' `when extracting integer part to handle empty string case for values like .1
3. Added tests to formatAmount

### DIFF
--- a/src/renderer/shared/lib/utils/__tests__/balance.test.ts
+++ b/src/renderer/shared/lib/utils/__tests__/balance.test.ts
@@ -2,7 +2,7 @@ import { BN, BN_ZERO } from '@polkadot/util';
 
 import { type AssetBalance, type AssetId, type Balance, type BalanceId, LockTypes } from '@/shared/core';
 import { type AccountId } from '@/shared/polkadotjs-schemas';
-import { formatBalance, stakeableAmountBN, transferableAmountBN, withdrawableAmount } from '../balance';
+import { formatAmount, formatBalance, stakeableAmountBN, transferableAmountBN, withdrawableAmount } from '../balance';
 
 const createBalance = (params: {
   free?: string | number;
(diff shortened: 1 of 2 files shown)


aaa123eee (approved):

Yanaty (approved):

cuteWarmFrog (approved):
lgtm

PR #4539, opened 2025-09-05
feat: group votes by voter in voting history

### LINKED ISSUE #2293: Governance. Vote history. There is no delegated votes
## Description
When user see voted history - there is no delegated votes.

## Steps to Reproduce

1. Find referenda with delegated votes
2. Open votes history
3. Find account with delegated votes

## Expected Behavior
Shows delegated votes alongside with direct votes

## Actual Behavior
Sows only direct votes
Screenshot 2024-09-05 at 17 50 10

Screenshot 2024-09-05 at 17 53 31

## Environment
v1.10.0

Yanaty:
## Description
Governance. Vote history. There is no delegated votes

## Related Issues
Closes #2293

## Changes Made
Added Delegated by label

## Screenshots/Recordings
Before:
Снимок экрана 2025-09-05 в 13 07 23

After:
Снимок экрана 2025-09-08 в 19 49 58

### DIFF
--- a/src/renderer/features/governance/aggregates/voteHistory.ts
+++ b/src/renderer/features/governance/aggregates/voteHistory.ts
@@ -2,10 +2,12 @@ import { combine, createEvent, sample } from 'effector';
 import { createGate } from 'effector-react';
 
 import { type ChainId, type Referendum, type ReferendumId } from '@/shared/core';
-import { entries, nonNullable } from '@/shared/lib/utils';
+import { entries, nonNullable, toAccountId } from '@/shared/lib/utils';
 import { type AccountId } from '@/shared/polkadotjs-schemas';
 import { identity, identityService } from '@/domains/network';
+import { contactModel } from '@/entities/contact';
 import { voteHistoryModel } from '@/entities/governance';
+import { walletModel } from '@/entities/wallet';
 import { votingListService } from '../lib/votingListService';
 import { networkSelectorModel } from '../model/networkSelector';
 import { votingAssetModel } from '../model/votingAsset';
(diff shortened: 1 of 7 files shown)


stepanLav:
Generally looks good!
There are some cases, which we need to handle:

### 1. Use `<Account/>` entity instead of Address in nested list, in order to have access to block explorers

Screenshot 2025-09-08 at 21 50 24

something like that:
```js
<Account
  hideAddress
  iconSize={20}
  title={resolveAccountName(delegatedVote.delegator)}
  accountId={delegatedVote.delegator}
  chain={chain}
  variant="truncate"
/>
```

### 2. Title should be taken with identities and show current wallet name, when we see it from particular wallet.

Example:
```js
  const walletNameByAccountId = useMemo(() => {
    const map: Record<string, string> = {};
    for (const w of wallets) {
      for (const acc of w.accounts) {
        map[acc.accountId] = acc.name;
      }
    }
    return map;
  }, [wallets]);

  const contactNameByAccountId = useMemo(() => {
    const map: Record<string, string> = {};
    for (const c of contacts) {
      map[toAccountId(c.address)] = c.name;
    }
    return map;
  }, [contacts]);

  const resolveAccountName = useMemo(() => {
    return (accountId: string): string | undefined => {
      if (!chain) return undefined;
      const id = identities[chain.chainId]?.[accountId];
      if (id) return identityService.getFullName(id);
      if (walletNameByAccountId[accountId]) return walletNameByAccountId[accountId];
      if (contactNameByAccountId[accountId]) return contactNameByAccountId[accountId];
      return undefined;
    };
  }, [identities, chain, walletNameByAccountId, contactNameByAccountId]);
  ```

### 3. Nice to have. Voting power should be aligned for "plate" and for single line, lets keep empty space from the right size for all lines, in order to be able align it. Delegated votes should be keep without empty space, as it is right now
Screenshot 2025-09-08 at 21 56 49

### 4. Search state. Currently, if you try to find delegator account or it's identity, search results will show nothing. Expected result - shows exact delegator address

**Comment by Yanaty** [comment]
fixed all comments

File: src/renderer/features/governance/aggregates/voteHistory.ts [hunk 1]
```diff
@@ -33,29 +35,48 @@ const $voteHistory = combine(
...
+      }
+    }
+
+    const contactNameByAccountId: Record<string, string> = {};
+    for (const contact of contacts) {
+      contactNameByAccountId[toAccountId(contact.address)] = contact.name;
+    }
+
+    const resolveAccountName = (accountId: string): string | null => {
+      const accountIdentity = chainIdentities?.[accountId as AccountId];
```
cuteWarmFrog: assertion might be removed (change to AccountId in couple of places)

src/renderer/features/governance/aggregates/voteHistory.ts
```diff
@@ -2,15 +2,17 @@ import { combine, createEvent, sample } from 'effector';
...
-import { entries, nonNullable } from '@/shared/lib/utils';
+import { entries, nonNullable, toAccountId } from '@/shared/lib/utils';
 import { type AccountId } from '@/shared/polkadotjs-schemas';
 import { identity, identityService } from '@/domains/network';
+import { contactModel } from '@/entities/contact';
 import { voteHistoryModel } from '@/entities/governance';
+import { walletModel } from '@/entities/wallet';
 import { votingListService } from '../lib/votingListService';
 import { networkSelectorModel } from '../model/networkSelector';
 import { votingAssetModel } from '../model/votingAsset';
```
cuteWarmFrog: deprecated api, use new one please

src/renderer/features/governance/aggregates/voteHistory.ts
```diff
@@ -33,29 +35,48 @@ const $voteHistory = combine(
...
 
     const result: Record<ReferendumId, AggregatedVoteHistory[]> = {};
     const chainIdentities = identities[chainId];
 
+    const walletNameByAccountId: Record<string, string> = {};
+    for (const wallet of wallets) {
+      for (const account of wallet.accounts) {
+        walletNameByAccountId[account.accountId] = account.name;
+      }
+    }
```
cuteWarmFrog: I would extract into sepparate store to not traverse each time

also I guess it's possible to not use wallets here at all (in this case afaik you are using wallets only as a way to get accounts data). accounts.$list gives you all accounts in the app

src/renderer/features/governance/components/VotingHistory/VotingHistoryList.tsx
```diff
@@ -25,10 +25,53 @@ export const VotingHistoryList = memo(({ items, asset, listName, chain, loading
...
+          .map((dv) => {
+            const rawAddress = dv.delegator;
+            const formattedAddress = toAddress(dv.delegator, { prefix: chain?.addressPrefix });
+            return `${rawAddress} ${formattedAddress}`;
+          })
+          .join(' '),
+      }),
+    });
+
+    if (!query.trim()) {
```
cuteWarmFrog: can we return items before search if query is empty string?

cuteWarmFrog (approved):

src/renderer/features/governance/aggregates/voteHistory.ts
Yanaty: fixed

src/renderer/features/governance/aggregates/voteHistory.ts
Yanaty: fixed

src/renderer/features/governance/aggregates/voteHistory.ts
Yanaty: fixed

src/renderer/features/governance/components/VotingHistory/VotingHistoryList.tsx
Yanaty: done

cuteWarmFrog (approved):

PR #4540, opened 2025-09-05
refactor(account-sync): handle only succeeded chains

### LINKED ISSUE #3586: Remove wallet. Remove proxy and multisigs if there is no connected wallets any more
## Description

When user deletes their wallets from the app, there are still connected wallets which he can not use, like multisig

1. Add possibility to return error from sync service by chain, if we can't get data from the chain
2. Delete proxieds only if we get account list, if get an error - do nothing

aaa123eee:
## Description

Respect the chains participated in sync

## Related Issues

closes #3586

### DIFF
--- a/src/renderer/domains/network/account-sync/service.ts
+++ b/src/renderer/domains/network/account-sync/service.ts
@@ -21,16 +21,25 @@ type Params<Providers extends AccountProvider<any>[]> = {
   providers: Providers;
 };
 
+type SyncResult<Providers extends AccountProvider<any>[]> = {
+  accounts: InferProviderAccount<Providers[number]>[];
+  chains: ChainId[];
+};
+
 async function syncAccounts<const Providers extends AccountProvider<any>[]>({
   accounts,
   chains,
   providers,
-}: Params<Providers>): Promise<InferProviderAccount<Providers[number]>[]> {
+}: Params<Providers>): Promise<SyncResult<Providers>> {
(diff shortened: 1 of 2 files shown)


src/renderer/features/account-sync/model/sync.ts
```diff
@@ -92,17 +92,19 @@ sample({
   source: {
     allAccounts: accounts.$list,
     allWallets: walletModel.$allWallets,
-    chains: networkModel.$chains,
+    allChains: networkModel.$chains,
   },
-  fn({ allAccounts, allWallets, chains }, [result, identities]) {
-    const syncedProxyAccounts = result.filter(accountSyncService.isSyncedProxyAccount);
+  fn({ allAccounts, allWallets, allChains }, [syncResult, identities]) {
+    console.log({ syncResult, allChains, allChainsLength: Object.values(allChains).length });
```
Yanaty: Needed?

src/renderer/domains/network/account-sync/service.ts
```diff
@@ -74,5 +86,5 @@ export const accountSyncService = {
   isSyncedProxyAccount,
   isSyncedMultisigAccount,
 
-  syncAccounts,
+  syncAccounts: syncAccounts,
```
cuteWarmFrog, line 89: syncAccounts

cuteWarmFrog (approved):

PR #4566, opened 2025-09-10
Fix/multisig evm

### LINKED ISSUE #4557: Regular multisig creation flow for EVM accounts is working wrong
## Description

There is a problem with regular multisig and EVM account selection. Basically algorithm should work:
1. My account field is responsible for which network will be used in multisig creation flow
- If in My account is selected EVM address: 
  - network fee will be filtered by only evm based networks
  - as a signatories is available only evm based addresses
  - contacts and my accounts contains only evm based variants, if lenght == 0 shows nothing in a group, if there is no suggestion, do not show drop down at all
- If in My account is selected Substrate address:
  - network fee will be filtered by only substrate based networks
  - as a signatories is available only substrate based addresses
  - contacts and my accounts contains only substrate based variants, if lenght == 0 shows nothing in a group, if there is no suggestion, do not show drop down at all

2. Network fee selection is connected to available balances
- Try to pay network fee on Polkadot with DOT
  - if there is no DOT on polkadot, check next network
  - if there is no available balance on any available network in network selector - shows an error that there is no available balance to pay network fee in Polkadot for substrate based account, in mythos for EVM based accounts

cuteWarmFrog:
## Description
Checking all balances for all possible signatories and choosing chain according to filtered data

## Related Issues
Closes #4557

### DIFF
--- a/src/renderer/features/multisig-wallet-create/multisig-wallet/model/chain-model.ts
+++ b/src/renderer/features/multisig-wallet-create/multisig-wallet/model/chain-model.ts
@@ -81,7 +81,7 @@ sample({
   target: $chainsWithFee,
 });
 
-const $chainsWithFeeAvailableForInitiatorChain = combine(
+const $filteredChainsWithFee = combine(
   { chains: $chainsWithFee, initiator: flowModel.$initiator },
   ({ chains, initiator }) => {
     if (!initiator) return [];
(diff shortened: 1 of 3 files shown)


src/renderer/features/multisig-wallet-create/multisig-wallet/model/chain-model.ts
```diff
@@ -84,16 +81,27 @@ sample({
...
+const $chainsWithFeeAvailableForInitiatorChain = combine(
+  { chains: $chainsWithFee, initiator: flowModel.$initiator },
+  ({ chains, initiator }) => {
+    if (!initiator) return [];
+
+    return chains.filter(chain => accountService.isAccountAvailableOnChain(initiator, chain.chain));
+  },
+);
+
+const filteredChains = $chainsWithFeeAvailableForInitiatorChain.map(chains => chains.map(chain => chain.chain));
```
aaa123eee: first naming is very detailed and the second is very not 😅

```suggestion
const $multisigChains = combine(
  { chains: $chainsWithFee, initiator: flowModel.$initiator },
  ({ chains, initiator }) => {
    if (!initiator) return [];

    return chains.filter(chain => accountService.isAccountAvailableOnChain(initiator, chain.chain).map(chain => chain.chain));
  },
);
```

src/renderer/features/multisig-wallet-create/multisig-wallet/model/chain-model.ts
```diff
@@ -134,9 +144,25 @@ sample({
   }),
 });
 
+sample({
+  clock: $availableChains,
+  source: formModel.form.fields.chainId.$value,
```
aaa123eee, line 147: shouldn't we clock when chainId is changed?

cuteWarmFrog: we are trying to change changeId when there are new signatories available and their new balances change available networks

aaa123eee (approved):

PR #4585, opened 2025-09-11
feat: filter flex mustisig accounts by chain, show proxy balance instead of msig balance for flex msigs

### LINKED ISSUE #4522: Flex multisig portfolio view (token-centric). Show only tokens from the flex multisig’s chain, not from all chains
## Description

Currently, in the token-centric portfolio view, tokens of a flexible multisig are displayed across all chains, even though the flex multisig only exists on a single chain.

An improvement was already implemented for the chain-centric mode: only the chain where the flex multisig exists is displayed.

The same logic should be applied to token-centric mode - tokens of a flex multisig should be visible only on its actual chain, and other chains should be hidden.

kalininilya:
Resolves [#4522](https://github.com/novasamatech/nova-spektr/issues/4522)

## Description
Fixes the token-centric portfolio view for flexible multisig accounts by ensuring tokens are only displayed on the chain where the flex multisig actually exists. Additionally, addresses the balance display issue where the portfolio view was showing the multisig balance instead of proxy balance for flexible multisig accounts.

## Changes Made

- Created accountService.filterAccountsAvailableForChain() for flexible account chain filtering
- Fix token-centric view: Apply chain-specific filtering to ensure flex multisig tokens only appear on their actual chain
- Correct balance display: Ensure proxy balance is shown instead of multisig balance for flexible multisig accounts

## Screenshots/Recordings

Screenshot 2025-09-11 at 13 48 00

## Checklist

### DIFF
--- a/src/renderer/domains/network/account/service.ts
+++ b/src/renderer/domains/network/account/service.ts
@@ -20,6 +20,7 @@ import {
 } from '@/shared/ui-entities';
 import { balanceUtils } from '@/entities/balance';
 import { networkUtils } from '@/entities/network';
+import { accountUtils } from '@/entities/wallet';
 import { type AnyTransaction } from '../transaction/types';
 
 import {
@@ -109,6 +110,19 @@ function isAccountAvailableOnChain(account: AnyAccount, chain: Chain) {
 
 function filterAccountsOnChain(accounts: AnyAccount[], chain: Chain) {
   return accounts.filter(account => isAccountAvailableOnChain(account, chain));
+}
+
+function filterAccountsAvailableForChain(accounts: AnyAccount[], chain: Chain): AnyAccount[] {
+  return accounts.filter(account => {
+    if (!isAccountAvailableOnChain(account, chain)) return false;
+    if (accountUtils.isFlexibleProxiedAccount(account)) {
+      return account.chainId === chain.chainId;
+    }
+    if (accountUtils.isFlexibleMultisigAccount(account)) {
+      return account.chainId === chain.chainId;
+    }
+    return true;
+  });
 }
 
 function filterAccountsByWallet(accounts: AnyAccount[], walletId: number) {
@@ -414,6 +428,7 @@ export const accountService = {
   hasPermissionToMakeActions,
 
   filterAccountsOnChain,
+  filterAccountsAvailableForChain,
   filterAccountsByWallet,
 
   // graph
--- a/src/renderer/domains/network/account/service.ts
+++ b/src/renderer/domains/network/account/service.ts
@@ -20,7 +20,6 @@ import {
 } from '@/shared/ui-entities';
 import { balanceUtils } from '@/entities/balance';
 import { networkUtils } from '@/entities/network';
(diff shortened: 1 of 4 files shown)


src/renderer/domains/network/account/service.ts
```diff
@@ -111,6 +112,19 @@ function filterAccountsOnChain(accounts: AnyAccount[], chain: Chain) {
...
 
+function filterAccountsAvailableForChain(accounts: AnyAccount[], chain: Chain): AnyAccount[] {
+  return accounts.filter(account => {
+    if (!isAccountAvailableOnChain(account, chain)) return false;
+    if (accountUtils.isFlexibleProxiedAccount(account)) {
+      return account.chainId === chain.chainId;
+    }
+    if (accountUtils.isFlexibleMultisigAccount(account)) {
+      return account.chainId === chain.chainId;
+    }
```
valentunn: I believe this should be done via `accountAvailabilityOnChainAnyOf` - base service should not know about flexible multisigs

src/renderer/features/assets/AssetsPortfolioView/lib/tokensService.ts
```diff
@@ -72,11 +73,13 @@ function sumTokenBalances(a: PortfolioTokenBalance, b: PortfolioTokenBalance): P
...
-  return accounts.reduce<AccountId[]>((acc, account) => {
-    if (accountService.isAccountAvailableOnChain(account, chain)) {
-      acc.push(account.accountId);
-    }
+  const availableAccounts = accountService.filterAccountsAvailableForChain(accounts, chain);
 
+  return availableAccounts.reduce<AccountId[]>((acc, account) => {
+    if (accountUtils.isFlexibleMultisigAccount(account)) {
+      return acc;
+    }
```
valentunn, line 82: This looks wrong - why token feature is aware of account internals?

src/renderer/domains/network/account/service.ts
kalininilya: You are right, moved the condition to accountService.accountAvailabilityOnChainAnyOf, thank for noticing!

src/renderer/features/assets/AssetsPortfolioView/lib/tokensService.ts
kalininilya: Do you see another place where we could apply this? We need to exclude FlexibleMultisig accounts from portfolio balance calculations when FlexibleMultisigs are selected, and only include balances of their proxies

I was thinking about adding it in the account service using something like hasBalanceAnyOf, but that would couple accounts to balances, which doesn’t seem like a desirable decision

Curious to hear your thoughts!

johnthecat: As I said recently, we don't want to know about account types inside features not related to account management. Instead you can rely on account properties and relations. In case of portfolio for example we are talking about "facade" accounts, that can represent this wallet. This facade account are actually graph roots, they don't have parent nodes. There is a special method in account service for this
```
accountService.findInitiators(walletAccounts, chain)
```

PR #4636, opened 2025-09-16
refactor(flex): proper wrap order

### LINKED ISSUE #4626: Flex multisig. Create transfer - not a proxy error
## Description

Image

aaa123eee:
## Description

Change transaction wrapping order for Flex:
1. wrap w/ proxy
2. wrap w/ multisig

## Related Issues

closes #4626

## Checklist

### DIFF
--- a/src/renderer/features/multisig-wallet/index.tsx
+++ b/src/renderer/features/multisig-wallet/index.tsx
@@ -171,8 +171,6 @@ transactionSDK(multisigWalletFeature, {
 
       return transactionService.getExtrinsicWeight(extrinsic).then(maxWeight => {
         if (accountUtils.isFlexibleMultisigAccount(account)) {
-          const encodedTransaction = transactionService.encodeTransaction(transaction, api);
-
           const proxyTransaction: ProxyTransaction = {
             type: 'decoded',
             section: 'proxy',
@@ -229,11 +227,10 @@ transactionSDK(multisigWalletFeature, {
       assert(signatory, 'Signatory not found');
 
       const otherSignatories = multisigOperationService.getOtherSignatories(account, signatory.accountId);
-      const extrinsic = getExtrinsic[transaction.type](transaction.args, api);
+      const multisigExtrinsic = getExtrinsic[transaction.type](transaction.args, api);
 
-      return transactionService.getExtrinsicWeight(extrinsic).then(maxWeight => {
+      return transactionService.getExtrinsicWeight(multisigExtrinsic).then(maxWeight => {
         if (accountUtils.isFlexibleMultisigAccount(account)) {
-          const originalExtrinsic = getExtrinsic[transaction.type](transaction.args, api);
 
           const proxyTransaction: Transaction = {
             type: TransactionType.PROXY,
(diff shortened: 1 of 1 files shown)


src/renderer/features/multisig-wallet/index.tsx
```diff
@@ -170,21 +170,8 @@ transactionSDK(multisigWalletFeature, {
...
-            otherSignatories,
-            maybeTimepoint: null,
-            call: encodedTransaction.callData,
-            maxWeight,
-          },
-        };
-
         if (accountUtils.isFlexibleMultisigAccount(account)) {
-          const multisigExtrinsic = transactionService.encodeTransaction(multisigTransaction, api);
+          const encodedTransaction = transactionService.encodeTransaction(transaction, api);
```
valentunn: Looks like `encodedTransaction` here will be exactly the same as on line 169

src/renderer/features/multisig-wallet/index.tsx
```diff
@@ -170,21 +170,8 @@ transactionSDK(multisigWalletFeature, {
       const extrinsic = transactionService.createExtrinsic(transaction, api);
 
       return transactionService.getExtrinsicWeight(extrinsic).then(maxWeight => {
-        const multisigTransaction: MultisigTransaction = {
```
valentunn, line 177: Its incorrect to use `maxWeight` of original transaction for flexible multisig since it additionaly wraps original tx into proxy. So if we want to do `transfer`, then weight of `transfer` will be different (and smaller) than the weight of `proxy(transfer)` so `maxWeight` used here might be underestimated and result in a failed msig execution

src/renderer/features/multisig-wallet/index.tsx
```diff
@@ -170,21 +170,8 @@ transactionSDK(multisigWalletFeature, {
...
-          args: {
-            threshold: account.threshold,
-            otherSignatories,
-            maybeTimepoint: null,
-            call: encodedTransaction.callData,
-            maxWeight,
-          },
-        };
-
         if (accountUtils.isFlexibleMultisigAccount(account)) {
```
valentunn: The branching in this function looks quite complex due to an attempt to combine flex and non flex flow into a single path. I suggest to explicitly separate them so you dont need to maintain "this path is only for flex" in your head while reading. It willl result in more code but will be easier to understand. Plus this function is big already and needs to be split

src/renderer/features/multisig-wallet/index.tsx
aaa123eee: yeah I was trying to reduce intersections and went a bit too far, thanks!

src/renderer/features/multisig-wallet/index.tsx
aaa123eee: fixed, thanks!!

johnthecat (approved):

src/renderer/features/multisig-wallet/index.tsx
johnthecat: We stick with this approach for now, it's counter-intuitive, but it's easier to maintain this large and ugly blob of if's)

aaa123eee: left some comments for easier reading

however we decided to leave it here so far until we come up with a better fitting entity

PR #4708, opened 2025-09-19
refactor: show orig extrinsic

aaa123eee:
## Description

use extrinsic from transaction for json view in operation

## Changes Made

- mv formatExtrinsic to entities/transaction 
- use original transaction extrinsic in operations for json view

## Screenshots/Recordings

image

## Checklist


src/renderer/entities/transaction/lib/transactionService.ts
```diff
@@ -363,6 +369,76 @@ function verifySignature(payload: Uint8Array, signature: HexString, accountId: A
...
+
+  if (arg instanceof GenericMultiAddress) {
+    if (arg.type === 'Id' || arg.type === 'Address20' || arg.type === 'Address32') {
+      return toAddress(arg.value.toString(), { prefix: chain.addressPrefix });
+    } else {
+      return arg.toHuman();
+    }
+  }
+
+  const isAccount = pjsSchema.accountId.safeParse(arg);
```
valentunn, line 401: Not any 32/20 byte arrays are account ids. For example, there are plenty of 32-byte hashes, e.g. call hash, block hash, e.t.c

Instead, we should check whether `arg` is an instance of `AccountId32`/`AccountId20`

src/renderer/entities/transaction/lib/transactionService.ts
```diff
@@ -363,6 +369,76 @@ function verifySignature(payload: Uint8Array, signature: HexString, accountId: A
   }
 }
 
+function formatArg(arg: Codec, chain: Chain): unknown {
```
valentunn, line 374: Nit: we might want to split this function so it rather uses a set of formatters, each one formatting one thing at a time.
Iirc we have `pipeline` primitive for that or something. Or it can be just a simple array of formatters as well

src/renderer/features/multisig-operations/components/OperationFullInfo.tsx
```diff
@@ -47,10 +46,16 @@ export const OperationFullInfo = memo(({ operation, account }: Props) => {
...
 
-  const jsonArgs = useMemo(
-    () => operation.callData && decodeCallData(api, account.accountId, operation.callData, chains),
-    [api, account.accountId, operation.callData, chains],
-  );
+  const jsonArgs = useMemo(() => {
+    if (!operation.callData || !api || !chain) return null;
+
+    try {
+      const extrinsic = getTxFromCallData(api, operation.callData);
```
valentunn: It is strange that we first need to construct extrinsic in order to format call-data. There is nothing in extrinsic that is really required to format a call-data.
So I would suggest to let `formatExtrinsic` actually be `formatCall` and accept `CallBase` instead

src/renderer/entities/transaction/lib/transactionService.ts
aaa123eee: do you mean that the check we have above is enough?
image

cuteWarmFrog (approved):

PR #4723, opened 2025-09-22
fix: flexible change signatories proxy deposit for pure proxied

### LINKED ISSUE #4700: Flexible multisig. Edit flow. Wrong validation for proxy deposit
## Description

There is a wrong validation for proxy deposit.

Image

In Edit flow we need to check:
- Signatory
1. He can pay network fee
2. He can pay for the multisig deposit

- Flexible multisig account (actual pure proxy)
1. He can pay for add_proxy connection

cuteWarmFrog:
## Description
Validator flexible change signatories proxy deposit for pure proxied

## Related Issues
Closes #4700

## Screenshots/Recordings
image
image

## Checklist

### DIFF
--- a/src/renderer/features/flexible-change-signatories/model/change-signatories-model.ts
+++ b/src/renderer/features/flexible-change-signatories/model/change-signatories-model.ts
@@ -1,3 +1,4 @@
+import { type ApiPromise } from '@polkadot/api';
 import { BN } from '@polkadot/util';
 import { combine, createEvent, restore, sample } from 'effector';
 import { createGate } from 'effector-react';
@@ -5,7 +6,9 @@ import { delay, spread } from 'patronum';
 
 import { proxyService } from '@/shared/api/proxy';
 import { type Wallet } from '@/shared/core';
+import { createStoreFromEffect } from '@/shared/effector';
 import { Step, nonNullable, nullable, toAccountId, toAddress } from '@/shared/lib/utils';
+import { type AccountId } from '@/shared/polkadotjs-schemas';
 import { Paths } from '@/shared/routes';
 import {
   createComplexTxStore,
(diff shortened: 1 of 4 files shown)


src/renderer/features/flexible-change-signatories/model/change-signatories-model.ts
```diff
@@ -179,7 +181,29 @@ const { $tx, $fee, $pendingFee } = createComplexTxStore({
...
+  additionalBalanceRules: [
+    ({ route, getBalance, asset, api }) => {
+      const initiator = accountService.findInitiator(route);
+      if (!initiator || !accountUtils.isFlexibleMultisigAccount(initiator)) return;
+
+      const balance = getBalance(initiator.accountId, initiator.chainId, asset.assetId);
+
+      if (nullable(balance)) return;
+
+      const deposit = proxyService.getProxyDeposit(api, '0', 1);
```
valentunn: This is not correct - at this point, there is already a proxy deposit made, so the runtime wont charge use the full `deposit + factor` bur rather just `factor`

This can be accounted by passing already existing deposit instead of `'0'` here
Also we should use `proxyNumber + 1` here since we cannot guarantee this flex msig has just 1 proxy, not to mentioned it will be 2 not 1, since we are adding second proxy in the simplest case

src/renderer/features/operations/OperationsValidation/lib/add-proxy-rules.ts
```diff
@@ -74,4 +76,26 @@ export const AddProxyRules = {
...
+      const initiator = accountService.findInitiator(route);
+      if (!initiator) return;
+
+      const chainId = api.genesisHash.toHex();
+
+      const balance = getBalance(initiator.accountId, chainId, asset.assetId);
+
+      if (nullable(balance)) return;
+
+      const deposit = proxyService.getProxyDeposit(api, '0', proxyNumber + 1);
```
valentunn: Same here. We should use existing deposit here instead of '0'. Overall, '0' can only be used if it is the very `first` proxy being created

valentunn (changes requested):

valentunn (approved):

PR #4760, opened 2025-09-25
refactor(operations): parse transfer asset

aaa123eee:
## Description

Properly pass transfer asset for transfer operations

## Changes Made

- use coreTx for transaction asset
- find coreTx in scm transaction
- provide native asset id for non asset transactions
- avoid passing plain operation, use a respective transaction instead
- check for typeExtras

## Screenshots/Recordings

image

image

image

## Checklist


src/renderer/entities/transaction/lib/callDataDecoder.ts
```diff
@@ -188,15 +189,26 @@ const getDecodedTransaction = (
   };
 };
 
+const getNativeAssetId = (chains: Record<ChainId, Chain>, chainId: ChainId) => {
+  const nativeAsset = getNativeAsset(chains[chainId].assets);
+  const typeExtras = nativeAsset.typeExtras;
+  const nativeAssetId = (typeExtras && 'assetId' in typeExtras && typeExtras.assetId) || nativeAsset.assetId;
+  return nativeAssetId?.toString();
+};
```
valentunn, line 197: Can we clean up the types here?
`nativeAsset.assetId` is a local identifier, which will be `0` for native assets
`typeExtras.assetId`is, on the other hand, onc-chain id that is used to query onchain data. This can be present for native asset if native asset is using not balances pallet, but assets one. Example of such native asset is INTR on Interlay iirc

So, for now I cannot wrap mu head around what is happening here - we are mixing up local and remote ids which can for sure lead to bugs

src/renderer/entities/transaction/hooks/useTransactionAsset.ts
```diff
@@ -28,14 +31,15 @@ export const useTransactionAsset = (operation: MultisigOperation) => {
         return foundAsset;
       }
 
-      if (api) {
-        return getAssetByTypeExtras(api, chain.assets, targetAssetId);
+      const assetByTypeExtras = api && getAssetByTypeExtras(api, chain.assets, targetAssetId);
```
valentunn: The `foundAsset = chain.assets.find()` above seems to be compelted covered by `getAssetByTypeExtras`. So we can just do `getAssetByTypeExtras`
image

src/renderer/entities/transaction/hooks/useTransactionAsset.ts
```diff
@@ -28,14 +31,15 @@ export const useTransactionAsset = (operation: MultisigOperation) => {
...
+      if (assetByTypeExtras) {
+        return assetByTypeExtras;
       }
 
-      return null;
-    } else {
-      return getAssetById(operation.transaction.args.asset, chain.assets) ?? null;
+      return getAssetById(coreTx.args.assetId, chain.assets) ?? null;
```
valentunn, line 31: When we are processing `coreTx.args` can it actually contain a local identifier? In my understanding, it always contains the remote id but I might be wrong. I want us to simplify this method as much as possible since currently it is a complete mess

src/renderer/entities/transaction/lib/common/utils.ts
```diff
@@ -158,7 +158,7 @@ export const findCoreBatchAll = (coreTx: Transaction | DecodedTransaction): Tran
 export const findCoreTransaction = (tx: DecodedTransaction | null): DecodedTransaction | null => {
   if (!tx) return null;
 
-  if (isProxyTransaction(tx)) {
+  if (isProxyTransaction(tx) || isXcmTransaction(tx)) {
```
johnthecat: It's shouldn't work this way, Xcm transaction are already unwrapped

valentunn: This looks very sus - why exactly those two cases are used for unwrapping?

src/renderer/features/multisig-operations/components/Operation.tsx
```diff
@@ -28,6 +35,9 @@ export const Operation = memo(({ operation, multisigAccount }: Props) => {
   const { t } = useI18n();
 
   const showCoreTransaction = accountUtils.isFlexibleMultisigAccount(multisigAccount);
```
valentunn, line 37: Is this something we should move to the account sdk? @johnthecat

src/renderer/entities/transaction/lib/common/utils.ts
aaa123eee: gq

I can relatively reliable go inside of `transaction.args?.transaction`

src/renderer/entities/transaction/hooks/useTransactionAsset.ts
aaa123eee: done 

@kalininilya pls check that it is aligned with ur PR https://github.com/novasamatech/nova-spektr/pull/4570

aaa123eee: checked, looks good

src/renderer/entities/transaction/hooks/useTransactionAsset.ts
valentunn: PS in the current implementation it actually can since for native tokens their local id is used

johnthecat (approved):

PR #4763, opened 2025-09-25
Clean up usage of assetId

valentunn:
* Introduce clean separation between on-chain and local ids by introducing`OnChainAsset` and using `AssetId` more
* Unify asset transaction args by removing `assetId` from call data decoder. We always use `asset` now
* Introduce `getOnChainAssetId` which is a replacement for `getAssetId` which was mixing local and on-chain ids
* Remove `getAssetByTypeExtras`

NOTE: I have not tested this PR so someone should go through it

### DIFF
--- a/src/renderer/shared/lib/utils/assets.ts
+++ b/src/renderer/shared/lib/utils/assets.ts
@@ -1,36 +1,48 @@
-import { type ApiPromise } from '@polkadot/api';
-
-import { type Asset, AssetType, type OrmlExtras, StakingType, type StatemineExtras } from '@/shared/core/types/asset';
+import {
+  type Asset,
+  type AssetId,
+  AssetType,
+  type OrmlExtras,
+  StakingType,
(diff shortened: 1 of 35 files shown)


src/renderer/shared/lib/utils/assets.ts
```diff
@@ -1,36 +1,48 @@
...
-    const firstAsset = assets.at(0);
-    assert(firstAsset, 'Native asset is not found');
-    return firstAsset;
-  }
+/**
+ * Asset id as seen by the chain. Useful when we want to transact specifying
+ * this asset. Note that this can have different type depending on the type of
+ * the corresponding asset
+ */
+export type OnChainAssetId = string | number | 'Native';
```
valentunn, line 16: Can someone confirm whether storing 'Native' for native in the db is fine in terms of backward compatibility?

valentunn:
Drafted untill correctness is verified

Issue #4776, opened 2025-09-26
CoinGecko request. Empty data for fiat on cold start


stepanLav:
## Description

When user run the app first time - fiat prices will be empty.

First coingecko request is sent without any parameters, and return empty dict - {}

<img width="1624" height="1056" alt="Image" src="https://github.com/user-attachments/assets/584cd936-466a-4993-b369-19abd5bf95f1" />

Steps to reproduce:
1. Clear app data via browser settings
2. Reload the page
3. Pair app with any wallet without additional reload

Expected result:
In the app has fiat data from coingecko

Actual result:
Send empty request to coingecko which result to unending loading state

stepanLav:
Status verification: ✅

PR #4777, opened 2025-09-26
fix: combine required data streams for assets fetch

aaa123eee:
## Description

Use combination of streams instead of clock. 

the currency store is not the only trigger we should rely on bc e.g. chains store is falsely non-empty on init

## Related Issues

https://github.com/novasamatech/nova-spektr/issues/4776

## Changes Made

- move currency as an equal source instead of using it as trigger
- check for chains store to be filled

## Checklist


Yanaty (approved):

src/renderer/entities/price/model/price-provider-model.ts
```diff
@@ -62,10 +62,10 @@ sample({
 });
 
 sample({
-  clock: currencyModel.$activeCurrency,
-  source: { chains: networkModel.$chains, provider: $priceProvider },
-  filter: ({ provider }, currency) => nonNullable(provider) && nonNullable(currency),
-  fn: ({ chains, provider }, currency) => {
+  source: { chains: networkModel.$chains, provider: $priceProvider, currency: currencyModel.$activeCurrency },
+  filter: ({ chains, provider, currency }) =>
+    !!Object.values(chains).length && nonNullable(provider) && nonNullable(currency),
```
savvyer, line 67: Can we use `nonNullableMap(chains)` ?

savvyer (approved):

src/renderer/entities/price/model/price-provider-model.ts
aaa123eee: it is always nonNullable, the init state is `{}`

PR #4782, opened 2025-09-28
fix: calculating stakable amount as reservable, similar to Nova Wallet

johnthecat:
Also added amount validation on Stake and Stake more forms

### DIFF
--- a/src/renderer/features/operations/OperationsValidation/lib/bond-extra-rules.ts
+++ b/src/renderer/features/operations/OperationsValidation/lib/bond-extra-rules.ts
@@ -2,9 +2,10 @@ import { BN } from '@polkadot/util';
 import { type Store } from 'effector';
 import { t } from 'i18next';
 
-import { formatAmount } from '@/shared/lib/utils';
+import { type Chain } from '@/shared/core';
+import { assert, formatAmount } from '@/shared/lib/utils';
 import { createTxValidator } from '@/shared/transactions';
-import { type AnyAccount } from '@/domains/network';
+import { type AnyAccount, accountService, balanceService } from '@/domains/network';
 import {
   type AmountFeeStore,
   type BondAmountBalanceStore,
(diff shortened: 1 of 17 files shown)


src/renderer/features/operations/OperationsValidation/lib/bond-extra-rules.ts
```diff
@@ -116,4 +117,24 @@ export const BondExtraRules = {
   },
 };
 
-export const bondExtraValidator = createTxValidator();
+export const bondExtraValidator = createTxValidator<{ chain: Chain; amount: string }>({
+  additionalBalanceRules: [
+    ({ route, asset, chain, amount, getBalance }) => {
+      const initiator = accountService.findInitiator(route);
+      assert(initiator, 'Initiator not found');
```
valentunn, line 124: I have seen this code used multiple times, maybe we can introduce something like `findInitiatorOrThrow`?

src/renderer/features/operations/OperationsValidation/lib/bond-extra-rules.ts
```diff
@@ -116,4 +117,24 @@ export const BondExtraRules = {
...
+
+      const amountBN = new BN(formatAmount(amount, asset.precision));
+      if (amountBN.isZero()) return;
+
+      const balance = getBalance(initiator.accountId, chain.chainId, asset.assetId);
+      assert(balance, `Balance for account ${initiator.accountId} not found`);
```
valentunn, line 130: Same here

valentunn (approved):

src/renderer/features/operations/OperationsValidation/lib/bond-extra-rules.ts
johnthecat: Next time! ✌️🌚

PR #4916, opened 2025-10-21
feat: support balance providers and consumers handling

johnthecat:


src/renderer/domains/network/balance/service.ts
```diff
@@ -67,17 +67,19 @@ function tryReserve(balance: Balance, amount: BN, transferableMode?: Transferabl
 
 function tryWithdraw(balance: Balance, amount: BN, balancePreservation: BalancePreservation): BalanceUpdateResult {
   const transferable = transferableAmountBN(balance);
+  // reset strategy to keep alive if withdraw will lead to account slashing
```
valentunn: dusting/reaping. slashing is a term used for staking punishments

src/renderer/shared/api/balances/service/balanceService.ts
```diff
@@ -234,10 +237,14 @@ function subscribeStatemineAssetsChange(
...
         chainId: chain.chainId,
         assetId: asset.assetId,
+        assetType: asset.type,
         frozen: BN_ZERO,
         reserved: BN_ZERO,
         free,
         ed,
+        providers: 0,
+        consumers: 0,
+        sufficients: 0,
```
valentunn, line 247: Can we add providers only when NATIVE type is used? So we wont store "placeholder" values for other types and maintina more consistent state

valentunn (approved):

PR #4930, opened 2025-10-23
Fix/fee_loading_state

stepanLav:
## Description

**Problem**
When opening transfer modals or recalculating fees, the UI briefly displayed "0" instead of showing a loading indicator. This caused test failures and poor user experience, as users couldn't distinguish between an actual zero fee and a fee that was still being calculated.

**The issue occurred when:**

Opening a transfer modal (before initial fee calculation)
Changing transaction parameters (during recalculation)
Fee calculation failures

**Root Cause**
In createFeeCalculator.ts, the fee store was being reset to BN_ZERO in two scenarios:
When the extrinsic became null (during transaction recalculation)
When fee calculation failed
However, $pendingFee (which was simply fetchFeeFx.pending) would be false in these cases because the effect wasn't running. This created a state where fee = 0 and isLoading = false, causing the UI to render "0" instead of a loading state.

## Changes Made

Modified createFeeCalculator.ts to properly track the initialization state:
Added $isInitialized flag - Tracks whether a fee has been successfully calculated at least once
Removed fee resets - Fee now preserves its previous value instead of resetting to zero
Enhanced $pendingFee logic - Now combines fetchFeeFx.pending with $isInitialized:
Reset initialization on extrinsic change - When extrinsic becomes null, $isInitialized is reset to false, ensuring the loader shows during recalculation

## Checklist

### DIFF
--- a/src/renderer/shared/transactions/createFeeCalculator.ts
+++ b/src/renderer/shared/transactions/createFeeCalculator.ts
@@ -1,9 +1,9 @@
 import { type SubmittableExtrinsic } from '@polkadot/api/types/submittable';
 import { BN, BN_ZERO } from '@polkadot/util';
-import { type Store, type UnitValue, createEffect, createStore, sample } from 'effector';
+import { type Store, type UnitValue, combine, createEffect, createStore, sample } from 'effector';
 
 import { takeLast } from '@/shared/effector';
-import { nonNullable, nullable } from '@/shared/lib/utils';
+import { nullable } from '@/shared/lib/utils';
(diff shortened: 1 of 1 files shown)


johnthecat:
IMO it's better to have a null as default value, it will solve this inconsistency
There is another PR that addresses this exact problem
https://github.com/novasamatech/nova-spektr/pull/4933

aaa123eee:
solved here https://github.com/novasamatech/nova-spektr/pull/4933

PR #5149, opened 2025-11-21
fix: fork protection

### LINKED ISSUE #5106: Multisigs. Duplicates in operations
## Description

There is a possibility to create duplicates in multisig operations

Image

Test database:

[spektr-database (8).json](https://github.com/user-attachments/files/23601991/spektr-database.8.json)

## Steps to Reproduce

1. Create operation
2. Go to Operations page
3. Wait a bit

## Expected Behavior

There is no possibilities to have duplicates in operations

## Actual Behavior

There are duplications

## Additional info
That happens here:
https://github.com/novasamatech/nova-spektr/blob/5f8b6a3f7a13611a05a59e122a06aee4bdc8ffa0/src/renderer/domains/network/multisig-operation/resource.ts#L322-L323

due to the fact, that this data should be taken from the storage

Image

tested data - 

[spektr-database (8).json](https://github.com/user-attachments/files/23603009/spektr-database.8.json)

call hash:
0x4a394c0bf86ace58cbc2ba4bcfe6fe1ad1c2277f0ea9914205486f77640a5986

The bug is in subscribeEventsResource, the code incorrectly uses data.timepoint.height and data.timepoint.index to generate the operation ID. However, this timepoint refers to when the EVENT happened, NOT when the operation was originally created.

cuteWarmFrog:
## Description
Sometimes forks can happen, we want to rely on chain data, which is always true, so we override pending operations for given users/chains;

## Related Issues
Closes #5106

## Checklist

### DIFF
--- a/src/renderer/domains/network/multisig-operation/resource.ts
+++ b/src/renderer/domains/network/multisig-operation/resource.ts
@@ -228,8 +228,6 @@ const fetchAllOnchainOperations = async (
   const requests = Object.values(apis).map(api => fetchOnchainOperations(api, accountId, chains));
   const operations = await Promise.allSettled(requests);
 
-  console.log('onchain operations', operations);
-
   return operations.map(result => (result.status === 'fulfilled' ? result.value : [])).flat();
 };
 
--- a/src/renderer/domains/network/multisig-operation/resource.ts
+++ b/src/renderer/domains/network/multisig-operation/resource.ts
@@ -45,7 +45,7 @@ const operationsGqlSchema = z.object({
 
 const operationsQuery = gql`
   query List($accountId: String!) {
-    multisigOperations(filter: { accountId: { equalTo: $accountId } }) {
+    multisigOperations(filter: { accountId: { equalTo: $accountId }, status: { notEqualTo: pending } }) {
       nodes {
         status
         accountId
@@ -258,7 +258,7 @@ export const fetchResource = createRemoteResource<RequestParams, MultisigOperati
     const chainOperations = await fetchAllOnchainOperations(apis, accountId, chains);
     const historicOperations = await fetchOperationsHistory(accountId, apis, chainOperations, chains);
 
-    return [...chainOperations, ...historicOperations];
+    return chainOperations.concat(historicOperations);
   },
 });
 
--- a/src/renderer/domains/network/multisig-operation/resource.ts
+++ b/src/renderer/domains/network/multisig-operation/resource.ts
@@ -228,6 +228,8 @@ const fetchAllOnchainOperations = async (
   const requests = Object.values(apis).map(api => fetchOnchainOperations(api, accountId, chains));
   const operations = await Promise.allSettled(requests);
 
+  console.log('onchain operations', operations);
+
   return operations.map(result => (result.status === 'fulfilled' ? result.value : [])).flat();
 };
 
--- a/src/renderer/domains/network/multisig-operation/service.ts
+++ b/src/renderer/domains/network/multisig-operation/service.ts
@@ -13,7 +13,7 @@ import {
   type FlexibleMultisigAccount,
   type MultisigAccount,
 } from '@/shared/core';
-import { isEqual, merge, nonNullable, nullable, validateCallData } from '@/shared/lib/utils';
+import { groupBy, isEqual, merge, nonNullable, nullable, validateCallData } from '@/shared/lib/utils';
 import { type AccountId, pjsSchema } from '@/shared/polkadotjs-schemas';
 import { transactionService } from '../transaction/service';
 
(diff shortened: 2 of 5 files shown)


src/renderer/domains/network/multisig-operation/service.ts
```diff
@@ -129,10 +129,23 @@ const mergeEvents = (oldEvents: MultisigEvent[], events: MultisigEvent[]) =>
...
+  operations: MultisigOperation[],
+): MultisigOperation[] => {
+  const newPendingOperations = operations.filter(operation => operation.status === 'pending');
+
+  const oldFilteredForPending = oldOperations.filter(
+    operation =>
+      operation.status !== 'pending' ||
+      !newPendingOperations.some(
+        newOperation => newOperation.chainId === operation.chainId && newOperation.accountId === operation.accountId,
+      ),
```
valentunn: Why do we need this condition? Why not to simply merge all old-not-pending with all new pending?

src/renderer/domains/network/multisig-operation/service.ts
```diff
@@ -129,10 +129,23 @@ const mergeEvents = (oldEvents: MultisigEvent[], events: MultisigEvent[]) =>
     sort: (a, b) => a.blockCreated - b.blockCreated,
   });
 
-const mergeMultisigOperations = (a: MultisigOperation[], b: MultisigOperation[]): MultisigOperation[] => {
+const mergeMultisigOperations = (
```
valentunn, line 145: Overall its totally unclear how this function is supported to work: it does filtering two times, merging and sorting.
I think we should add tests or at least the comment that explains how this function is intended to work

src/renderer/domains/network/multisig-operation/store.ts
```diff
@@ -40,6 +40,14 @@ const removeTransactionsFx = createEffect((operations: MultisigOperation[]) => {
 });
 
 const syncOperationsFx = createQueuedEffect(async (operations: MultisigOperation[]) => {
+  const dbOperations = await storageService.multisigOperations.readAll();
```
valentunn, line 48: Are `operations` here a the result of `mergeMultisigOperations` from the other file?
The naming here is too generic, `operations` does not carry any insight about what is the content of the list

src/renderer/domains/network/multisig-operation/service.ts
cuteWarmFrog: In case of fork we can be in a case where there is an operation in our db/in memory which is no longer in the chain or in the indexer (was executed/canceled)

src/renderer/domains/network/multisig-operation/service.ts
cuteWarmFrog: Added a comment

src/renderer/domains/network/multisig-operation/store.ts
cuteWarmFrog: Added a comment as well + renamed var

src/renderer/domains/network/multisig-operation/service.ts
```diff
@@ -129,10 +129,28 @@ const mergeEvents = (oldEvents: MultisigEvent[], events: MultisigEvent[]) =>
...
+  operations: MultisigOperation[],
+): MultisigOperation[] => {
+  /**
+   * Because we work with Events AT BEST, it may happen that our db has extra
+   * operations that are not on the chain anymore nor were executed. We want to
+   * filter them out.
+   *
+   * For example, if we have a fork and operation was executed in the different
+   * block than we first recevied event of.
+   */
```
valentunn: This is not what I meant by a meaninfull comment.
Okay, just look at `mergeMultisigOperations`: how quickly can you infer what is the expected output of this function and which assumptions does it work with? This comment is useless as it gives no undestanding of why `mergeMultisigOperations` is implemented the way it is implemented. Please add a comment so an external reader (me) can fully understand all the context needed to evaluate correctness of the function

This will greatly help to maintin this code in the feature

src/renderer/domains/network/multisig-operation/store.ts
valentunn: `syncOperationsFx` is still too generic and confusing. `sync` can be anything. If we are syncing in-memory store to db, then the naming could be 

```ts
const syncInMemoryOperationsToDbFx = createQueuedEffect(async (inMemoryOperations: MultisigOperation[]) => {
```

src/renderer/domains/network/multisig-operation/store.ts
```diff
@@ -39,8 +39,17 @@ const removeTransactionsFx = createEffect((operations: MultisigOperation[]) => {
...
+ * We want to sync operations between database and in-memory store. Its
+ * important to do so because operations can be deleted from the database
+ * without calling removeTransactionsFx (f.e. in case of fork).
+ */
+const syncOperationsFx = createQueuedEffect(async (allOperations: MultisigOperation[]) => {
+  const dbOperations = await storageService.multisigOperations.readAll();
+  const dbOperationIds = dbOperations.map(op => op.id);
+
+  await storageService.multisigOperations.deleteAll(dbOperationIds);
+  await storageService.multisigOperations.insertAll(allOperations.map(serializeOperation));
```
valentunn, line 52: Do we subscribe to `multisigOperations` from our local db? Those two changes are not atomic and if we are subscribing, it might lead to flushing "empty state" or something similar

src/renderer/domains/network/multisig-operation/service.ts
valentunn: After the last commit it makes even less sense than before:

```
 const oldFilteredForPending = oldOperations.filter(
    oldOperation =>
      !operations.some(
        operation => operation.chainId === oldOperation.chainId && operation.accountId === oldOperation.accountId,
      ),
  );

```

Why is it needed? Why `oldFilteredForPending` is derived without even checking for pending?

src/renderer/domains/network/multisig-operation/service.ts
```diff
@@ -129,10 +129,28 @@ const mergeEvents = (oldEvents: MultisigEvent[], events: MultisigEvent[]) =>
...
+  const oldFilteredForPending = oldOperations.filter(
```
johnthecat: Looks like you're removing old operations for chains and accounts that are existed in new operations. let's put it that way in code (like what is `oldFilteredForPending`?) and optimize it. Now it's terribly unoptimized

src/renderer/domains/network/multisig-operation/resource.ts
```diff
@@ -228,6 +228,8 @@ const fetchAllOnchainOperations = async (
   const requests = Object.values(apis).map(api => fetchOnchainOperations(api, accountId, chains));
   const operations = await Promise.allSettled(requests);
 
+  console.log('onchain operations', operations);
```
johnthecat: remove

johnthecat (dismissed):

johnthecat (approved):

PR #5727, opened 2026-03-20
Fix multisig confirmation screen data

stepanLav:
## Summary
- **Wallet field**: now shows the multisig wallet (from `multisigAccount.walletId`), not the currently selected wallet from wallet management
- **Signatory field**: correctly shows the signing account's wallet
- **Account field**: hidden by default, only shown when signatory is a proxied account (proxy case)
- **Account/Sender**: uses `multisigAccount.accountId` (the actual transaction sender) instead of the initiator from context model
- Added `AccountExplorers` (info button with block explorer link) to the Wallet field
- Added operation metadata block (Time Point with explorer link, Call Data with copy + JSON viewer) above network fee on the confirmation screen
- Added `expandDepth` prop to `Json` component for partial collapse of large operation trees
- Applied fixes to both Approve and Reject flows
- Integration tests for approve/reject model data assembly

Before:
Screenshot 2026-03-20 at 14 11 02
Screenshot 2026-03-20 at 14 12 13

After:

1. Right resolve for Wallet
2. Remove unnecessary account (it will show only in case when signing with proxy)
3. Added call data for operation
4. Added time-point with link to block explorer

Screenshot 2026-03-20 at 14 36 27
Screenshot 2026-03-20 at 14 36 32

5. Account shows for proxy signatory (wallet - actual multisig wallet, signatory - who is a participant of that multisig and will sign an operation, account - actual address from which tokens will charge for the fee)

Screenshot 2026-03-20 at 14 38 50


savvyer (dismissed):

cuteWarmFrog:
1. is timepoint value clickable? Couldn't tell only by looking
2. Open json calldata icon feels bad. We use it to open explorers usually, mb use another one...

stepanLav:
> 1. is timepoint value clickable? Couldn't tell only by looking
> 2. Open json calldata icon feels bad. We use it to open explorers usually, mb use another one...

yes, it is done the same as for operations details, basically - reuse all components

savvyer (approved):

PR #5791, opened 2026-04-17
fix(contacts): guard signatories against undefined in BackendContactRow

MaliDo:
## Description

Fixes a crash in the Address Book when contacts loaded from IndexedDB cache have `signatories: undefined` instead of `null`.

## Related Issues

Fixes [SPEK-445](https://novasama-technologies.atlassian.net/browse/SPEK-445)

## Changes Made

- Changed `!== null` to `!= null` (loose equality) for `threshold` and `signatories` guards in `BackendContactRow`
- Loose equality covers both `null` and `undefined`, handling old cached DB rows that predate the `BackendContact` type

## Root Cause

Old IndexedDB-cached `BackendContact` rows were stored before the `signatories` field was introduced. When read back from IndexedDB, missing fields come back as `undefined`, not `null`. The strict `!== null` guard failed to catch this, causing `contact.signatories.length` to throw.

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[SPEK-445]: https://novasama-technologies.atlassian.net/browse/SPEK-445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/renderer/entities/contact/ui/BackendContactRow.tsx
```diff
@@ -55,7 +55,7 @@ export const BackendContactRow = ({ contact, onSendTo }: Props) => {
           {contact.categoryName && <Label variant="blue">{contact.categoryName}</Label>}
           {contact.contactTypeName && <Label variant="gray">{contact.contactTypeName}</Label>}
           {contact.chainName && <Label variant="lightBlue">{contact.chainName}</Label>}
-          {contact.threshold !== null && contact.signatories !== null && (
+          {contact.threshold != null && contact.signatories != null && (
```
savvyer: I would prefer to use nullable() helper instead of != loose inequality

savvyer (dismissed):

src/renderer/entities/contact/ui/BackendContactRow.tsx
MaliDo: fixed

aaa123eee (approved):

PR #5803, opened 2026-04-22
feat: update ChangeSignatories flow to support trust/verified paths

savvyer:
## Description
Flexible multisig “change signatories” previously pushed users down a single path. Users who want a safer, proxy-first workflow (add and verify a new multisig proxy before removing the old one) need an explicit choice, while others may prefer the faster trust-based batch update. This PR adds a mode picker (trust vs verified), wires the verified path to the existing Add proxy flow, and disables or explains the verified option when the wallet or chain cannot create proxies—so we do not advertise a path that cannot be completed.

## Related Issues
Closes https://novasama-technologies.atlassian.net/browse/SPEK-424

## Changes Made
- Introduce ChangeSignatoriesFlow (still exported as ChangeSignatories from @/features/flexible-change-signatories) with a modal to choose trust-based (existing change-signatories multisig edit) vs with verification (delegate via proxy).
- Pass canUseVerifiedPath from wallet details using proxy-creation permissions and networkUtils.isProxySupported for the flexible multisig’s chain.
- Support programmatic opening of nested flows: launchOpen / hideTrigger on the existing ChangeSignatories and AddProxy modals so the picker can open either flow without a duplicate trigger.
- Add flexibleMultisig.changeSignatoriesMode.* copy in en.json (picker titles, mode descriptions, unavailable message).

## Screenshots/Recordings

https://github.com/user-attachments/assets/75d2f536-bc1f-4107-a033-8ec25ae39ac3

## Checklist

- [X] I have performed a self-review of my own code
- [X] I have tested the changes and verified the happy path works as expected
- [X] I have provided screenshot of the working feature (if applicable)
- [X] I have provided description of the changes and any additional context that might help reviewers perform more accurate review
- [X] My code follows the project's coding standards and style guidelines

### DIFF
--- a/src/renderer/features/proxy-add/ui/AddProxy.tsx
+++ b/src/renderer/features/proxy-add/ui/AddProxy.tsx
@@ -1,5 +1,5 @@
 import { useUnit } from 'effector-react';
-import { type PropsWithChildren, useEffect, useState } from 'react';
+import { type PropsWithChildren, useEffect, useRef, useState } from 'react';
 
 import { type Chain, type Wallet } from '@/shared/core';
 import { useI18n } from '@/shared/i18n';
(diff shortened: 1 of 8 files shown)


src/renderer/features/proxy-add/ui/AddProxy.tsx
```diff
@@ -37,19 +52,33 @@ export const AddProxy = ({ wallet, onClose, children }: Props) => {
...
-    if (step === Step.NONE) {
-      setIsModalOpen(false);
+    if (step !== Step.NONE) {
+      hasStartedRef.current = true;
+      return;
+    }
+
+    if (!hasStartedRef.current) return;
+
+    hasStartedRef.current = false;
```
aaa123eee, line 71: why hasStartedRef is a ref not just state?

aaa123eee (approved):

src/renderer/features/proxy-add/ui/AddProxy.tsx
savvyer: hasStartedRef and hasClosedRef are guards that get set and checked in the same tick to stop onClose/flowClosed from firing twice — setState is async, so a state version would read the old value and fire anyway.

PR #5863, opened 2026-05-13
feat: cover address-book connection states

stepanLav:
## Description

Covers all states for the address book global connection state

## Checklist

### DIFF
--- a/src/renderer/features/contacts-navigation/__tests__/connection-indicator.test.ts
+++ b/src/renderer/features/contacts-navigation/__tests__/connection-indicator.test.ts
@@ -1,0 +1,325 @@
+import { type Scope, allSettled, fork } from 'effector';
+import { describe, expect, test, vi } from 'vitest';
+
+import { $features, updateFeatureStatus } from '@/shared/config/features';
+import { toAccountId } from '@/shared/lib/utils';
+import { HttpError } from '@/domains/backend';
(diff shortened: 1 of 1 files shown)


savvyer (approved):

src/renderer/features/contacts-navigation/__tests__/connection-indicator.test.ts
```diff
@@ -0,0 +1,325 @@
...
+
+/**
+ * Mirrors the inline logic of `AddressBookStatusDot` +
+ * `AddressBookReconnectPill`. Tests assert the derived UI state — kept here in
+ * one place so a future refactor that extracts this into the model layer still
+ * satisfies the same matrix.
+ */
+type DotState = 'hidden' | 'positive' | 'warning' | 'default';
+
+function readIndicatorState(scope: Scope): { dot: DotState; pillVisible: boolean } {
```
MaliDo, line 18: This helper reimplements the same decision tree as `AddressBookStatusDot` and `AddressBookReconnectPill`, so the tests can pass even if the real UI logic regresses. Could we either extract this derived indicator state into a shared model/helper and test that directly, or render the actual components here and assert the real dot/pill output?

Actually, it would be nice to move the indicator decision logic out of the test/helper and out of the components into connection-status-model.ts, then have both the UI and tests consume that model.
That will make the tests cover the real contract instead of a copied version of it.

MaliDo (changes requested):

PR #5924, opened 2026-06-10
fix: verify proxy remark field and operation modal scroll

stepanLav:
### Remark instead of memo

- The form field, confirmation row and multisig operation details are renamed **Memo → Remark**, matching the wording of the Drafts flow.
- The user text now lives under the `remark` key of the on-chain marker payload (`VerifyProxyMarkerPayload`). The parser keeps accepting the legacy `memo` key, so pings submitted by older builds still match and display correctly (covered by tests).
- An info button next to the Remark field (same pattern as the execution-mode popover in the flexible multisig edit flow) shows the exact JSON payload that will go on chain, live-updated as the user types. The popover and the transaction build share one payload helper, so the preview cannot drift from what is submitted.
- Added a feature spec `features/proxy-verify/README.md` describing the remark payload, states and lifecycle.

### Modal scroll fix (app-wide)

`Modal.Content disableScroll` skips the built-in ScrollArea while the modal is capped at `min(100%, 736px)` with `overflow-hidden` — tall confirmation steps clip the sign button with no way to reach it. An audit of all 31 `disableScroll` usages found 4 init→confirm→sign flows with this bug; `disableScroll` is removed from them (same setup as `RemoveProxy`, which already scrolls correctly):

- `proxy-verify/VerifyProxy` (reported)
- `multi-transfer/MultiTransferModal`
- `call-data-execute/CallDataSubmitModal`
- `vested-transfer/VestedTransferModal`

The remaining usages have internal scroll areas or bounded layouts and are intentionally untouched. Borderline cases that may deserve a follow-up visual check: dapp-browser `SignCustomPayloadModal` (confirm step), `WalletConnectDetails` accounts tab, fellowship-evidence `IPFSUploadModal`, `VotingHistoryDialog`, `FellowshipOverviewModal`.

Additionally there is problem to show parsed data on Polkadot Vault, when data inside remark operation - shows as byte array, it works for Legacy QR format, but broken for the New metadata-less signing, fixing here:

https://github.com/novasamatech/parity-signer/pull/2549

## Screenshots/Recordings

https://github.com/user-attachments/assets/1dad7c48-697e-4339-97bf-fe5a89cadb77

## Checklist

Fixes novasamatech/nova-spektr-tracker#77

### DIFF
--- a/src/renderer/aggregates/backend/README.md
+++ b/src/renderer/aggregates/backend/README.md
@@ -1,0 +1,56 @@
+# Address book backend connection & authentication
+
+## Overview
+
+Lets the user connect Nova Spektr to a self-hosted address-book backend and prove account ownership by signing a
+challenge. Once authenticated, contacts sync with the backend; the aggregate keeps the session alive and surfaces expiry
+or network issues.
+
+## Who can use it / when it applies
+
+- Reached from the address book via the backend configuration modal (add or edit a backend URL).
+- Signing requires at least one signable account: a Polkadot browser-extension account or a Polkadot Vault account.
+
+## States / scenarios
+
+| State             | When it appears                                | What the user sees                                                                          |
+| ----------------- | ---------------------------------------------- | ------------------------------------------------------------------------------------------- |
+| URL entry         | Modal opened with no backend or while editing  | URL input with live reachability probe (checking / reachable / unreachable / wrong backend) |
+| Account selection | URL is valid and the user is not yet connected | Account select, preselected with the previously used account when known                     |
+| Signing           | Connect pressed                                | Signing flow; for Vault accounts also a network selector (see below)                        |
+| Connected         | Session is live and the URL is unchanged       | Connected account with a disconnect action                                                  |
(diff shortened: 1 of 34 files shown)


savvyer (commented):
Removing disableScroll from these flow modals wraps the whole step in Modal.Content’s default ScrollArea, but the rendered steps already contain their own ScrollArea plus footer. Examples: CallDataSubmitModal.tsx:39 with CallDataForm.tsx:54/:127, MultiTransferModal.tsx:60 with MultiTransferForm.tsx:43/:71, and VerifyProxy.tsx:69. This creates nested scroll regions and makes the action footer part of the outer scroll content instead of a fixed modal footer.

cuteWarmFrog (commented):
## Code review — PR #5924 (verify proxy remark + operation modal scroll)

Reviewed the remark-verification field changes and the operation-modal scroll fix, including the full context of each changed file (not just the diff hunks) and the surrounding callers/Effector wiring.

No correctness, removed-behavior, or cross-file regressions found. The field rename is consistent with sensible backward-compatibility, and the scroll fix is sound and doesn't break other modal states.

LGTM. ✅

PR #5925, opened 2026-06-10
fix: extrinsic builder - address suggestions, click area, template apply

stepanLav:
### 1. Pasted full address now surfaces the matching contact
Previously the suggestion list went empty whenever the input was a valid address — the user had to delete characters to see the contact. The query is now decoded to its `accountId` and matched against contacts and wallet accounts, so a full address pasted in **any** ss58 prefix shows the named entry for explicit confirmation before selecting.

### 2. Cleared destination address stays cleared
`handleBlur` skipped the commit for an empty field, so the previous value reappeared after clicking another cell. Blur now commits whatever is in the field, including empty.

### 3. Clickable area limited to the text box
`Field` wrapped the caption ("Pallet", parameter names) and the control in one full-width `<label>`, so clicking the caption focused the combobox, which opens its dropdown on focus. Added an opt-out `clickableLabel` prop to `Field` and applied it to `PalletSelect`, `CallSelect` and `ParameterField`. Field behavior elsewhere is unchanged.

### 4. Complete contact/account list in suggestions
The dropdown was hard-capped at 10 contacts / 20 accounts, which made the on-focus list look incomplete. Caps removed — the list is scrollable.

### 5. Apply template → jump to Build tab
After clicking **Apply** on a template the view now switches to the Build tab automatically (in both the draft-creation and call-data-execute modals), since the call data is already filled and the user's next step is tweaking the arguments.

### Supporting ui-kit change
`Combobox` gained an optional `onSelect` prop fired only on list-item selection, letting the account input distinguish picking a suggestion from typing/pasting. Backwards-compatible — existing consumers are unaffected.

## Screenshots/Recordings

https://github.com/user-attachments/assets/276529c2-54a7-4139-a285-18f7c7db6842

## Checklist

Fixes novasamatech/nova-spektr-tracker#79
Fixes novasamatech/nova-spektr-tracker#80

### DIFF
--- a/src/renderer/features/call-data-execute/model/form.ts
+++ b/src/renderer/features/call-data-execute/model/form.ts
@@ -330,6 +330,13 @@ sample({
   target: form.fields.callData.change,
 });
 
+// jump straight to the builder — the user's next step is tweaking the arguments
+sample({
+  clock: templateApplied,
+  fn: () => InputMode.BUILD,
+  target: inputModeChanged,
+});
+
 // flow setup
 
 // Preselect initiator on form open
--- a/src/renderer/features/drafts/components/CreateDraftModal.tsx
+++ b/src/renderer/features/drafts/components/CreateDraftModal.tsx
@@ -153,6 +153,8 @@ export const CreateDraftModal = () => {
 
   const handleTemplateApply = (templateCallData: string) => {
     createDraftModel.callDataChanged(templateCallData);
+    // jump straight to the builder — the user's next step is tweaking the arguments
+    createDraftModel.inputModeChanged('build');
   };
 
   const handleToggle = (open: boolean) => {
--- a/src/renderer/features/extrinsic-builder/ui/CallSelect.tsx
+++ b/src/renderer/features/extrinsic-builder/ui/CallSelect.tsx
@@ -20,7 +20,7 @@ export const CallSelect = memo(({ options, value, disabled, onChange }: Props) =
   );
 
   return (
-    <Field text={t('extrinsicBuilder.call')}>
+    <Field text={t('extrinsicBuilder.call')} clickableLabel={false}>
       <Combobox
         placeholder={t('extrinsicBuilder.callPlaceholder')}
         value={displayValue}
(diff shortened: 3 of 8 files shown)


savvyer (approved):

cuteWarmFrog (commented):
## Code review — PR #5925 (extrinsic builder QA feedback)

The address-suggestion change in `AccountParamInput.tsx` introduces two behavioral regressions that both stem from `const searchQuery = displayValue` (where `displayValue = isEditing ? inputText : value`).

### 🟠 Major — re-focusing a filled field hides every account except the selected one
`src/renderer/features/extrinsic-builder/ui/parameterInputs/AccountParamInput.tsx:46`

When the field already holds a value and is re-focused, `isEditing` is still `false` (it only flips on `onInput`), so `displayValue === value` (the committed full address) → `searchQuery` is that full address → `queryAccountId` resolves → both `accountOptions` and `contactOptions` filter to `account.accountId === queryAccountId`, i.e. **only the already-selected entry**.

The comment says re-focusing "lists the matching entries instead of the full address book," but the actual effect is that the user **cannot switch to a different account/contact without first manually clearing the input**. That's a worse UX than before.

**Failure scenario:** user picks account A, re-opens the field to change to B → dropdown shows only A.

### 🟠 Major — empty-field dropdown is now unbounded
`src/renderer/features/extrinsic-builder/ui/parameterInputs/AccountParamInput.tsx:60`

The previous `.slice(0, 20)` / `.slice(0, 10)` caps were removed. With an empty field (`searchQuery === ''`), the account filter falls through to `!searchQuery → true` for every entry, rendering **all** accounts and contacts at once.

**Failure scenario:** a wallet with hundreds of accounts/contacts focuses the field with empty input → the full list renders, causing a noticeable render/perf spike. Consider restoring a reasonable cap on the rendered suggestions.

### 🟡 Minor — `handleBlur` commits partial/invalid typed text
`src/renderer/features/extrinsic-builder/ui/parameterInputs/AccountParamInput.tsx:100`

`handleBlur` now commits `inputText` verbatim whenever `isEditing`, with no validity gate. Two consequences:
- If `onBlur` fires before Ariakit's selection (`onSelect` → `handleSelect`) on a mousedown-to-pick, the typed *partial* text could be committed instead of the chosen address — the outcome depends on Ariakit's focus-retention ordering.
- A malformed/incomplete address typed and then blurred is committed as the parameter value and only surfaces later as a generic `encodeCallData` error rather than being rejected at the input boundary.

Consider only committing on blur when the text is a valid address (or matches a known account/contact), otherwise reverting to the prior value.

PR #5927, opened 2026-06-10
feat: remember last used chain for address book pairing

stepanLav:
## Problem

When connecting the address book backend, the signing chain selector always defaulted to the Polkadot relay chain. Users whose keys live on another chain (e.g. Asset Hub) had to re-select their network on every single login.

Closes the second option from the reported feedback: let the default network follow user preference.

## Solution

- New persisted store `$defaultAuthChainId` (localStorage key `address-book-default-auth-chain-id`, `persist` from `effector-storage/local` with `sync: true` per project convention — synchronous hydration, no race with the first modal open).
- First-ever pairing still defaults to the Polkadot relay chain.
- After a successful sign-in, the chain that was used is saved and preselected on every later pairing/re-connection. Failed or cancelled attempts don't change the preference, and it deliberately survives sign-out / URL clear (unlike the last-authed account, which is cleared).
- If the saved chain was removed from the chains config in a later release, the selector falls back to the Polkadot relay chain instead of feeding a stale `chainId` into the Vault QR signing payload.
- Chain list ordering needed no change: `ChainSelect` already sorts options via `chainsService.sortChains`, the same ordering used across the app.

## Spec

Added `src/renderer/aggregates/backend/README.md` — product spec for the backend connection/auth flow, including the new default-chain rules.

## Screenshots/Recordings

https://github.com/user-attachments/assets/73e68bf6-8492-4d9e-acf8-50ad9c8bfda9

## Checklist

Fixes novasamatech/nova-spektr-tracker#76
Fixes novasamatech/nova-spektr-tracker#81

### DIFF
--- a/src/renderer/aggregates/backend/README.md
+++ b/src/renderer/aggregates/backend/README.md
@@ -1,0 +1,56 @@
+# Address book backend connection & authentication
+
+## Overview
+
+Lets the user connect Nova Spektr to a self-hosted address-book backend and prove account ownership by signing a
+challenge. Once authenticated, contacts sync with the backend; the aggregate keeps the session alive and surfaces expiry
+or network issues.
+
+## Who can use it / when it applies
+
+- Reached from the address book via the backend configuration modal (add or edit a backend URL).
+- Signing requires at least one signable account: a Polkadot browser-extension account or a Polkadot Vault account.
+
+## States / scenarios
+
+| State             | When it appears                                | What the user sees                                                                          |
+| ----------------- | ---------------------------------------------- | ------------------------------------------------------------------------------------------- |
+| URL entry         | Modal opened with no backend or while editing  | URL input with live reachability probe (checking / reachable / unreachable / wrong backend) |
+| Account selection | URL is valid and the user is not yet connected | Account select, preselected with the previously used account when known                     |
+| Signing           | Connect pressed                                | Signing flow; for Vault accounts also a network selector (see below)                        |
+| Connected         | Session is live and the URL is unchanged       | Connected account with a disconnect action                                                  |
+| Error             | Challenge or signature verification failed     | Error alert with a "try again" action                                                       |
+| Session expired   | Background health check loses the session      | Toast with a reconnect shortcut                                                             |
+
+### Signing network selection
+
+Vault accounts sign on a specific network, so the signing step shows a network selector (networks listed in the same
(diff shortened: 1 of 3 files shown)


savvyer (approved):

cuteWarmFrog (commented):
## Code review — PR #5927 (remember last used chain for address book pairing)

The core implementation is sound: persistence, read-back, the `chainsMap[selectedChainId] ?? null` UI guard, and the model fallback to Polkadot all behave correctly, and the invalid-chain case is handled in two places. The findings below are all minor / hardening observations — none block merge.

### 🟡 Minor — stale `$defaultAuthChainId` is never repaired in localStorage
`src/renderer/aggregates/backend/model/auth-model.ts:209`

When the saved chain has been permanently removed from config, the fallback corrects `$selectedChainId` per reset but never rewrites the persisted `$defaultAuthChainId`. localStorage keeps the dead `chainId` indefinitely for sessions that never sign in. Consider re-persisting the resolved fallback chain.

### 🟡 Minor — empty-`$chains` reset trusts a possibly-removed saved chainId
`src/renderer/aggregates/backend/model/auth-model.ts:213`

While `$chains` is still rehydrating (`Object.keys(chains).length === 0`), the reset trusts the saved `chainId` unconditionally, so `$selectedChainId` can momentarily hold an invalid chain. If a Vault user manages to press Connect before `$chains` populates, `messageSignModel.init` receives a chainId absent from the chain list. The window is tiny (effector-storage rehydrates synchronously and signing needs multiple clicks), but worth a guard.

### 🟡 Minor — remembered chain is overwritten on every sign-in, including extension accounts
`src/renderer/aggregates/backend/model/auth-model.ts:299`

`verifySignatureFx.done` re-saves `$selectedChainId` for *all* sign-ins. For extension accounts the chain selector is Vault-only and `$selectedChainId` is never user-chosen, so the "remembered" value gets clobbered by a preselected value the user never picked. Consider only saving when the chain was actually user-selected (Vault flow).

### 🟡 Minor — correct preselect relies on implicit sample declaration order
`src/renderer/aggregates/backend/model/auth-model.ts:296`

The post-connect preselect of the freshly-saved chain depends on two samples — both clocked on `verifySignatureFx.done` — running in declaration order (save at ~299 before the connect trigger at ~307). A future reorder would silently break "remember the chain just used for sign-in." A brief comment noting the ordering dependency (or an explicit dependency) would make this robust.

PR #5942, opened 2026-07-06
fix: multisig operation handling with execution results

savvyer:
## Description
Multisig operations completed via live event subscription only distinguished between cancelled (a reject event was seen) and executed — every non-cancelled completion was marked executed. But a MultisigExecuted event carries the inner call's DispatchResult: the multisig extrinsic can succeed while the dispatched call fails. Those failures were silently shown as successful executions.

This surfaces failed inner calls as error on the live path, matching the status the off-chain indexer already assigns (getExecutionResult), so live-caught and indexed operations stay consistent.

## Related Issues
https://novasama-technologies.atlassian.net/browse/SPEK-695

## Changes Made
- resource.ts — the multisig event subscription now normalizes a MultisigExecuted event's DispatchResult into an ExecutionResult ('success' | 'error' | null) and threads it through $completionEvents. Approvals/cancellations carry null (no DispatchResult).
- store.ts — $completedLiveOperations derives cancelled → error (failed execution) → executed, preserving cancellation precedence. Exposed $completedLiveOperations and $removedFromChainStorageOperations on __test.
- store.test.ts — added coverage for the three status derivations: successful execution → executed, failed inner call → error, and cancellation taking precedence over a failed execution.

## Checklist

- [X] I have performed a self-review of my own code
- [X] I have tested the changes and verified the happy path works as expected
- [X] I have provided description of the changes and any additional context that might help reviewers perform more accurate review
- [X] I have added tests that cover the base logic (if applicable)
- [X] My code follows the project's coding standards and style guidelines

### DIFF
--- a/.claude/commands/pr.md
+++ b/.claude/commands/pr.md
@@ -1,0 +1,16 @@
+Generate a pull request description for the current branch.
+
+Instructions:
+1. Run `git diff dev...HEAD --stat` to see which files changed
+2. Run `git diff dev...HEAD` to see the full diff
+3. Analyze the changes and produce a PR description in this format:
+
+.github/PULL_REQUEST_TEMPLATE.md
+
+Rules:
+- Keep it concise — a reviewer should understand the PR in 30 seconds
+- Focus on the "why" not the "what" — the diff shows the what
+- Group related changes together
+- Include specific test steps, not generic ones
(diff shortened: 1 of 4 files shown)


stepanLav (approved):
lgtm

PR #5947, opened 2026-07-07
fix: resolve address book names for Polkadot Vault derived accounts

MaliDo:
## Description
Polkadot Vault derived accounts were flagged as "custom named" both at creation time and by a historical migration, so the address-book/identity resolution chain was always skipped for them — the transfer "From account" selector (and any other consumer of the shared name resolver) showed the raw derivation path or "Main" instead of a saved contact name, even when the same address was already in the address book. Separately, once a name was resolved and cached, it stayed stale if contacts changed while a view was still mounted — reconnecting the backend address book from an open Transfer modal required closing and reopening it to see the update.

To verify: pair a Vault wallet with multiple derived keys, add one of its addresses to the address book with a custom name, open Transfer and check the "From account" dropdown shows that name (and that typing filters the list). Then, with the modal still open, reconnect/disconnect the backend address book and confirm names update without closing the modal.

Also includes draft spec READMEs for `polkadot-vault-wallet` and `polkadot-vault-wallet-pairing` (required by the Feature Map CI check since this PR touches their account-creation code). Written from reading the code — **flagged as pending review from the feature's author** before being treated as source of truth.

## Related Issues
https://novasama-technologies.atlassian.net/browse/SPEK-698

## Changes Made
- Vault account creation (`manage-vault-model.ts`, `polkadot-vault-wallet/service.ts`) now marks newly derived keys `GENERATED` instead of relying on a `CUSTOM` default, making them eligible for address-book resolution going forward.
- New migration (`migration-21.ts`, DB v53) resets existing Vault chain/shard accounts from `CUSTOM` back to `GENERATED`, deliberately excluding the single BASE/universal account created when a pairing yields zero derived keys, since that name really was typed by the user.
- `resolveAccountName` (`service.ts`) gains a fallback tier — the account's own stored name is tried before falling back to a short address, so accounts without a contact/identity match still show "Main"/derivation path rather than a truncated address.
- `AccountSelect` (the transfer "From account" dropdown) now resolves names through the same contact/identity chain as the rest of the app, and supports search, mirroring the recipient selector's `Select` + `onSearch` pattern.
- `resource.ts` tracks which accounts/wallets currently have a resolved name cached and recomputes them whenever contacts change, so mounted views update live instead of only on remount.
- Draft spec READMEs added for `polkadot-vault-wallet` and `polkadot-vault-wallet-pairing`, and linked in the Feature Map.

### Addressed review feedback
A `CHANGES_REQUESTED` review flagged several follow-on issues in the above; all are fixed in a follow-up commit:
- `migration-21.ts` now keys off the owning wallet's type instead of the account's own `signingType` — accounts that migrated from a pre-`accounts2` Multishard wallet kept a stale `PARITY_SIGNER` signingType and were being skipped. Also fixed `.modify()`'s early-exit branches to `return false`, so untouched rows are no longer rewritten on every migration run.
- Live name recompute (`resource.ts`) now reacts to identities/accounts/chains changing, not just contacts, and merges onto the existing cache instead of replacing it wholesale; tracked params are capped to bound memory growth.
- `resolveAccountName` now uses the caller's already-known account when available instead of re-deriving it by `accountId` alone, and otherwise prefers a chain match, then a `CUSTOM`-named account, over array order — fixes a Vault key's raw derivation path being able to shadow a differently-named account that happens to share its `accountId` across wallets. The account-name cache key and `accountsNameResource`'s own request key were both extended with the same per-account identity so two accounts (or two account lists) sharing an `accountId` no longer collide on the same cache/request slot.
- `wallet-model.ts`'s `createWalletFx` now defaults `nameType` based on `accountType` (Vault `CHAIN`/`SHARD` keys → `GENERATED`, everything else keeps the existing `CUSTOM` default) instead of one blanket default, so any future Vault-derived-key creation path is correct by default without relying on every call site to remember to stamp it.

## Screenshots/Recordings
https://github.com/user-attachments/assets/f44fd5c9-cb22-4530-bc5f-c714daa8bd78

## Checklist

### DIFF
--- a/src/renderer/domains/network/account/hooks.ts
+++ b/src/renderer/domains/network/account/hooks.ts
@@ -31,6 +31,7 @@ export const useAccountsNames = (accounts: AnyAccount[], chain?: Chain | null) =
         const key = createAccountNameCacheKey({
           accountId: account.accountId,
           chain,
+          account,
         });
         result[key] = cache[key] ?? account.name;
       }
@@ -42,6 +43,7 @@ export const useAccountsNames = (accounts: AnyAccount[], chain?: Chain | null) =
     const key = createAccountNameCacheKey({
       accountId: account.accountId,
       chain,
+      account,
     });
     const resolvedName = accountNames[key];
     return resolvedName ? { ...account, name: resolvedName } : account;
--- a/src/renderer/domains/network/account/resource.test.ts
+++ b/src/renderer/domains/network/account/resource.test.ts
@@ -1,0 +1,84 @@
+import { describe, expect, it } from 'vitest';
+
+import { CryptoType, SigningType } from '@/shared/core';
+import { createAccountId, polkadotChain, polkadotChainId } from '@/shared/mocks';
+
+import { accountsNameResource, createAccountNameCacheKey } from './resource';
+import { type ChainAccount } from './types';
+
+describe('createAccountNameCacheKey', () => {
+  it('should key on the specific account id, not just accountId, so accounts sharing an accountId across wallets do not collide', () => {
+    const sharedAccountId = createAccountId('shared');
+
+    const vaultDerivedAccount: ChainAccount = {
+      id: 'vault-account',
+      type: 'chain',
+      accountId: sharedAccountId,
+      chainId: polkadotChainId,
+      name: '//polkadot//0',
+      walletId: 1,
+      signingType: SigningType.POLKADOT_VAULT,
+      cryptoType: CryptoType.SR25519,
+      createdAt: Date.now(),
+    };
+    const watchOnlyAccount: ChainAccount = {
+      ...vaultDerivedAccount,
+      id: 'watch-only-account',
+      walletId: 2,
+      name: 'My Watch-Only',
+    };
+
+    const vaultKey = createAccountNameCacheKey({
+      accountId: sharedAccountId,
+      chain: polkadotChain,
+      account: vaultDerivedAccount,
+    });
+    const watchOnlyKey = createAccountNameCacheKey({
+      accountId: sharedAccountId,
+      chain: polkadotChain,
+      account: watchOnlyAccount,
+    });
+
+    expect(vaultKey).not.toBe(watchOnlyKey);
+  });
+
+  it('should fall back to accountId when no specific account is known (e.g. an arbitrary address)', () => {
+    const accountId = createAccountId('arbitrary');
+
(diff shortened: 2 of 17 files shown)


savvyer (changes requested):
Ran a deep review on this PR — a few real issues worth addressing before merge:

Bugs

- Migration-21 misses legacy Vault accounts (migration-21.ts:21) — pre-accounts2 Vault/Parity Signer derived keys carry signingType: 'signing_ps' (migration-2 set it, migration-3 copied it per-account, migration-6 preserved it when regrouping into Vault wallets — only the synthesized BASE account got 'signing_pv'). The signingType !== POLKADOT_VAULT guard skips them, so the oldest Vault users keep CUSTOM and never get address-book resolution — exactly the users this fix targets.

- Live recompute only reacts to contacts (resource.ts:244) — the new samples are clocked solely on $contacts, but resolveAccountName also depends on identity.$list, accounts.$list, and $chains (all already in the sample source). An identity resolving after a modal mounts, or a wallet/account rename, still shows a stale name until remount — the same staleness class the comment above the code promises to fix.

- Second-writer race on the caches (resource.ts:270) — the samples write directly into $accountNameCache/$walletNameCache. A name fetch in flight (its attach snapshotted the old $contacts before an await) can push its stale resolution over a fresh recompute. The full-replacement (next = {}) also drops any cache entry not covered by the params stores.

- relatedAccount?.name fallback is chain/wallet-blind (service.ts:337) — accounts.find(a => a.accountId === accountId) has no chain or wallet filter, so with duplicate accountIds across wallets the shown name depends on store order (e.g. a Vault key's raw //kusama//hot//0 can shadow a user's custom-named watch-only account for the same address).

- Regression trap (wallet-model.ts:80) — createWalletFx defaults nameType ?? CUSTOM while createWalletsFx defaults ?? GENERATED; the fix stamps GENERATED at 3 leaf sites instead of the choke point. Since migration-21 never re-runs, any future creation path that forgets the stamp permanently reintroduces the bug.

Minor

- migration-21.ts: bare return; in .modify doesn't skip the write-back (Dexie requires return false) → the migration rewrites the entire accounts2 table.
resource.ts:200: $accountNameParams/$walletNameParams grow unboundedly (full Wallet snapshots, no eviction) and every backend contact sync rebuilds the whole cache.
- service.ts:344: the chain-prefix fallback branch below the new early return is now dead code (name is a required string).

savvyer (approved):

PR #5948, opened 2026-07-13
feat: vesting schedules

savvyer:
## Description
Adds vesting claim: recipients of vested transfers can now see what has unlocked and release it to their transferable balance (vesting.vest()), surfaced as a callout in the dashboard's Portfolio Overview card. This closes the loop with the existing vested-transfer feature — one side creates schedules, this side lets recipients claim what's vested.

The claim UI is backed by a new vesting-portfolio aggregate that answers "what is vesting, across every account, and can that answer be trusted yet" — gathering live schedules across all non-hidden accounts on every vesting-capable chain and deciding when it's safe to show "no vesting" instead of a stale guess.

Along the way: entities/vesting migrated to domains/vesting (per FSD deprecation), the Portfolio Overview distribution bars gained a Vested category (carved out of Locked, not added beside it), and a couple of correctness bugs surfaced and got fixed.

## Related Issues
https://novasama-technologies.atlassian.net/browse/SPEK-638

## Screenshots/Recordings

https://github.com/user-attachments/assets/8561da85-6053-49cb-b1fa-f96932cc303f

https://github.com/user-attachments/assets/3b2915b0-e483-489d-adcd-38e02fc2b179

Screenshot 2026-07-13 at 10 01 29 PM
Screenshot 2026-07-13 at 9 43 46 PM

Screenshot 2026-07-14 at 4 51 19 PM

## Checklist


src/renderer/features/vesting-claim/model/claim.ts
```diff
@@ -0,0 +1,296 @@
...
+
+sample({
+  clock: sign.filter({ fn: ({ signingPayloads }) => signingPayloads.length > 0 }),
+  target: signModel.events.formInitiated,
+});
+
+sample({
+  clock: signModel.signed,
+  source: $step,
+  filter: (step) => step !== Step.NONE,
```
stepanLav: **[correctness]** Both samples on the global `signModel.signed` (here and at line 240) are gated only by `step !== Step.NONE`, which also passes at `CONFIRM` and `SUBMIT`. A signature emitted by a *different* flow while this one sits at CONFIRM (or is abandoned — `ClaimFlow`'s unmount doesn't reset `$step`, and SIGN's `onGoBack` returns to CONFIRM while an async Vault/WalletConnect sign request may still be pending) gets forwarded to `submitModel.init` with the foreign payload and flips the claim flow to SUBMIT.

The reference flow guards strictly on the signing step (`features/transfer/model/transfer-model.ts:184` uses `isStep(step, Step.SIGN)`).

```suggestion
  filter: (step) => step === Step.SIGN,
```

(Same change needed on the second sample at line 240.)

src/renderer/aggregates/vesting-portfolio/lib/views.ts
```diff
@@ -0,0 +1,311 @@
...
+
+    const asset = getNativeAsset(chain.assets);
+    if (!asset) continue;
+
+    // Vesting schedules are denominated in blocks of the chain's timeline chain
+    // (the relay chain, for migrated Asset Hubs) — so the height *and* the block
+    // time below must both be read from that chain, never from `chain` itself.
+    const timelineChainId = getTimelineChainId(chain);
+    const blockHeight = currentBlock[timelineChainId];
+    if (blockHeight == null) continue;
```
stepanLav, line 169: **[correctness]** This `continue` silently drops the whole chain's account views when the timeline chain's block height is unknown, while `summary.schedulesCount` (line 261) still counts those schedules straight from raw data, and `resolveStatus` returns `'ready'` because `$hasSchedules` is also derived from raw `$data`.

The state is reachable **non-transiently**: for a migrated Asset Hub, `getTimelineChainId` points at the relay chain; if the relay is disabled in network settings (or stuck in ERROR) it never appears in `$apis`, so neither `$desiredHeads` nor the 60s poll in `block/store.ts` ever produces its height. Result: the dashboard callout permanently says “N active vesting schedules” while the schedule modal shows zero rows, $0 totals, and no claim path — with no loader or error, since `countUnresolvedChains` only tracks schedule-cache resolution, not timeline-height resolution.

Consider either counting these chains as unresolved (status stays loading / degrades explicitly) or rendering the rows with an unknown-date placeholder instead of dropping them.

src/renderer/aggregates/vesting-portfolio/lib/views.ts
```diff
@@ -0,0 +1,311 @@
...
+    summary: {
+      totalVestingFiat,
+      claimableFiat,
+      perDayFiat,
+      // Counted straight from the fetched schedules, so the callout can appear as
+      // soon as they load — without waiting for the block height, balances and
+      // prices the figures above depend on.
+      schedulesCount: countSchedules(data),
+      lastUnlockDate,
+      hasClaim: claimableFiat.gt(0),
```
stepanLav: **[correctness]** `hasClaim` is derived from the *fiat* value, but `addFiat` returns 0 whenever `!asset.priceId` or the price entry is missing (lines 142–145). For assets without a price feed (Westend/dev chains, newly listed tokens) or while the CoinGecko fetch fails, the per-account rows correctly show a “Ready” badge and an enabled Claim button (`view.claimable.gtn(0)` in `VestingScheduleModal.tsx:78`, `AccountScheduleModal.tsx:95`), while the callout badge (`VestingCallout.tsx:90`) and the “Ready to unlock” tile (`VestingScheduleModal.tsx:48`) stay hidden — the two surfaces disagree about the same fact.

`hasClaim` should be computed from the token-level BN claimable totals; fiat is display-only. The same fiat-gating also zeroes `totalVestingFiat`/`perDayFiat` tiles for priceless assets.

src/renderer/features/multisig-operations/components/modals/getMultisigSignOperationTitle.ts
```diff
@@ -17,6 +17,7 @@ const TransactionTitlesModal: Record<TransactionType, (crossChain: boolean) => s
   [TransactionType.TRANSFER_ALLOW_DEATH]: crossChain =>
     `operations.modalTitles.${crossChain ? 'transferFrom' : 'transferOn'}`,
   [TransactionType.VESTED_TRANSFER]: () => `operations.modalTitles.vestedTransfer`,
+  [TransactionType.VEST]: () => `operations.modalTitles.vest`,
```
stepanLav, line 20: **[correctness / incomplete integration]** `TransactionType.VEST` got entries here and in `transactionConfirmIcon.ts`, but no feature registers it with `multisigOperationsSDK` — `vested-transfer-operation-details/index.tsx` maps only `VESTED_TRANSFER` in its `getOperationIcon`/`getOperationTitle` transformers.

Since the claim flow supports multisig accounts (multisig deposit, signing-path wrapping), a pending multisig `vesting.vest()` operation is constructible — and co-signers will see it in the operations list with the `unknownMst` icon and the raw `formatSectionAndMethod('vesting', 'vest')` fallback title, instead of the “Claim vested” card the feature README promises (`features/multisig-operations/README.md:59`). Only the sign-modal title works.

Please register icon/title (and details, if applicable) transformers for `VEST` via the SDK like the other operation types.

src/renderer/features/dashboard-portfolio-overview/hooks/useBalanceAllocation.ts
```diff
@@ -12,62 +14,75 @@ export type AllocationData = {
...
-      const asset = chain.assets.find((a) => a.assetId === balance.assetId);
-      if (!asset?.priceId) continue;
+    const priceItem = prices[asset.priceId]?.[currency.coingeckoId];
+    if (!priceItem) continue;
 
-      const priceItem = prices[asset.priceId]?.[currency.coingeckoId];
-      if (!priceItem) continue;
+    const transferable = transferableAmountBN(balance);
+    const locked = BN.max(BN_ZERO, balance.free.sub(transferable));
+    const vested = BN.min(vestedLockedAmountBN(balance), locked);
```
stepanLav, line 53: **[correctness / consistency]** The widget-level allocation now carves the vesting lock out of “Locked” into a separate “Vested” segment, but the per-row allocation used by the chain/asset detail modals (`lib/computeRowAllocations.ts`, untouched by this PR) still lumps vesting into `lockedPct`.

An account whose only lock is vesting shows **Vested 40% / Locked 0%** in the overview bars, then **Locked 40% with no Vested segment** in the detail modal opened from that same widget — two contradictory “Locked” figures one click apart. `computeRowAllocations` needs the same vested carve-out.

src/renderer/domains/vesting/lib/vestingService.ts
```diff
@@ -0,0 +1,110 @@
...
+  getMinStartingBlock,
+  getMinVestedTransfer,
+  getMaxVestingSchedules,
+  getExistingVestingSchedules,
+  subscribeSchedulesForAccounts,
+  subscribeVestingLocksForAccounts,
+};
+
+function getMinStartingBlock(currentBlock: Record<ChainId, BlockHeight>, timelineChainId: ChainId): BN {
+  const minStartingBlock = currentBlock[timelineChainId]!;
```
stepanLav: **[correctness]** The non-null assertion hides a real hole: when the timeline chain's height is absent (right after startup before the 60s poll ticks, or when the relay chain is disabled), `currentBlock[timelineChainId]` is `undefined`, and `new BN(undefined)` silently yields `BN(0)` (verified against this repo's @polkadot/util). `$minStartingBlock` then becomes 0 and the `START_BLOCK_IN_PAST` refine in `features/vested-transfer/utils.ts:117` passes any positive starting block — the past-block warning is silently disabled during that window.

Return `null` (or skip validation explicitly) when the height is unknown instead of asserting.

src/renderer/domains/vesting/schedules/resource.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    const locksUnsub = vestingService.subscribeVestingLocksForAccounts(api, accountIds, next => {
+      locks = next;
+      sawLocks = true;
+      emit();
+    });
+
+    return () => {
+      schedulesUnsub.then(unsub => unsub());
+      locksUnsub.then(unsub => unsub());
```
stepanLav: **[correctness]** No `.catch` on either chain: `api.query.vesting.vesting.multi(...)`'s promise rejects if the RPC connection drops mid-subscribe, and this teardown then surfaces an unhandled rejection.

The reference implementation (`domains/governance/voting/resource.ts`) returns the subscription promise directly, letting `createSubscriptionResource`'s `unsubscribeFx` absorb the rejection as an effect failure (`shared/query/createSubscriptionResource.ts:103` handles `unsubscribe instanceof Promise`). The sync-closure wrapper here bypasses that path and leaves the inner promises floating — either return the promises to the framework or add `.catch` handling.

src/renderer/aggregates/vesting-portfolio/model.ts
```diff
@@ -0,0 +1,296 @@
...
+const $hasSchedules = $data.map(data => Object.keys(data.schedules).length > 0);
+
+/**
+ * A terminal state has been shown for the current account set — see
+ * `resolveStatus`.
+ */
+const $settledOnce = createStore(false);
+
+sample({
+  clock: combine($hasSchedules, $fullyResolved, (hasSchedules, fullyResolved) => hasSchedules || fullyResolved),
```
stepanLav: **[reliability / known pitfall]** An inline `combine(...)` derived store used as a `sample` clock is the exact pattern documented in this repo's Effector fork pitfall: emission tracking of a derived-store clock is not scope-local under `fork()`, so the `$settledOnce` latch can misbehave in forked integration tests (passes alone, fails in-suite). The integration framework runs under `fork()` (`tests/integrations/utils/framework/FeatureTestBuilder.ts:232`); no current test imports this aggregate, but the first one that does will hit it.

Safer equivalent: clock on the source stores —

```ts
sample({
  clock: [$hasSchedules, $fullyResolved],
  source: { hasSchedules: $hasSchedules, fullyResolved: $fullyResolved },
  filter: ({ hasSchedules, fullyResolved }) => hasSchedules || fullyResolved,
  fn: () => true,
  target: $settledOnce,
});
```

src/renderer/aggregates/vesting-portfolio/lib/views.ts
```diff
@@ -0,0 +1,311 @@
...
+  currentBlock,
+  blockTimes,
+  availableAccounts,
+  claimResolutions,
+  prices,
+  currency,
+}: VestingSource): { accountViews: AccountVestingView[]; summary: VestingSummary } => {
+  const accountsById = groupAccountsById(availableAccounts);
+
+  const balanceByKey = new Map<string, Balance>();
```
stepanLav: **[efficiency]** `computeVesting` rebuilds this Map over the *entire* global balance map (all accounts × chains × assets) on every recomputation, unconditionally — there's no early return when `data.schedules` is empty. The `$vesting` combine is constructed eagerly at bootstrap (`app/bootstrap/index.ts` → `vestingClaimFeature` → `VestingRoot` → this model), `$active` gates only the subscriptions, and `balanceModel.$balanceMap` emits roughly once per second while balance subscriptions push — so this is a permanent O(all-balances) rebuild per second even for users with zero vesting schedules. On top of that, `$blockHeights` (model.ts:125) folds the global 60s all-chains poll into the same combine with no `updateFilter`, adding ~one full recompute per connected chain per minute for chains irrelevant to vesting. The downstream fingerprint stores only limit re-renders, not the recompute itself.

Cheap wins: early-return an empty view when there are no schedules *before* building the Map; index only the `(accountId, chainId)` pairs that hold schedules; narrow `$blockHeights` to the timeline chains of the current requests with an `updateFilter`.

src/renderer/domains/network/block/hooks.ts
```diff
@@ -20,10 +20,16 @@ export const useBlockTime = (api?: ApiPromise | null, chain?: Chain | null) => {
...
+    map: (cache, { chain }) => cache[chain.chainId],
   });
 };
 
+/**
+ * Projects a block height on `chain`'s timeline to a wall-clock moment. `api`
+ * must be the api _of that chain_ — both the current height and the block time
+ * are read from it, and a mismatched pair silently dates the block against a
+ * foreign clock.
+ */
```
stepanLav, line 32: **[correctness — pre-existing, re-exposed by this PR]** Just below, `useBlockTimestamp`'s memo body reads `blockHeight`, but the dependency array (line 44) is `[currentBlock, blockTime]` — `blockHeight` is missing. If a component re-renders with a different `blockHeight` while `currentBlock`/`blockTime` are unchanged, the hook returns the previous stale timestamp. This bites the new consumer added in this PR: `VestingSchedulePreview`'s `startingBlockHeight` transitions from `null` to a real value once data resolves (same pattern as fellowship's `period?.from ?? null`), and if `currentBlock`/`blockTime` were already cached the memo keeps returning `null` until the next block-resource update.

Fix: `}, [currentBlock, blockHeight, blockTime]);`

stepanLav (commented):
## Code review — 10 findings (posted as inline comments)

Pipeline: 8 independent finder angles → dedup → per-candidate adversarial verification. One candidate was refuted during verification (re-subscription on ApiPromise replacement — network-model deletes/recreates the api entry, which re-keys the desired set correctly).

### Correctness (8)
1. **`vesting-claim/model/claim.ts:232`** — `signModel.signed` gated by `step !== Step.NONE` instead of `step === Step.SIGN`; a foreign/late signature hijacks the claim flow to SUBMIT.
2. **`vesting-portfolio/lib/views.ts:162`** — chains with an unresolved timeline (relay) height are silently dropped from views while `schedulesCount`/status still report them: callout advertises schedules the modal can never show, no claim path.
3. **`vesting-portfolio/lib/views.ts:263`** — `hasClaim` derived from fiat; priceless assets show a Claim button in rows while the callout/'Ready to unlock' stay hidden.
4. **`getMultisigSignOperationTitle.ts:20`** — `VEST` never registered with `multisigOperationsSDK`: pending multisig vest claims render as `unknownMst` / raw "Vesting: Vest" in the operations list.
5. **`useBalanceAllocation.ts:53`** — vested carve-out applied to the overview bars but not to `computeRowAllocations`: contradictory Locked % one click apart.
6. **`domains/vesting/lib/vestingService.ts:20`** — `new BN(undefined)` silently yields 0 when the timeline height is absent, disabling the `START_BLOCK_IN_PAST` floor.
7. **`domains/network/block/hooks.ts:44`** — `useBlockTimestamp` memo omits `blockHeight` from deps (pre-existing, re-exposed by the new `VestingSchedulePreview` consumer).
8. **`domains/vesting/schedules/resource.ts:82`** — teardown `.then(unsub => unsub())` without catch: unhandled rejection when RPC drops mid-subscribe; reference `governance/voting/resource.ts` returns the promise to the framework.

### Reliability / efficiency (2)
9. **`vesting-portfolio/model.ts:218`** — inline `combine` as sample clock: the documented Effector fork pitfall; `$settledOnce` can fail under forked integration tests.
10. **`vesting-portfolio/lib/views.ts:131`** — `computeVesting` rebuilds a Map over the entire global balance map ~1/s from bootstrap with no empty-schedules early return, plus a full recompute per connected chain per minute from the global block poll.

### Confirmed but below the cut (not posted inline)
- `vesting-claim/lib/datetime.ts` uses `toLocaleString(undefined, …)` — unlock dates ignore the in-app language; use `formatDate` from `useI18n`.
- `AccountScheduleModal.tsx` local `formatToken` duplicates `formatAsset` and drops digit grouping.
- `views.ts` `blockToDate` re-derives `getCreatedDate` (same formula this PR uses via `useBlockTimestamp`).
- `as ChainId`/`as AccountId` casts in `views.ts` (typed `entries` helper exists) and `AccountVestingView.chainId: string` instead of `ChainId` forcing downstream casts.

stepanLav (approved):

PR #5949, opened 2026-07-14
feat: multisig operations table redesign + all-types e2e

stepanLav:
## Summary

Reworks the **Multisig Operations** view to the new table-based design and adds a live-metadata end-to-end test covering every operation type.

The operations list becomes a **table** (sortable Operation / Value / Submitter / Description columns, sticky header) with rows grouped into collapsible **status sections** (In progress / Completed / Rejected / Hidden). Drafts move into the table as a first collapsible section and gain expandable detail panels. A new **Status filter** unifies drafts, in-progress, completed, rejected and hidden under one control, collapsing the tabs into an "All operations" merged scope when active.

## What changed

**Operations view**
- Table layout with a shared column spec (`shared/ui/operations-table-layout.ts`), sortable header (type / value / submitter), and status-sectioned virtualized list.
- Status filter + merged "All operations" scope; hidden ops surface only when the Status filter selects `Hidden`.
- Row redesign: inline description cell with a shared editor modal, `FROM DRAFT` badge, restyled status pill, actions fitted to columns.
- Signatories panel: flat list + embedded **Log** tab (replacing the old `LogModal`), with Overview/Share header actions. The log now shows only lifecycle events (initiated / signed / cancelled) with explorer links — the operation-title header was removed, and the `logTitle` SDK method was dropped from every operation-details feature.
- Advanced panel: hide/unhide keyed off actual hidden state (works in merged scope), multisig-deposit row removed, explorer popover portalled.

**Drafts**
- Drafts render inside the operations table (column-aligned), rows are expandable (Details / Signing path + overview/share / Advanced), Edit/Delete moved into the expanded panel. Delete is now gated by `operation-draft:write` (the backend dropped the dedicated `:delete` permissions).

**End-to-end test** (`tests/system/cases/multisig-operations/all-operation-types.system.test.ts`)
- Connects to a live **Polkadot Asset Hub** node, pulls its metadata, and builds genuine extrinsics (real callData/callHash) for every recognised operation family — for both a regular multisig and a flexible multisig — plus the special cards: edit-flexible **Atomic swap** and **Verified swap**, and the **verify-proxy** verification ping.
- Seeds them into IndexedDB, opens the view, asserts every family title + swap tags render, and **expands each of the 50 operations** to verify all parsed and show their details.
- Families whose pallet is absent from the live runtime are logged as `skipped` (never silently dropped).

## Testing
- `pnpm types:go` clean; eslint clean.
- Unit: `multisig-operations` + `drafts` suites — 259 tests green.
- E2E: `pnpm test:system` (needs `pnpm start:renderer`, `chains_dev`) — 50 operations rendered and expanded, 0 skipped.

## Review follow-ups (non-blocking)
- DRY: the operation row shell (`Operation.tsx`) and draft row shell (`DraftRow.tsx`) are near-identical scaffolds — candidate for a shared `RowLayout`.
- Hardcode: `ROW_HEIGHT = 68` (JS) vs `h-[68px]` (Tailwind) and `w-[110px]` are duplicated constant/literal pairs; `overscan: 15` and `maxLength: 500` could be named constants.
- KISS: the merged-scope tab pill wraps `Tabs` with `onChange={noop}` + `contents` spans.

## Notes
- Design sources for the rework live in the standalone design-system project (`nova-spektr-design`, extracted from this repo) and the claude.ai design project.
- Feature specs updated (`features/multisig-operations/README.md`, `features/drafts/README.md`) and indexed in the Feature Map.

## Example:

https://github.com/user-attachments/assets/43f9c2df-85fd-4c9f-bd6e-cdc1b8147b9a


savvyer (commented):
Needs fixing
1. Vested transfers with a cliff sort into the "no value" tier despite showing an amount
src/renderer/features/multisig-operations/lib/operations-sort.ts:43-69

getDisplayedAmount matches VESTED_TRANSFER only as the core tx, but buildVestedTransfer emits a utility.batchAll of vestedTransfer calls whenever the schedule has a cliff or more than one entry — and the title transformer renders the summed amount for exactly that case. So findCoreTransaction returns BATCH_ALL, which is neither in DISPLAYED_AMOUNT_TYPES nor a multi-transfer batch → null → -1 tier. A row displaying 1,000 DOT sorts below a 0.1 DOT plain transfer under Value ↓.

2. Submitter sort key doesn't match the name rendered in the Submitter column
src/renderer/features/multisig-operations/lib/operations-sort.ts:84-88

The key resolves against multisigWallets (MULTISIG / FLEXIBLE_MULTISIG only) and falls back to the raw hex multisigAccountId; the row renders <NamedAccount> against all wallets. Contact-backed multisigs (walletId = -1) and the proxied→flexible display accounts synthesized in model/context.ts (PROXIED walletId) therefore sort by 0x… while displaying a name. With "Alpha" (real multisig) and "Zeta" (contact multisig), Submitter ↑ puts Zeta first — the column reads as unsorted.

3. Virtualizer scrollMargin goes stale when the Drafts section collapses
src/renderer/features/multisig-operations/components/Operations.tsx:90-95

scrollMargin is only recomputed when Operations itself re-renders. This PR makes DraftsSection height-mutable from its own local state (isCollapsed, new here) and makes draft rows expandable — neither of which Operations subscribes to. Collapsing the section removes hundreds of px without scrollMargin following, so every item's start is offset by the delta: rows stop rendering and blank regions appear on scroll (the exact symptom the comment above the effect says it prevents), and scrollToIndex for a deep-linked op lands off-target. overscan: 15 masks one expanded panel; a section of ~15+ drafts exceeds it. Self-heals only when an unrelated store ticks.

4. Drafts ignore every filter except Status, yet are counted in the merged "All operations" total
src/renderer/pages/Operations/Operations.tsx:32, features/multisig-operations/components/Operations.tsx:66-67

mergedCount = visibleCount + visibleDrafts.length, while DraftsSection applies no network/type/date/search filtering. Select Network = Kusama: tabs collapse to "All operations N", the operations list is Kusama-only, but every Polkadot draft still renders and is counted in N. The unfiltered rendering is pre-existing; folding those rows into a count labelled "All operations" is what makes it wrong. README.md:14 documents Status-only gating, so this may be intentional — worth an explicit call.

docs/design/multisig-operations/Element Chat.dc.html
savvyer, line 1: i guess we should cleanup these design mockups. Lets prune the entire folder?

savvyer (approved):

PR #5950, opened 2026-07-14
fix: polkadot vault derivation keys editing

savvyer:
## Description
- **Derivation uniqueness** — paths are now unique per chain, and a collision is reported on both keys at once.
- **Export keys** — the export (file + QR) now reflects keys added during the current session, not just those present when wallet details opened.
- **Wallet details redesign** — key rows in the vault accounts list now use the shared `Account` component: chain icon, address, and an explorers popover carrying the derivation path.

## Related Issues
https://novasama-technologies.atlassian.net/browse/SPEK-701

## Screenshots/Recordings

https://github.com/user-attachments/assets/b25196c0-7cb0-4956-8f67-167e9c8f865e

Screenshot 2026-07-14 at 11 23 18 AM

## Checklist


src/renderer/entities/wallet/ui/Cards/DerivedAccount.tsx
```diff
@@ -1,32 +1,26 @@
...
-  children,
-  onClick,
-}: Props) => {
+/** Children are rendered as an extra section inside the explorers popover. */
+export const DerivedAccount = ({ account, chain, showSuffix, className, children, onClick }: Props) => {
   const isShardedAccount = accountUtils.isAccountWithShards(account);
-  const chainWithAccountId = !isShardedAccount && account.accountId;
-  const chainWithoutAccountId = !isShardedAccount && !account.accountId;
   const derivationPath = accountUtils.getDerivationPath(account);
+  const accountId = isShardedAccount ? null : account.accountId;
```
stepanLav: The old code branched on `!account.accountId`, so an empty-string `accountId` fell into the icon-and-path branch. `nullable()` only catches `null`/`undefined`, and per the README a partial Vault scan can produce exactly this broken account — it would now render `<Account accountId={''}>`: empty address, identicon of `''`, and explorer links built from an empty string.

```suggestion
  const accountId = isShardedAccount ? null : account.accountId || null;
```

src/renderer/features/wallets/KeyConstructor/model/constructor-model.ts
```diff
@@ -230,11 +240,6 @@ sample({
   }),
 });
 
-sample({
-  clock: [submitFx.doneData, submitFx.failData],
```
stepanLav, line 234: With this sample gone, the `Record<string, DerivationError[]>` that `submitFx` returns/throws is dead — nothing reads it anymore, the effect is just a gate for `onConfirm`. Its typing suggests errors still flow through it, which misleads the reader. KISS: make it return `void` and throw a plain `Error` when any key is invalid — the actual error surfacing already happens via `$touched` + derived `$errors`.

src/renderer/features/wallets/README.md
```diff
@@ -0,0 +1,163 @@
...
+A key is a **chain plus a derivation path** — nothing else. The user never picks a "key type": the strings `main`,
+`hot`, `public`, `sharded` offered as autocomplete chips are just text appended to the path, and every key the
+constructor produces is stored as a custom key. Crypto type (Ethereum vs sr25519) follows from the chosen chain.
+
+A path ending in a range token `0...N` is **sharded**: it expands into N+1 sibling keys (`//0`, `//1`, …) that stay
+grouped for balance display and signing.
+
+### Derivation path validation
+
+Errors surface **when the user finishes editing a key**, not on every keystroke — a half-typed path is not an error
```
stepanLav: Small precision nit: this holds only until a key's first blur. Once a key is in `$touched`, the derived `$errors` recomputes on every `updateKey`, so re-editing a touched key shows/clears errors live on each keystroke. The behaviour itself seems fine (errors clear as you fix them), but the spec shouldn't claim errors never surface mid-typing.

stepanLav (approved):

PR #5952, opened 2026-07-15
fix: teleport recipient list and myself button for key-set wallets

stepanLav:
## Description

Users with a key-set Polkadot Vault wallet (chain-scoped derived keys only, no universal root account) could not complete a teleport:

1. **Recipient dropdown was empty** — own-account suggestions were filtered by `isAccountAvailableOnChain`, a *signing* availability predicate under which a derived key only exists on its own chain, so on the destination chain everything was filtered out.
2. **"Myself" filled an unused Polkadot address** — it picked the single surviving availability-filtered account (an unused derived key or the root) instead of the sender.

The fix separates "who can sign here" from "who can receive here":

- New `transferUtils.canReceiveOnChain`: keyed accounts (vault keys, extension, WalletConnect) can receive on any scheme-compatible chain; keyless constructs (multisig, proxied/pure proxy) keep the strict availability rule — offering those cross-chain would send funds to uncontrolled addresses.
- "Myself" now resolves to the **sender address** on the destination chain whenever the sender can receive there, falling back to the previous single-account autofill / selection modal otherwise.

Note the intentional behavior change: with a scheme-compatible sender, "Myself" always fills the sender address — the account-picker modal only appears when the sender can't receive on the destination (e.g. substrate key → ethereum-based chain).

## Changes Made

- `features/transfer/lib/transfer-utils.ts` — `canReceiveOnChain` + `filterRecipientAccounts` (recipient-scoped filtering, extracted from `TransferForm`)
- `features/transfer/model/form-model.ts` — `$myselfDestination` (sender-first resolution), consolidated `myselfClicked` samples, `$isMyselfXcmEnabled` follows the same derivation
- `features/transfer/ui/TransferForm.tsx` — recipient options use the new helper; fixed stale `t` in options memo deps
- Tests: unit coverage for the recipient filter (incl. keyless-account exclusion) and an integration suite reproducing the reported scenario (sender preference, fallback autofill, proxied sender kept disabled)

## Checklist


savvyer (commented):
1. **Watch-only accounts offered as recipients** — `canReceiveOnChain` in `transfer-utils.ts:54-57`. Only multisig/proxied are excluded, so watch-only addresses (keys the user doesn't hold) get suggested as receivable on any scheme-compatible chain. Contradicts the "don't offer addresses you don't control" rationale — consider an explicit keyed-type check.

non-blocking:
2. **Arbitrary account pick on shared accountId** — `recipientAccount` in `TransferForm.tsx:454`. Plain `.find` returns an arbitrary account when an accountId exists in multiple wallets, so the wallet badge/name can be wrong. `accountService.findRelatedAccount` already handles this.

savvyer (approved):

PR #5955, opened 2026-07-16
fix: submitted drafts showing in both drafts and in progress

savvyer:
## Description
Submitted drafts stayed in the Drafts section and the in-progress row lost its description.

Cause: "is this draft submitted" was derived by fetching every operation description in the workspace and matching `draftId`. `fetchAllDescriptions` parsed the `{ data, total }` envelope as `[]`, so that cache was always empty — nothing was hidden, and the same cache backs the row's description.

Fix: the backend already returns the linked `operation` on each draft. Read it; delete the join.

## Related Issues
https://novasama-technologies.atlassian.net/browse/SPEK-702

## Screenshots/Recordings

https://github.com/user-attachments/assets/b90b32df-7a4c-4fa1-8d65-e11c29e96815

## Checklist


src/renderer/domains/backend/drafts/service.ts
```diff
@@ -76,11 +84,33 @@ async function createDraft(
...
-  const parsed = parseResponse(result, listResponseSchema);
+  const firstPage = await fetchDraftsPage(baseUrl, 1);
+
+  if (firstPage.total <= PAGE_SIZE) {
+    return firstPage.data;
+  }
+
+  const totalPages = Math.ceil(firstPage.total / PAGE_SIZE);
+  const remainingPages = Array.from({ length: totalPages - 1 }, (_, i) => i + 2);
+  const remainingResults = await Promise.all(remainingPages.map(page => fetchDraftsPage(baseUrl, page)));
```
stepanLav, line 106: This fires all remaining pages at once — a 10k-draft workspace means ~100 concurrent requests against the backend. Since this runs on a background poll, latency hardly matters; fetching sequentially (or in chunks of 3–5) would be kinder to the server.

src/renderer/domains/backend/drafts/service.ts
```diff
@@ -76,11 +84,33 @@ async function createDraft(
...
+
+  const allDrafts = [firstPage.data, ...remainingResults.map(result => result.data)].flat();
 
-  return parsed.data;
+  // Offset paging isn't atomic and the backend's `createdAt DESC` has no
+  // tiebreaker, so a concurrent insert can return the same draft on two pages.
+  // Map keeps the later page's copy, which was read later.
```
stepanLav: Two nits on this comment: pages are fetched in parallel, so "read later" really means "higher page number"; and the dedup fixes duplicates but not misses — an insert can also push a row across the boundary of a not-yet-read page, and it's gone until the next poll. Both fine given the 30s poll heals it, but worth saying so here. The real fix is a tiebreaker in the backend ordering (`createdAt DESC, id DESC`) — maybe a follow-up issue.

src/renderer/domains/backend/drafts/resource.ts
```diff
@@ -31,4 +31,11 @@ $cache.on(draftUpdated, (list, updated) => list.map(d => (d.id === updated.id ?
 $cache.on(draftDeleted, (list, id) => list.filter(d => d.id !== id));
 $cache.on(resetDrafts, () => []);
 
-export const draftsResource = { ...resource, draftCreated, draftUpdated, draftDeleted, resetDrafts };
+// `$cache` starts as `[]`, so its length can't distinguish "not loaded" from
+// "loaded and empty". `useResource`'s `pending` can't serve either: it tests
+// `map(cache) ?? empty`, and `[]` is never nullish.
+const $loaded = createStore(false)
+  .on(resource.push, () => true)
```
stepanLav, line 38: Heads-up: `push` only fires for non-cached results (`createQueryResource` filters on `!result.cached`), so after `resetDrafts` a re-start within `staleAfter` (20s) is served from the request cache and neither `$cache` nor `$loaded` recover until the next poll — the widget can sit in "loading" for up to ~30s after a reconnect. Consistent and self-healing, so fine to ship; if it ever bites, resetting the resource's request cache alongside `resetDrafts` is the fix.

src/renderer/domains/backend/drafts/service.ts
savvyer: Browsers cap concurrent connections per host, so web never actually opens 100 sockets — and a 10k-draft workspace isn't reachable since drafts are deleted on submit. Serializing would be worse: at ~20s for 100 pages it risks overlapping the 30s poll.

But I've added additional guard, so we dont send fetch requests if user is not authenticated

src/renderer/domains/backend/drafts/resource.ts
savvyer: No action needed — the failure mode is a spinner (never wrong data) and it self-heals within one poll tick

stepanLav (approved):

PR #5957, opened 2026-07-17
fix: validate and name the selected key for key-set wallets

stepanLav:
## Description

Key-set Polkadot Vault wallets hold several derived keys on one chain, but the flexible multisig creation flow picked the **first chain-available account of the signatory's wallet** for fee/proxy-deposit validation and for the signing payload — ignoring the key the user actually selected. A user who funded the selected key kept seeing "insufficient funds" because a different (empty) key of the same wallet was being validated. The explicit signer pick from `SignerSelection` was also silently overridden by the derived `signatories[0]`.

On top of the fix, balance errors are now key-specific app-wide: for wallets with more than one distinct key the shared `TransactionValidationError` names the exact failing account ("Account KEY of wallet W has insufficient funds… top up the balance of this account…"), so the user knows which address to fund. Single-account wallets keep the current wallet-level wording.

## Related Issues

Reported by a user creating a flexible multisig on Polkadot Asset Hub with a key-set vault (validation demanded a top-up despite the selected key being funded 16 minutes earlier).

## Changes Made

- `accountService.resolveSelectedAccount` — canonical resolver "walletId + address + chain → exact account entity" (prefers the chain-scoped key when several accounts share an accountId).
- `createSelectedInitiatorStore` (`shared/transactions`) — store factory for flows whose initiator comes from a picker outside effector-forms; the route is always built from the explicit user selection.
- Flexible multisig model: `$initiator` derives from the selected signatory via the factory; `$signatory` honors the explicit pick while it is present in the derived list (falls back to `signatories[0]` otherwise).
- `TransactionValidationError`: adaptive `introAccount`/`requiredAccount` wording for key-set wallets (all ~28 consumers get it with zero call-site changes); hooks hoisted above early returns.
- Tests: resolver unit tests, factory tests, fork-based model tests (including a 2-leaf signatory graph proving the pick beats array order, and a non-key-set regression case), component tests for the adaptive wording incl. shared-accountId-across-chains and no-name fallback.

Audited the other 30+ consumers of the route-based validation system: they take the account from an explicit form field or the signing path — no changes needed. The regular multisig flow already matches the selected accountId and was deliberately left untouched.

## Screenshots/Recordings

Verified live on Polkadot Asset Hub with a 2-key vault wallet: selecting the empty key shows "Account KEY_BRAVO of wallet … top up the balance of this account by 0.41466 DOT"; switching to the partially funded key changes both the named account and the required amount (0.40466 DOT) — validation reads the selected key's balance.

## Checklist


savvyer (commented):
Bug: [TransactionValidationError.tsx:243](vscode-webview://095ag9s4fhhcjcevpns6hn8dc13gppdr5o1ud7t7bmt2mfa4cfcc/src/renderer/shared/ui-entities/TransactionValidationError/TransactionValidationError.tsx#L243) treats any wallet with 2+ distinct accountIds as a key-set, but WalletConnect/Nova wallets get a separate account per chain — an EVM chain returns a 20-byte H160 accountId alongside the 32-byte substrate one, so a single-key wallet counts as 2.

Since those accounts are named after the wallet ([form.ts:99](vscode-webview://095ag9s4fhhcjcevpns6hn8dc13gppdr5o1ud7t7bmt2mfa4cfcc/src/renderer/features/wallet-connect-wallet-pairing/model/form.ts#L99)), the error then reads "Account Nova Wallet of wallet Nova Wallet has insufficient funds" — the picker already guards this exact case at [Signatory.tsx:136](vscode-webview://095ag9s4fhhcjcevpns6hn8dc13gppdr5o1ud7t7bmt2mfa4cfcc/src/renderer/features/multisig-wallet-create/flexible-multisig/components/components/Signatory.tsx#L136), the new code doesn't.

PR #5966, opened 2026-07-21
feat: add initiator column to drafts and multisig operations tables

savvyer:
## Description
Multisig operations initiator column

## Related Issues
Closes https://novasama-technologies.atlassian.net/browse/SPEK-704

## Screenshots/Recordings

https://github.com/user-attachments/assets/bd17e115-8ee0-4d3f-8271-dcfc98088880

## Checklist


src/renderer/features/multisig-operations/lib/operations-filter.ts
```diff
@@ -146,48 +150,63 @@ type WalletSearchSources = {
...
+        chain: isFlex ? (chains[account.chainId] ?? null) : null,
+        walletName: walletNames.get(account.walletId) ?? null,
+      },
+      {
+        // Initiator, shown in the expanded details. When the depositor belongs
+        // to a local wallet, `OperationDetails` renders its wallet name rather
+        // than the resolved account name, so search must cover it too.
+        accountId: operation.depositor,
+        chain: initiatorChain,
+        walletName: resolveWalletName(operation.depositor, initiatorChain),
```
stepanLav, line 194: Search matches the *resolved* wallet name here, but `OperationDetails` still renders the raw `depositorWallet.name` (line 44 there). Whenever the two diverge (custom account name, contact or identity overrides the wallet's primary account), the user types the exact string they see in the panel and gets zero results — the exact bug class `docs/code/style/search.md` forbids. The aggregate README defers this as a follow-up, but the fix is one line in the panel — `useWalletName(depositorWallet)` instead of `wallet.name` — so I'd rather close it in this PR than ship a documented search miss.

src/renderer/features/multisig-operations/components/Operation.tsx
```diff
@@ -198,6 +198,16 @@ export const Operation = memo(({ operation, multisigAccount, isDefaultOpen = fal
               )}
             </div>
 
+            <div className={cnTw(operationColumns.initiator, 'items-center')}>
+              <NamedAccount
+                accountId={operation.depositor}
```
stepanLav, line 209: The depositor is very often a local-wallet signatory, and CLAUDE.md asks to pass `wallet={…}` to `NamedAccount` in that case. Without it, the collapsed row can show a short address / raw account name while the expanded Depositor detail shows the wallet name — two different names for the same account in one card. It also contradicts the README's claim that this column is "resolved the same way the expanded Depositor detail is". Suggest looking up the depositor's account/wallet the way `OperationDetails` (lines 27–28) and `DraftRow`'s submitter already do, and passing it — that would also make the `resolveWalletName` search meta on this row match what the collapsed row displays.

src/renderer/features/drafts/lib/draft-scope.ts
```diff
@@ -38,37 +39,79 @@ const matchesDateRange = (draft: Draft, dateRange: DateRange | undefined): boole
...
+    seen.add(accountId);
+
+    accounts.push({
+      accountId,
+      chain,
+      walletName:
+        accountId === submitterAccountId
+          ? // A proxied draft renders through a synthetic wallet named after the
+            // proxy contact, which no resolver produces.
+            (draft.proxyContact?.name ?? resolveWalletName(accountId, chain))
```
stepanLav, line 82: Narrow edge worth a test: for a proxied draft *without* `proxyContact`, display resolves the name through the synthetic PROXIED wallet (`useWalletName` → wallet's primary account), while search resolves through the real wallet via `findRelatedAccount`. For a multi-account wallet (e.g. a Vault holding the proxied key) the two chains can pick different accounts and produce different names — a displayed name the search misses, which is exactly the divergence class this PR hunts. The existing tests only cover the `proxyContact` path.

stepanLav (approved):

